PR #1, opened 2015-03-24
Initial take to create sti-python image

soltysh:


3.3/.sti/bin/assemble
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -e
+#
+# Default STI assemble script for the python-3.3 image.
+#
+source "../lib/python_context"
```
mfojtik: I would prefer to have one way to enable SCL than this and X_SCLS

soltysh: So that's another thing that we should agree on and do for other language images. 
@jhadvig @mnagy what do you guys say? I can do that for the other images as well :wink:

mnagy: Please use `source scl_source enable python33`, it was recommended to us by SCL people.

mnagy: Also, I'd recommend not using -e in shebang and instead add it after sourcing scl_source. That way you can avoid needing to set +e inside .sti/bin/run

3.3/.sti/bin/run
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -e
+#
+# Default STI run script for the python-3.3 image.
+#
+
+# TODO: Nuke this hack when SCL-2.0 are out
+# For now this is the only way how to prevent
+# shebang from spawning a new bash process.
+set +e
```
mnagy: /bin/bash -e and set +e right after it? :)

README.md
```diff
@@ -1 +1,169 @@
...
+CentOS versions currently supported are:
+* CentOS7
+
+
+Installation
+---------------
+To build Python image, choose between CentOS or RHEL based image:
+*  **RHEL based image**
+    This image is not available as automated build in [Docker Index](https://index.docker.io).
+
```
mnagy: I thing it was agreed not to write this into READMEs?

soltysh:
@mnagy @jhadvig @mfojtik @bparees PTAL. I'd say this is the first acceptable version of python image. I'll be working on it more over the next few days, but as a follow up PRs.

3.3/.sti/bin/assemble
```diff
@@ -0,0 +1,14 @@
...
+source scl_source enable python33
+set -e
+
+echo "---> Installing application source"
+cp -Rf /tmp/src/* ./
+
+echo "---> Building your Python application from source"
+if [ -f setup.py ]; then
+    python setup.py install --user
+elif [ -f requirements.txt ]; then
```
mfojtik: can you have both? (setup.py && requirements.txt)

3.3/.sti/bin/run
```diff
@@ -0,0 +1,23 @@
...
+
+if [[ ! -v APP_MODULE  && -f setup.py ]]; then
+    APP_MODULE=`python setup.py --name`":application"
+fi
+
+if is_gunicorn_installed && [[ -v APP_MODULE ]]; then
+    if [[ -v APP_CONFIG ]]; then
+        export CONFIG="--config ${APP_CONFIG}"
+    fi
+    exec gunicorn ${APP_MODULE} --bind=:8080 ${CONFIG}
```
mfojtik: can we use metadata for this instead of hardcoding?

3.3/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+ENV IMAGE_TAGS            python,python33
+ENV IMAGE_EXPOSE_SERVICES 8080:http
+
+RUN yum install -y \
+    https://www.softwarecollections.org/en/scls/rhscl/python33/epel-7-x86_64/download/rhscl-python33-epel-7-x86_64.noarch.rpm && \
+    yum install -y --setopt=tsflags=nodocs python33 python33-python-devel python33-python-setuptools && \
+    yum clean all -y && \
+    scl enable python33 "easy_install pip"
+
+ENV PATH .local/bin/:$PATH
```
mfojtik: what is in .local/bin ?

README.md
```diff
@@ -1 +1,188 @@
...
+    To build a rhel-based python-3.3 image, you need to run the build on properly
+    subscribed RHEL machine.
+
+    ```
+    $ git clone https://github.com/openshift/sti-python.git
+    $ cd sti-python
+    $ make build TARGET=rhel7 VERSION=3.3
+    ```
+
+*  **CentOS based image**
```
jhadvig: Worth to mention that centos7 image is available on DockerHub

3.3/.sti/bin/assemble
soltysh: Yes, and that's usually the case where `requirements.txt` are read into `setup.py`. But sometimes people want to have two different requirement sets: one for test and the other for running. This should address this.

3.3/Dockerfile
soltysh: Since I'm running as `default` user I don't have access to system wide lib dir where usually python keeps deps, also I'm not using virtualenv because I don't think it's necessary inside docker container. To overcome this I'm using --user flag for installing deps which puts all of those into .local dir. The .local/bin contains then binaries to run, eg. gunicorn.

3.3/.sti/bin/run
soltysh: Done.

README.md
soltysh: We don't have that kind of information in other images, should we?

3.3/.sti/bin/usage
```diff
@@ -0,0 +1,16 @@
...
+This is a STI python-3.3 ${DISTRO} base image:
+To use it, install STI: https://github.com/openshift/source-to-image
+
+Sample invocation:
+
+sti build https://github.com/openshift/sti-python.git --contextDir=3.3/test/test-app/ openshift/python-33-${DISTRO}7 python-sample-app
+
+
+You can then run the resulting image via:
+docker run -p 8080:8080 pythn-sample-app
```
jhadvig: typo - `python`

README.md
jhadvig: we do, but for some reason its missing in ruby && nodejs, but it is in all other images
sti-base : https://github.com/openshift/sti-base#installation-and-usage
mongodb : https://github.com/openshift/mongodb#installation

3.3/.sti/bin/assemble
bparees: should they be able to choose which one via an env variable they can put into their .sti/environment file?

3.3/Dockerfile.rhel7
```diff
@@ -0,0 +1,24 @@
...
+ENV IMAGE_TAGS            python,python33
+ENV IMAGE_EXPOSE_SERVICES 8080:http
+
+RUN yum-config-manager --enable rhel-server-rhscl-7-rpms \
+    yum-config-manager --enable rhel-7-server-optional-rpms && \
+    yum install -y --setopt=tsflags=nodocs python33 python33-python-devel pip3 && \
+    yum clean all -y && \
+    scl enable python33 "easy_install pip"
+
+ENV PATH .local/bin/:$PATH
```
bparees: $PATH doesn't work for RHEL right now, right?

README.md
```diff
@@ -1 +1,188 @@
...
+CentOS versions currently supported are:
+* CentOS7
+
+
+Installation
+---------------
+To build Python image, choose between CentOS or RHEL based image:
+*  **RHEL based image**
+
+    To build a rhel-based python-3.3 image, you need to run the build on properly
```
bparees: on a properly subscribed...

README.md
```diff
@@ -1 +1,188 @@
...
+Test
+---------------------
+This repository also provides [STI](https://github.com/openshift/source-to-image) test framework,
+which launches tests to check functionality of a simple python application built on top of sti-python image.
+
+User can choose between testing python test application based on RHEL or CentOS image.
+
+*  **RHEL based image**
+
+    This image is not available as automated build in [Docker Index](https://index.docker.io).
```
bparees: i'm not sure the point in mentioning this since "make test" builds the image anyway.  (I think we have it on other READMEs and we should remove it there too, unless someone can tell me why we have it)

jhadvig: +1
its only in `sti-nodejs` but will remove that

3.3/.sti/bin/assemble
soltysh: @bparees not sure what you mean by that. `setup.py` is a complete information about your package/application (including name, author and deps). In that case using `requirements.txt` doesn't make sense as in most cases its contents will be already read into `setup.py`. In all other cases I'm just falling back to `requirements.txt` if it exists for a list of deps.

3.3/Dockerfile.rhel7
soltysh: It's not defined in the base image, that's why we're doing [this](https://github.com/openshift/sti-base/blob/master/Dockerfile.rhel7#L48), so I'm safe at this level.

README.md
soltysh: It's there, but wrapped in the next line :wink: I'll move it into this line then.

soltysh:
Comments addressed PTAL.

bparees:
lgtm.

soltysh:
@mfojtik @jhadvig @mnagy if you don't mind I'd like to merge it by tomorrow. I'll be fixing any leftovers in following PRs.

mfojtik:
Yes, LGTM

jhadvig:
Other then those couple of nits, LGTM 
gj :+1:

soltysh:
All the comments were addressed, so merging as is.

PR #7, opened 2015-04-21
[DO_NOT_MERGE] Moved easy_install pip command to assemble script

soltysh:


soltysh:
@rhcarvalho please review.

rhcarvalho:
@soltysh could you describe briefly why this change is needed?

soltysh:
After discussing this offline, we'll defer that change for a later time.

PR #8, opened 2015-04-23
Django + PostgreSQL demo

rhcarvalho:
This is a demo project demonstrating how to deploy a Python application powered by Django and PostgreSQL to OpenShift.


rhcarvalho:
I'm looking for a way to perform the actions in `bootstrap.sh` in a way other than via `docker exec -it <frontend-container> bash`

rhcarvalho:
@mfojtik @soltysh is there a cleaner way to run a one-off command in the context of the application? (see `3.3/examples/django/run-in-app-container.sh`)

mfojtik:
@rhcarvalho going to review this today :-) thanks

3.3/examples/django/bootstrap.sh
```diff
@@ -0,0 +1,50 @@
...
+
+if [ $USER = vagrant ]; then
+	OPENSHIFT_BIN=/data/src/github.com/openshift/origin/_output/local/go/bin/openshift 
+	OPENSHIT_PARAMS="--public-master=localhost"
+else
+	OPENSHIFT_BIN=openshift
+fi
+
+# Start OpenShift in the background
+echo "-> starting OpenShift in the background...  (tail -f $OPENSHIFT_LOGFILE)"
```
mfojtik: Can you sync these messages with what Python cartridge is printing out to make it consistent?
Also for reporting the progress, we agree on using this format:

echo "Message ..."

3.3/examples/django/README.md
```diff
@@ -0,0 +1,82 @@
...
+This section assumes that you have built the `openshift` binary and generally have OpenShift 3 ready to go. See https://github.com/openshift/origin/blob/master/CONTRIBUTING.adoc.
+
+Start by cloning this repository and changing your current directory appropriately:
+
+```bash
+git clone https://github.com/openshift/sti-python.git
+cd sti-python/3.3/examples/django
+```
+
+### Start OpenShift ###
```
soltysh: Although this is great idea, I'm not sure we want to have it here, since the start-up procedure is changing rapidly in OpenShift, we might forget to update this. Thus I'd prefer you point to the sample-app demo with remark to use your template file instead of the original one. This way it'll be easier for us to maintain, IMHO.

3.3/examples/django/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+Cleaning up
+-----------
+
+You can use the `cleanup.sh` script to clean up your local environment after running this demo:
+
+```bash
+./cleanup.sh
+```
```
soltysh: Generally speaking I'd focus inside of this file entirely on running your sample application inside of openshift, leaving out all the usual general OS details to OS's sample-app.

3.3/examples/django/application-template-stibuild.json
```diff
@@ -0,0 +1,282 @@
...
+          "type": "STI"
+        }
+      },
+      "triggers": [
+        {
+          "github": {
+            "secret": "secret101"
+          },
+          "type": "github"
+        },
```
soltysh: You're not using the GH webhook anywhere in the code, I'd remove it from here, so this file is as small as needed.

3.3/examples/django/application-template-stibuild.json
```diff
@@ -0,0 +1,282 @@
...
+        {
+          "generic": {
```
soltysh: The same applies to generic.

3.3/examples/django/application-template-stibuild.json
```diff
@@ -0,0 +1,282 @@
...
+    {
+      "description": "database name",
+      "name": "POSTGRESQL_DATABASE",
+      "value": "django_demo"
+    }
+  ],
+  "labels": {
+    "template": "application-template-stibuild"
+  }
+}
```
soltysh: If this is example, maybe before each object put a detailed comment what it does, and its purpose. Then in the README you'd point to this file and its comments. Other option is to copy parts of this file into the README, but the latter is more cumbersome in the long run when it comes to updating, and having this in one place is simpler.

3.3/examples/django/demo/settings.py
```diff
@@ -0,0 +1,116 @@
...
+import os
+
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+
+# Quick-start development settings - unsuitable for production
+# See https://docs.djangoproject.com/en/1.8/howto/deployment/checklist/
+
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = '+u3anikga+6zkzk&@6l_rimzw$48p!2#52=$c2v#x#q2n@c0&e'
```
soltysh: This is a good candidate for a template mechanism, IMHO.

3.3/examples/django/application-template-stibuild.json
mfojtik: I would keep it as this application is not indented to be 'effective' but 'descriptive' ;-) It is fine to show that you can also use GH/generic web hooks to trigger a build (even thought they are not reachable and hardly demoable atm.)

3.3/examples/django/notes/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
+
+# Create your tests here.
```
soltysh: If that is our flagship app, let's have some :smile:

3.3/examples/django/run-in-app-container.sh
```diff
@@ -0,0 +1,12 @@
...
+# Use this script to run one-off commands inside a container of a frontend pod
+# (where you application code lives in)
+#
+# Examples:
+# ./run-in-app-container.sh date
+# ./run-in-app-container.sh env
+# ./run-in-app-container.sh ./manage.py migrate
+
+quoted_args="$(printf " %q" "$@")"
+osc exec -p $(osc get pods -l name=frontend -t '{{ with index .items 0 }}{{ .metadata.name }}{{ end }}') -it -- bash -c "cd \$HOME && scl enable python33 \"$quoted_args\""
```
soltysh: Question, I see you're enforcing `scl enable python33` doesn't it work without it? I was hoping `.bashrc` will be called upon entering the container, but I might be wrong.

soltysh:
I'm missing here a step by step guide how to run the app, how to access it, how to use your `run-in-app-container.sh` script in the readme.
I haven't tried it yet, but I will later today to see what I get :wink:

3.3/examples/django/notes/tests.py
mfojtik: negative :-) this is just demo app, not a flagship app :)

3.3/examples/django/run-in-app-container.sh
mfojtik: just a note, you should not do `osc exec` to perform migration, but the migration should happen automatically when you create the app or you do an update to your application source code.

soltysh: On Mon, Apr 27, 2015 at 12:28 PM, Michal Fojtik <notifications@github.com>
wrote:

> In 3.3/examples/django/run-in-app-container.sh
> <https://github.com/openshift/sti-python/pull/8#discussion_r29135774>:
>
> > @@ -0,0 +1,12 @@
> > +#!/bin/bash
> > +
> > +# Use this script to run one-off commands inside a container of a frontend pod
> > +# (where you application code lives in)
> > +#
> > +# Examples:
> > +# ./run-in-app-container.sh date
> > +# ./run-in-app-container.sh env
> > +# ./run-in-app-container.sh ./manage.py migrate
> > +
> > +quoted_args="$(printf " %q" "$@")"
> > +osc exec -p $(osc get pods -l name=frontend -t '{{ with index .items 0 }}{{ .metadata.name }}{{ end }}') -it -- bash -c "cd \$HOME && scl enable python33 \"$quoted_args\""
>
> just a note, you should not do osc exec to perform migration, but the
> migration should happen automatically when you create the app or you do an
> update to your application source code.
>
I partially agree with the migration, but still having such helper command
does makes sense and will be useful instead of doing osc get pods etc... On
the other hand this could be the same as with osc log, pick first pod and
run exec, if you want more fine grained control you still can, just a
thought :wink:.

3.3/examples/django/application-template-stibuild.json
```diff
@@ -0,0 +1,282 @@
+{
+  "apiVersion": "v1beta1",
+  "items": [
+    {
+      "apiVersion": "v1beta1",
+      "containerPort": 8080,
+      "id": "frontend",
+      "kind": "Service",
+      "port": 5432,
```
soltysh: I know we have that weird port in our sample-app example, but maybe changing it here to something different.

3.3/examples/django/bootstrap.sh
rhcarvalho: I'll remove this script and point to the origin/sample-app for instructions in the README.md, keeping those steps just for myself.

3.3/examples/django/README.md
rhcarvalho: Agreed. Will update accordingly.

3.3/examples/django/application-template-stibuild.json
rhcarvalho: @soltysh AFAICT we can't have comments in a JSON file. Explaining in the README is ok. Is this feature fully documented somewhere already?

3.3/examples/django/demo/settings.py
rhcarvalho: :+1:

mfojtik: @soltysh @rhcarvalho yeah, I think this is pretty natural to have in the template as parameter with generator that generates this value.... This value must be the same for all Pods created by RC right? So it must be an ENV var we set from top.

3.3/examples/django/application-template-stibuild.json
mfojtik: @rhcarvalho you can use annotations to provide better description ("description" annotation field). Just keep in mind that this annotation most likely will be showed in the UI.

3.3/examples/django/demo/settings.py
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+
+
+# Internationalization
+# https://docs.djangoproject.com/en/1.8/topics/i18n/
+
+LANGUAGE_CODE = 'en-us'
+
+TIME_ZONE = 'UTC'
```
mfojtik: does the system `TZ` is ignored? What if I want to run python app with CET timezone?

3.3/examples/django/demo/settings.py
```diff
@@ -0,0 +1,116 @@
...
+            'ENGINE': 'django.db.backends.sqlite3',
+            'NAME': os.path.join(BASE_DIR, 'db.sqlite3'),
+        }
+    }
+
+
+# Internationalization
+# https://docs.djangoproject.com/en/1.8/topics/i18n/
+
+LANGUAGE_CODE = 'en-us'
```
mfojtik: This is a good candidate for parameter as well.

3.3/examples/django/notes/tests.py
soltysh: Flagship demo app :wink: still I'd opt for having at least one or two

3.3/examples/django/application-template-stibuild.json
soltysh: +1 for the description

soltysh:
One final comment, I (dumb user :wink, I know I'm being picky) just run the example (I followed the sample-app tutorial) and after reaching my frontend address I see:
```

Page not found (404)
Request Method: 	GET
Request URL: 	http://localhost:9999/

Using the URLconf defined in demo.urls, Django tried these URL patterns, in this order:

    ^notes/
    ^admin/

The current URL, , didn't match any of these.
```
OK, I enter `notes/` and I see:
```
relation "notes_note" does not exist
LINE 1: SELECT COUNT('*') AS "__count" FROM "notes_note"
```
Bad luck, I try `admin/` and I'm getting logging form but what's the user and password I should use?

Of course you provided the `bootstrap-django.sh` but it's not being invoked anywhere and there's no explanation what should I do with it. Like I said before focus on explaining how to run the application inside of OpenShift so that any user with any level of knowledge will be able to follow the steps (even not reading the comments around it) and have the application running.

3.3/examples/django/run-in-app-container.sh
rhcarvalho: @soltysh it did NOT work without explicitly doing `scl enable python33`. I'd rather have it work without it as well. Somehow, without that:

```bash
$ osc exec -p $(osc get pods -l name=frontend -t '{{ with index .items 0 }}{{ .metadata.name }}{{ end }}') -it -- bash -c "python --version"
Python 2.7.5
```

3.3/examples/django/application-template-stibuild.json
rhcarvalho: I'm keeping those, as per @mfojtik's comment.

soltysh: ok

3.3/examples/django/run-in-app-container.sh
soltysh: Probably osc exec does not invoke .bashrc, @ncdc would that be possible at all?

3.3/examples/django/demo/settings.py
rhcarvalho: The timezone setting here is arbitrary. It's used to show dates on your web app.

3.3/examples/django/demo/settings.py
rhcarvalho: I do not agree. If we would parameterize all Django settings we'll have a monstrous Template.
Django users will look at `settings.py` to change Django settings, not in some OpenShift-specific configuration.
LANGUAGE_CODE, TIME_ZONE, etc, are essentially different than SECRET_KEY, because parameterizing SECRET_KEY and generating a random value for it provides *value* to the end user. But creating one more level of abstraction between `settings.py` and other settings' keys would be a bad move IMHO.

rhcarvalho:
@soltysh we're not focusing on that kind of scenario now. As things stand now, the entry-level barrier to get to use the app is much higher. I'll be working on a more "user-usable" code base as a follow up to this demo example.

3.3/examples/django/demo/settings.py
```diff
@@ -0,0 +1,116 @@
...
+            'PORT': os.getenv('DATABASE_SERVICE_PORT'),
+        }
+    }
+else:
+    DATABASES = {
+        'default': {
+            'ENGINE': 'django.db.backends.sqlite3',
+            'NAME': os.path.join(BASE_DIR, 'db.sqlite3'),
+        }
+    }
```
rhcarvalho: Normally Django projects have multiple `settings.py`, each with specific settings for specific environments. Here we clearly need to support at least 2: local development and openshift.

@mfojtik @soltysh any insights on how we could do this better?

What I miss the most is a way to run the app locally.... (should we encourage developers to have their own local openshift server for development?!)

soltysh: One that comes to my mind is using env vars, as usual.
As for running apps, the usual way, test locally (as you'd normally do) and then deploy to OpenShift. I don't see any benefit from running OS locally.

3.3/examples/django/README.md
```diff
@@ -0,0 +1,91 @@
+Django application example
+==========================
+
+This is a basic [Django](https://www.djangoproject.com/) project and some helper scripts to demonstrate how to deploy Python web applications with [OpenShift 3](http://www.openshift.org).
+
+We will be using Django 1.8 and the [PostgreSQL](http://www.postgresql.org/) database.
+
+OpenShift supports multiple deployment strategies. This example uses the Source-To-Image strategy, and the [Python 3.3 STI image](https://github.com/openshift/sti-python/tree/master/3.3) for producing application builds.
```
soltysh: STI is a build strategy not deployment.

3.3/examples/django/README.md
```diff
@@ -0,0 +1,91 @@
...
+------------------------------------------
+
+Start by cloning this repository and changing your current directory appropriately:
+
+```bash
+git clone https://github.com/openshift/sti-python.git
+cd sti-python/3.3/examples/django
+```
+
+Follow the instructions in https://github.com/openshift/origin/tree/master/examples/sample-app to build the `openshift` all-in-one binary and start your OpenShift 3 server and docker registry. You won't need to clone that repository, and you should use `sti-python/3.3/examples/django` as your working directory.
```
soltysh: It would be good to add that one should proceed the sample-app until he reaches a point with `osc process -f application-template-stibuild.json | osc create -f -` in which case he should use provided `application-template-stibuild.json` file.

3.3/examples/django/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+Before proceeding, make sure you have the OpenShift server with a docker registry, a project and a user.
```
soltysh: ... make sure you have OpenShift server **running**, with a docker registry...

3.3/examples/django/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+
+### Building your app ###
+
+In the future, creating a new app in OpenShift from a git repo should automatically detect what kind of application it is, and create some entities to support it.
```
soltysh: Nuke this, it's not needed.

3.3/examples/django/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+For now these entities are declared explicitly in the `application-template-stibuild.json` file, which is a template we use to create the entities.
```
soltysh: After nuking previous paragraph, just explain what this file contains, maybe a list of objects will be useful.

3.3/examples/django/README.md
```diff
@@ -0,0 +1,91 @@
...
+osc process -f application-template-stibuild.json | osc create -f -
+```
+
+A build of your app should have been triggered automatically. Wait until the build completes with this command:
+
+```bash
+osc get builds --watch
+```
+
+The STI build process essentially copies your source files and install dependencies listed in `requirements.txt`, finally producing a new Docker image.
```
soltysh: Add: For more information about STI build process see https://github.com/openshift/source-to-image/

3.3/examples/django/README.md
```diff
@@ -0,0 +1,91 @@
...
+In the future, creating a new app in OpenShift from a git repo should automatically detect what kind of application it is, and create some entities to support it.
+
+For now these entities are declared explicitly in the `application-template-stibuild.json` file, which is a template we use to create the entities.
+You may want to inspect this file and tweak some configuration. Then, go on and create the entities:
+
+```bash
+osc process -f application-template-stibuild.json | osc create -f -
+```
+
+A build of your app should have been triggered automatically. Wait until the build completes with this command:
```
soltysh: Add: in case one does not start, you can do so invoking `osc start-build....`

3.3/examples/django/README.md
```diff
@@ -0,0 +1,91 @@
...
+You should now the able to see your Django application running. To find out the IP and port of your service, run this:
+
+```bash
+osc describe service frontend
+```
+
+
+### One-off commands ###
+
+The Django app requires the execution of some initial one-off commands in order finish setting it up.
```
soltysh: in order **to** finish

soltysh: Name those initial steps, namely copy the appropriate contents of demo.md here.

3.3/examples/django/cleanup.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+echo "-> removing openshift.local.* directories..."
+sudo rm -rf openshift.local.*
+
+echo "-> stopping all k8s docker containers..."
+docker ps | awk '{ print $NF " " $1 }' | grep ^k8s_ | awk '{print $2}' |  xargs -l -r docker stop
```
soltysh: Remove cleanup, as this is part of openshift stuff, you don't want to mess with this :wink:

3.3/examples/django/demo.md
```diff
@@ -0,0 +1,73 @@
...
+	./run-in-app-container.sh ./manage.py migrate
+	./run-in-app-container.sh ./manage.py createsuperuser
+
+
+## Scale
+
+	# edit first frontend replication controller
+	osc edit rc $(osc get rc -l 'name=frontend-rc' -t '{{ with index}}{{ .metadata.name }}{{ end }}')
+	osc edit rc frontend-1     # change currently running replication controller for immediate scaling
+	osc edit dc frontend       # changing deployment config will affect only next deploy
```
soltysh: I'd merge it into README, users won't have to switch back and forth between this and readme.

3.3/examples/django/demo/settings.py
rhcarvalho: Running OS locally would allow you to use the same containerized infrastructure as you'll have in production. A big asset when your project is more than a single django app with a database.
Think about launching multiple microservices written in different technologies to compose a single system.

And re: env vars, they're good,  but I find it very cumbersome if I had to set 6+ env vars to run my project locally.

3.3/examples/django/README.md
rhcarvalho: Fixed

3.3/examples/django/README.md
rhcarvalho: It's very hard to give pointers to a quick-changing document. I saw in other README's references like "follow steps 1-9", and those numbers are now outdated. That full line is mentioned right in the next section, and IMHO it is long and daunting enough to appear twice in a row.
I opted for briefly mentioning the main things you need to have ready -- in the next line below

3.3/examples/django/README.md
rhcarvalho: Added

3.3/examples/django/README.md
rhcarvalho: re: missing "to", added.
re: initial steps, this is not a Django tutorial. And those steps will change, should change. `./manage.py collectstatic` should be part of the build process, and `./manage.py migrate` part of the deployment process.
These are in the scope of the quickstart I'm working on: https://github.com/rhcarvalho/openshift-django-quickstart

3.3/examples/django/README.md
rhcarvalho: I changed this part to be more concise and point to the documentation about templates.

3.3/examples/django/README.md
rhcarvalho: There's already a link to STI in the first section

3.3/examples/django/cleanup.sh
rhcarvalho: This is useful like `demo.md` is/was, even though they can be outdated quickly

3.3/examples/django/demo.md
rhcarvalho: I left README.md more "informative", while this file was a cheat sheet with copy-and-paste commands for me to use during the live demo.

soltysh: I agree it was useful during demo, but since we're going to maintain this example, I'd rather have one readme file with everything in it than two file that might diverge in the future.

3.3/examples/django/cleanup.sh
soltysh: Since this is part of the maintained image, I'd rather have it up to date more or less and this file isn't required here, the same as with bootstrap.sh you had previously.

3.3/examples/django/README.md
soltysh: OK, it looks better now.

3.3/examples/django/README.md
soltysh: :+1:

3.3/examples/django/cleanup.sh
rhcarvalho: `bootstrap.sh` didn't work... that's why it is gone :smile: 
It is complex to bootstrap reliably via scripting, as can be seen in the `hack/test-end-to-end.sh` script.
So, `bootstrap.sh` was gone and the summarized steps went into `demo.md`.

Cleanup is easier, and I found it convenient to have it here. When I am `cd`'ed in the demo's directory, there's no way I'll be calling the cleanup script from `origin/examples/sample-app`...

Maybe this is a sign that a `openshift ex cleanup` command would be useful?

3.3/examples/django/demo.md
rhcarvalho: If we're merging `README.md` and `demo.md`, then what happens to the first 30 lines of `demo.md`? They're (almost) the commands from sample-app that you didn't want repeated here...

soltysh: Nuke them.

soltysh: The first 30 lines of this file are already outdated, please remote it and we're ready to merge this.

soltysh: And merge it into readme.md.

rhcarvalho: I've updated the `README.md` with useful bits from `demo.md` and removed the latter.

soltysh:
LGTM.

PR #10, opened 2015-05-06
merge env and run commands to reduce layers

bparees:


bparees:
@soltysh i'm baffled by this test failure:
I0506 00:17:19.688018 03818 sti.go:385] Successfully installed gunicorn
  File "/opt/rh/python33/root/usr/lib64/python3.3/os.py", line 673, in __getitem__
    value = self._data[self.encodekey(key)]
KeyError: b'IMAGE_EXPOSE_SERVICES'

can you see any reason that ENV variable isn't being found?  when i build the image locally, i can see it's defined.

bparees:
[test]

soltysh:
@bparees wait till #11 lands and then please rebase your changes on top of that.

soltysh:
@mfojtik since you'll be updating your PR, I'm merging Ben's in the mean time. I've checked it locally and it works. LGTM.

PR #12, opened 2015-05-07
Build support for Django

rhcarvalho:
* Change `.sti/bin/assemble` script to support custom build tasks for Django apps
* Test building Django apps with and without support for serving static files
* Test for expected build failure (for repos missing both `setup.py` and `requirements.txt`)

-------------
General comments:
* Tests are rather simple, but very slow :cry: 
  * Test runtime doubled from about a minute to 2 minutes in my environment.
  * Rebuilding the docker image on every run is too expensive for an iterative workflow.
* No way to run single tests (well, except by commenting lines in the `run` script) -- maybe we lack a better test "framework"?

That said, I'd love to work on improving the tests next time I touch this -- for now, I'm focusing on adding the bits to support a smooth deploy of Django apps with OpenShift.


rhcarvalho:
@soltysh @mfojtik PTAL

rhcarvalho:
I don't understand why the jenkins build failed, could anyone explain the [console output](https://ci.openshift.redhat.com/jenkins/job/sti-python/26/console)?

[test]

3.3/.sti/bin/assemble
```diff
@@ -5,12 +5,42 @@ source .bashrc
...
 
-echo "---> Building your Python application from source"
-if [ -f setup.py ]; then
-    python setup.py install --user
-elif [ -f requirements.txt ]; then
-    pip install --user -r requirements.txt
+# If no requirements given, assume project root is installable
+# i.e., contains a valid `setup.py`.
+if [ ! -f requirements.txt ]; then
+  echo "-e ." > requirements.txt
```
soltysh: I think you've meant here 
```
echo -e "." > requirements.txt
```

3.3/.sti/bin/assemble
```diff
@@ -5,12 +5,42 @@ source .bashrc
...
+echo "---> Copying application source ..."
 cp -Rf /tmp/src/* ./
 
-echo "---> Building your Python application from source"
-if [ -f setup.py ]; then
-    python setup.py install --user
-elif [ -f requirements.txt ]; then
-    pip install --user -r requirements.txt
+# If no requirements given, assume project root is installable
+# i.e., contains a valid `setup.py`.
```
soltysh: That's a false assumption, I'd rather leave the previous part which did the actual check for `setup.py`.

3.3/test/django-test-apps/no-static/run.py
```diff
@@ -0,0 +1,2 @@
+import os
+os.execl('./manage.py', './manage.py', 'runserver', '0.0.0.0:8080')
```
soltysh: I'd rather avoid showing this as an option for users to run django app. Generally speaking using dev server is discouraged, I'd prefer this app be run in gunicorn or differently as a standalone.

soltysh:
@rhcarvalho I think this part should be of your interest, regarding jenkins failure:
```
[INFO] Preparing to test django-test-apps/no-static ...

Initialized empty Git repository in /tmp/image_test.bHIOtov/sti-python/3.3/test/django-test-apps/no-static/.git/
[master (root-commit) 861b9a8] Sample commit
 9 files changed, 213 insertions(+)
 create mode 100644 .gitignore
 create mode 100644 .sti/environment
 create mode 100755 manage.py
 create mode 100644 project/__init__.py
 create mode 100644 project/settings.py
 create mode 100644 project/urls.py
 create mode 100644 project/wsgi.py
 create mode 100644 requirements.txt
 create mode 100644 run.py

[INFO] TEST FAILED (1)
```

3.3/.sti/bin/assemble
rhcarvalho: Nope, it's indeed "-e .", http://pip.readthedocs.org/en/stable/reference/pip_install.html#editable-installs.
Same as what Heroku does. And I've been using that before as well for libs with setup.py and not on PyPI.

3.3/test/run
```diff
@@ -152,8 +153,83 @@ for app in ${WEB_APPS[@]}; do
...
+# Wait for the container to write it's CID file
+wait_for_cid
+
+test_connection
+check_result $?
+
+info "All tests for the ${app} finished successfully."
+cleanup ${app}
+
+
```
soltysh: Instead of copy&pasting the contents of those 2 tests, wouldn't be better to run them in previous for loop, and add conditional check at the end of that for loop. I'm quite reluctant on accepting this solution.

soltysh: Of course the failing case is different and can be kept separately.

soltysh:
@rhcarvalho regarding your comments:

> Tests are rather simple, but very slow :cry:

You're adding 3 new test case, which is more then we currently have, so I wouldn't worry about that.

> No way to run single tests (well, except by commenting lines in the run script) -- maybe we lack a better test "framework"?

I don't mind extending this test framework, remember though we use this in all our images, so you'll have to update that in all of them as well :wink:

3.3/.sti/bin/assemble
rhcarvalho: Are we aiming to support repositories that neither have a `requirements.txt` nor `setup.py`?
I added a test case for trying to build a repo without either `setup.py` or `requirements.txt` (build should fail).

Those files are two good indicators that the repo is "Python-compatible", and we could use that when auto-detecting, for instance, in `osc new-app`.

In current master tip there is no else clause to that if statement. When neither `setup.py` nor `requirements.txt` is present, the `assemble` script will only copy the source code to a new location. And `run` will have only a single option to do it's job: `exec python -u ${APP_FILE}`. How realistic is that that it will launch an HTTP server with a working application?

rhcarvalho:
> I don't mind extending this test framework, remember though we use this in all our images, so you'll have to update that in all of them as well :wink:

I was puzzled when I read about this "framework" in the source code. Is there really any framework, or is it just a pattern followed in all sti-* repos?!

> @rhcarvalho I think this part should be of your interest, regarding jenkins failure: ...

Hmm, seems the build warning was not detected. Need to add a message in that case.

I might work on changes later today or weekend...

soltysh:
> I was puzzled when I read about this "framework" in the source code. Is there really any framework, or is it just a pattern followed in all sti-* repos?!

No framework, we copy this script from image to image. And since it has all we need to to do useful testing we started calling it a framework, because it sounds serious :wink: But as I said before, any improvements to it are more than welcome.

3.3/test/django-test-apps/no-static/run.py
rhcarvalho: IIRC my intention here was to have less dependencies. It's a pain already to run tests because it takes 2~4 min each run.
Probably the time saved for not using `gunicorn` here is negligible, so that it is not a very compelling trade off in that direction.
However, this does add some value by testing the `.sti/environment:APP_FILE` way of running the app. @soltysh what do you think about that?

3.3/.sti/bin/assemble
rhcarvalho: I've remove the two commits about always using pip and requiring either `setup.py` or `requirements.txt` -- we can discuss that in another moment.

rhcarvalho:
Needs to have `find` available in the system, waiting for https://github.com/openshift/sti-base/pull/25

rhcarvalho:
[test]

3.3/test/django-test-apps/no-static/run.py
soltysh: For your local dev, to speed up the process you can do whatever you feel comfortable with, but in the repo I'd prefer to avoid such things. The additional 2 mins for running tests are not a problem, and we should evangelize users to run applications inside of OpenShift correctly, rather than going the easy path :wink:

soltysh:
[test]

rhcarvalho:
@soltysh @bparees PTAL

3.3/.sti/bin/assemble
```diff
@@ -1,23 +1,66 @@
...
+  set +e
+  django_collectstatic
+  set -e
+
+  # remove pip temporary directory
+  rm -rf /tmp/pip_build_default
+}
+
+function django_collectstatic() {
+  [ -n "$DISABLE_COLLECTSTATIC" ] && return
```
soltysh: If you're introducing new variable, please update README.md and don't forget to update openshift-docs as well. We want to have all those places up to date.

3.3/.sti/bin/assemble
```diff
@@ -1,23 +1,66 @@
...
 
-if [ -f setup.py ]; then
-    echo "---> Installing application ..."
-    python setup.py install --user
-elif [ -f requirements.txt ]; then
-    echo "---> Installing dependencies ..."
-    pip install --user -r requirements.txt
-fi
+  echo "---> Copying application source ..."
+  cp -Rf /tmp/src/* ./
```
mfojtik: this will not copy hidden files/folders.

3.3/.sti/bin/assemble
```diff
@@ -1,23 +1,66 @@
...
-# For SCL enablement
-source .bashrc
+# This script uses these environment variables:
+#
+# DISABLE_COLLECTSTATIC: if not empty, inhibits execution of 'manage.py collectstatic'.
 
-set -e
+function assemble() {
+  # For SCL enablement
+  source .bashrc
```
mfojtik: this should be outside function

3.3/.sti/bin/assemble
```diff
@@ -1,23 +1,66 @@
...
-rm -rf /tmp/pip_build_default
+  if [ -f setup.py ]; then
+      echo "---> Installing application ..."
+      python setup.py install --user
+  fi
+
+  # set permissions for any installed artifacts
+  chmod -R og+rwx /opt/openshift
+
+  set +e
```
mfojtik: set +e/-e should be part of that function

3.3/.sti/bin/run
```diff
@@ -1,25 +1,112 @@
...
+  django_migrate "$MANAGE_FILE"
+
+  # try to use gunicorn
+  run_gunicorn "$APP_MODULE"
+  # or fallback to Django's development server
+  django_runserver "$MANAGE_FILE"
+}
+
+function django_migrate() {
+  [ -n "$DISABLE_MIGRATE" ] && return
```
soltysh: Same here -> docs update please!

3.3/.sti/bin/run
```diff
@@ -1,25 +1,112 @@
...
+#             If not provided, tries to find the Python path to a 'wsgi.py' file
+#             in the source tree.
+#             That file is present in Django projects by default.
+# APP_CONFIG: Optional configuration file for gunicorn.
+# APP_FILE:   Optional path to Python script to run your application.
+#             Defaults to 'app.py' if it exists.
+#
+# DISABLE_MIGRATE: if not empty, inhibits execution of 'manage.py migrate'.
+
+BIND_ADDR="0.0.0.0:8080"
```
mfojtik: we will always bind on this address, no need to make in env var

3.3/.sti/bin/run
```diff
@@ -1,25 +1,112 @@
...
+
+function run() {
+  # For SCL enablement
+  source .bashrc
```
mfojtik: move this to top of this script

3.3/.sti/bin/run
```diff
@@ -1,25 +1,112 @@
...
+
+function django_migrate() {
+  [ -n "$DISABLE_MIGRATE" ] && return
+  echo "---> Migrating database ..."
+  django_manage_cmd "$1" migrate --noinput
+}
+
+function django_runserver() {
+  echo "---> Serving application with 'manage.py runserver' ..."
+  echo "---> WARNING: this is NOT a recommended way to run you application in production!"
```
soltysh: s/NOT a recommended/NOT the recommended

3.3/.sti/bin/run
```diff
@@ -1,25 +1,112 @@
...
+function run() {
+  # For SCL enablement
+  source .bashrc
+
+  set -e
+
+  # Try to run application with the strategies below, in precedence order.
+  # The first successful strategy takes over this script's process via exec.
+  # If no strategy succeed, report an error message and terminate.
+  run_django
```
mfojtik: instead of this I would prefer `if`.... this way it looks that it is going to run all of these and it is very hard to read

3.3/.sti/bin/run
```diff
@@ -1,25 +1,112 @@
...
+}
 
-set -e
+function django_manage_cmd() {
+  local MANAGE_FILE="$1"
+  local CMD="${@:2}"
 
-export APP_FILE=${APP_FILE:-"app.py"}
+  if [ ! -f "$MANAGE_FILE" ]; then
+    echo "WARNING: seems that you're using Django, but we could not find a 'manage.py' file."
```
soltysh: find ~~a~~ 'manage...

3.3/.sti/bin/run
```diff
@@ -1,25 +1,112 @@
...
+  return 1
 }
 
-# For SCL enablement
-source .bashrc
+function run_django() {
+  ! pip show -q django && return
+
+  # Find shallowest manage.py script, either ./manage.py or <project>/manage.py
+  local MANAGE_FILE=$(find . -maxdepth 2 -type f -name 'manage.py' -printf '%d\t%P\n' | sort -nk1 | cut -f2 | head -1)
```
mfojtik: local vars should be lower-case

3.3/test/run
```diff
@@ -1,51 +1,62 @@
...
-# TODO: Make command compatible for Mac users
-test_dir="$(readlink -zf $(dirname "${BASH_SOURCE[0]}"))"
-image_dir=$(readlink -zf ${test_dir}/..)
+function run() {
+  test_dir=$(readlink -m $(dirname "${BASH_SOURCE}"))
 
-# TODO: This should be part of the image metadata
-test_port=8080
+  # TODO: This should be part of the image metadata
+  test_port=8080
```
mfojtik: extra space?

mfojtik: remove this TODO... there are no metadata for this and the port will be 8080

3.3/.sti/bin/assemble
soltysh: Yes, please update it to match what we have in [sti-php](https://github.com/openshift/sti-php/blob/master/5.5/.sti/bin/assemble#L8)

3.3/.sti/bin/run
```diff
@@ -1,25 +1,112 @@
...
+#             That file is present in Django projects by default.
+# APP_CONFIG: Optional configuration file for gunicorn.
+# APP_FILE:   Optional path to Python script to run your application.
+#             Defaults to 'app.py' if it exists.
+#
+# DISABLE_MIGRATE: if not empty, inhibits execution of 'manage.py migrate'.
+
+BIND_ADDR="0.0.0.0:8080"
+
+function run() {
```
soltysh: Can we instead of creating this function have its contents at the end of this file, to make it a little bit more readable. I don't see the need for this function.

3.3/.sti/bin/assemble
```diff
@@ -1,23 +1,66 @@
 #!/bin/bash
 
-# For SCL enablement
-source .bashrc
+# This script uses these environment variables:
+#
+# DISABLE_COLLECTSTATIC: if not empty, inhibits execution of 'manage.py collectstatic'.
 
-set -e
+function assemble() {
```
soltysh: The same as with run, move the entire contents of this function to the end of this file. I'm not sure about the need for that function.

3.3/.sti/bin/run
bparees: the "a" is appropriate here.

3.3/.sti/bin/assemble
rhcarvalho: My mistake... inheritance from old starting point and imperfect rebase. Fixed.

3.3/.sti/bin/assemble
rhcarvalho: Why?

mfojtik: you want to have this globally enabled (ideally we should never be doing this), also sourcing files in functions is bad practice because it makes it hard to test, etc... also if you call this function twice, you will enable SCL twice... 

personally I would like to break the assemble function into smaller functions and not call it 'assemble'.

soltysh: >personally I would like to break the assemble function into smaller functions and not call it 'assemble'.

Let's not over-complicate it that much, the functions @rhcarvalho created are readable. If we split it that will be a paint to maintain. It's not that big.

3.3/.sti/bin/assemble
rhcarvalho: It wouldn't be right to have it as part  of `django_collectstatic`.
The reasoning here is to avoid side effects, keep them in as few places as possible. `set +e; set -e` change the behavior of the current shell, if I put it inside `django_collectstatic`, after calling that function the caller will inadvertently have the '-e' flag enabled... even if it was not set before.

If I knew how to do it, I'd save the state of the shell flags in the beginning of `django_collectstatic`, do `set +e`, and later set the flags back to the original value.

3.3/.sti/bin/run
rhcarvalho: This is not an env var, it is a global variable, used by Django's `runserver` and `gunicorn`.

3.3/.sti/bin/run
rhcarvalho: Hmmm I don't see how your proposal makes it more readable. @soltysh do you read the files from the bottom up?!

soltysh: That's an old habbit from C :wink: besides we've been practicing that in our bash scripts for most of the time.

3.3/.sti/bin/assemble
rhcarvalho: @mfojtik 
> you want to have this globally enabled

For both scripts, the way I thought about it, is that the `assemble` function is the entry point of this script. I wanted to read from top to bottom and be able to understand and explain what is going to happen all the time without jumping in the file -- trying to get us a little closer to literate programming? Instead of having some commands, then function declarations and then more commands, and have to read everything and jump top-bottom-top to understand what is going on.

The call in or outside the function has the same "global" effect, as it affects the current shell.

But I understand perhaps I should accommodate to whatever style and tradition is already in place...

> sourcing files in functions is bad practice because it makes it hard to test

Agree, for all functions that are not `main`... maybe I should call it `main()` and not `assemble()`?! :smile: 
I'd avoid `source`'ing inside a function just like avoid changing shell options with `set`.

> also if you call this function twice, you will enable SCL twice...

`source scl_source python33` is idempotent, calling it multiple times makes no harm (won't duplicate entries in PATH, etc).

mfojtik: @rhcarvalho bash scripts are not intended for programming and we should not bend bash to look as programming language. The bash scripts should just run commands in order to automate administrator / developer tasks but not in programmatic way. Bash functions are nice but they should be used wisely (not as assemble() or main()).  

Bash is super hard to test (or unit test) and having monolithic functions just makes the problem worse. Ideally we should not be creating a 'SDK' or bash function libraries (we went that path in v2 and it was a mess).

3.3/.sti/bin/run
mfojtik: then I would move it down to place where you call those. having it on top makes it look as something we set and allow to customize (or we expect to change that at some point, which is not valid)

3.3/.sti/bin/assemble
mfojtik: why is this hidden in function? if this variable is set that function should never be called

3.3/.sti/bin/assemble
mfojtik: @soltysh readable? so having an 'assemble' script with 'assemble()' function is readable? I think the script itself is named assemble, why we should have function inside? Why it will be pain to maintain? Having easy to read and understand bash function that are doing exactly what they should do based on their name, without too much logic hidden inside is from my PoV much easier to maintain that having one big monolithic assemble() function.

3.3/.sti/bin/assemble
```diff
@@ -1,23 +1,66 @@
...
+  # remove pip temporary directory
+  rm -rf /tmp/pip_build_default
+}
+
+function django_collectstatic() {
+  [ -n "$DISABLE_COLLECTSTATIC" ] && return
+  ! pip show -q django && return
+
+  # Find shallowest manage.py script, either ./manage.py or <project>/manage.py
+  local manage_py=$(find . -maxdepth 2 -type f -name 'manage.py' -printf '%d\t%P\n' | sort -nk1 | cut -f2 | head -1)
```
mfojtik: manage_script? manage_file?

3.3/.sti/bin/assemble
rhcarvalho: Not trying to make an SDK, just make it easier (at least for me) to understand, especially for the lack of tests and the time it takes to run the ones with have for sti scripts.
I'm totally open to whatever format we'll have prior to merging this. Going through the comments one by one. Will continue on it tomorrow.

3.3/.sti/bin/assemble
```diff
@@ -1,23 +1,66 @@
...
+
+  if [ ! -f "$manage_py" ]; then
+    echo "WARNING: seems that you're using Django, but we could not find a 'manage.py' file."
+    echo "'manage.py collectstatic' ignored."
+    return
+  fi
+
+  echo "---> Collecting Django static files ..."
+  if ! python $manage_py collectstatic --dry-run --noinput &> /dev/null; then
```
mfojtik: why we redirect all output to /dev/null? if this fails, all I can do is to run it manually? can't we just print the errors?

3.3/.sti/bin/assemble
```diff
@@ -1,23 +1,66 @@
...
+    echo "WARNING: could not run 'manage.py collectstatic'. To debug, run:"
+    echo "    $ python $manage_py collectstatic --noinput"
+    echo "Ignore this warning if you're not serving static files with Django."
+    return
+  fi
+
+  if ! python $manage_py collectstatic --noinput  2>&1 | sed '/^Copying/d;/^$/d;/^ /d'; then
+    local status=$?
```
mfojtik: why you need this variable?

3.3/.sti/bin/run
mfojtik: get rid of run() and assemble()... we already have scripts with same name, I don't see any reason why we should also have functions with same name.

3.3/.sti/bin/run
```diff
@@ -1,25 +1,112 @@
...
+  set -e
+
+  # Try to run application with the strategies below, in precedence order.
+  # The first successful strategy takes over this script's process via exec.
+  # If no strategy succeed, report an error message and terminate.
+  run_django
+  run_gunicorn "$APP_MODULE"
+  run_python_script "${APP_FILE:-app.py}"
+
+  echo "ERROR: don't know how to run your application."
```
mfojtik: technically, this should go to stderr

3.3/.sti/bin/run
rhcarvalho: I might be wrong, but the way I understand "not the recommended" implies there is one true recommended way, which is not the case, since Django list [many ways](https://docs.djangoproject.com/en/1.8/howto/deployment/) to deploy your app.

Using gunicorn is "one of the recommended ways", using runserver is "not one of the recommended ways", thus "not a recommended..."

3.3/.sti/bin/run
```diff
@@ -1,25 +1,112 @@
...
+  echo "Please set either APP_MODULE or APP_FILE environment variables, or create a file 'app.py' to launch your application."
+  return 1
 }
 
-# For SCL enablement
-source .bashrc
+function run_django() {
```
mfojtik: this should be done in this way:

```sh
if has_migrations; then
  run_django_migrate "${manage_script}"
fi
if has_gunicorn; then
  run_gunicorn "${APP_MODULE}"
else
  run_django_server "${manage_script}"
fi
```

If the run_django_server and run_unicorn are just 2 lines I would nuke that functions.... we should move the **detection** to functions not the actual server run.

3.3/test/run
rhcarvalho: it's been there before... gone.

3.3/.sti/bin/run
```diff
@@ -1,25 +1,112 @@
...
+}
+
+function django_runserver() {
+  echo "---> Serving application with 'manage.py runserver' ..."
+  echo "---> WARNING: this is NOT a recommended way to run you application in production!"
+  django_manage_cmd "$1" runserver "$BIND_ADDR"
+}
 
-set -e
+function django_manage_cmd() {
```
mfojtik: this require documentation... it has too much hidden functionality inside... if the command starts from *runserver then this will exit after run, otherwise just run and continue...

3.3/.sti/bin/run
```diff
@@ -1,25 +1,112 @@
...
+
+function run_gunicorn() {
+  ! pip show -q gunicorn && return
+
+  local app_module="$1"
+
+  if [ -z "$app_module" ]; then
+    # Find shallowest wsgi.py file, one of ./wsgi.py, <project>/wsgi.py or <project>/<project>/wsgi.py,
+    # replace "/" with "." and remove ".py" suffix
+    app_module=$(find . -maxdepth 3 -type f -name 'wsgi.py' -printf '%d\t%P\n' | sort -nk1 | cut -f2 | head -1 | sed 's:/:.:;s:.py$::')
```
mfojtik: we should move this to bash function that returns the app_module.

3.3/.sti/bin/assemble
rhcarvalho: I read it as "manage py", because that's always the name of it...

-- "python manage py do something"

3.3/.sti/bin/assemble
soltysh: @mfojtik I'm saying it's not readable having an assemble function inside assemble script. I'm saying that having functions that group certain functionality is ok and the kind of "main" (which should not be a function) at the end which is showing nice flow of the script using those previously defined functions.

3.3/.sti/bin/assemble
rhcarvalho: Look below... this is a dry run to check "collectability", only then we actually run it.
If you don't use static files, this command might always fail depending on your settings, and it doesn't mean the build should fail.

If you don't see your static files on the browser, you will see the message below in your build log, and from there it is better to run the command below on your local development until you've configured your project correctly.

What you do want in the build logs is the output of the second run (without the --dry-run flag).

Disclaimer: this is exactly what you get in Heroku.

mfojtik: @rhcarvalho all I want is >/dev/null and show the error messages

3.3/.sti/bin/assemble
mfojtik: @soltysh in bash, you should never write functions to show 'the flow' because everyone that will try to understand it will have to go to that functions anyway... if you get problem running gunicorn/django server... understanding the flow will not fix your problem...

we should explain the 'flow' in documentation, not inside bash scripts... people will not go read them to understand the flow but to fix problems

3.3/.sti/bin/assemble
mfojtik: @rhcarvalho that variable holds the path to that script right? so to be explicit, it should be manage_py_path... i don't like that '_py' part... it just looks weird in bash. i would prefer that variable name to be something that express the intent of it... provide a path to manage.py and not the content of manage.py itself.

3.3/.sti/bin/assemble
soltysh: Makes sense.

3.3/.sti/bin/assemble
soltysh: I must have missed that, what you guys decided about assemble and run functions. I thought they will be placed in-line at the end of this file, to get rid of unnecessary functions.

mfojtik: @soltysh no more assemble() or run()

soltysh: :8ball:

README.md
```diff
@@ -184,16 +184,91 @@ file inside your source code repository.
...
+  This file is present in Django projects by default.
+
+  If you have both Django and Gunicorn in your requirements, your Django project
+  will be automatically served with Gunicorn.
+
+* **Django development server**
+
+  If you have Django in your requirements, but don't have Gunicorn, then your
+  application will be served with Django's development web server. This is not,
+  however, a recommended way to serve your application in production.
```
soltysh: Make the last sentence about not recommended way of running up bold, to stress this. Still though I'm wondering if we want to write about this...

README.md
```diff
@@ -184,16 +184,91 @@ file inside your source code repository.
...
+
+* **Python script**
+
+  This is the most general way of executing your application. It will be used
```
soltysh: s/most general/simplest

soltysh:
Two small nits, but generally LGTM. Squash and rebase probably after #25 lands. I'll merge once its ready.

soltysh:
Thanks @rhcarvalho !

PR #13, opened 2015-05-07
Support running python images with a random UID

csrwng:


csrwng:
@bparees @soltysh ptal

bparees:
lgtm but would like @soltysh opinion.

3.3/.sti/bin/assemble
```diff
@@ -14,3 +14,12 @@ if [ -f setup.py ]; then
...
 fi
+
+# set permissions for any installed artifacts
+chmod -R og+rwx /opt/openshift
+
+# remove pip temporary directory
+pip_tmp=/tmp/pip_build_default
+if [ -d $pip_tmp ]; then
+    rm -rf $pip_tmp
+fi
```
soltysh: Why this is needed?

soltysh:
LGTM, just one question, I'll wait with merging to get the answer for that.

3.3/.sti/bin/assemble
csrwng: When the application starts, under a different user, pip complains that the /tmp/pip_build_default directory already exists and is not owned by the current user. The app seems to keep on working... it will serve html, but I didn't want potentially other function to be broken because of it.

rhcarvalho: Why do we test for directory existence if we do `rm -rf`?

```bash
$ rm -rf /tmp/nothing/is/here
$ echo $?
0
```

soltysh: Right, you can fix that in your PR @rhcarvalho.

3.3/test/run
```diff
@@ -84,7 +89,8 @@ wait_for_cid() {
   local sleep_time=1
   local attempt=1
   local result=1
-  info "Waiting for application container to start"
+  local cargs=${CONTAINER_ARGS:-}
+  info "Waiting for application container to start $cargs"
```
rhcarvalho: Hmm... I'm wondering why do we need this `local cargs`...

If we had somewhere `set -u`, trying to expand `$CONTAINER_ARGS` when it is not set would be an error. But since we don't, what would be the difference if we simply had `info "Waiting for application container to start $CONTAINER_ARGS"`?

@csrwng, what's the reason behind this? Maybe I can learn something :smile:

csrwng: @rhcarvalho - actually now that I look at it, no purpose. I needed it here: https://github.com/openshift/sti-python/pull/13/files#diff-4d94a409b126ccb3db3b109be441a31cR57 and I ended up using the same thing everywhere else. You're welcome to submit a fix to remove it :-)

rhcarvalho: Same question then, why is it needed there?! How about `docker run $CONTAINER_ARGS --rm --cidfile=$cid_file -p $test_port ${IMAGE_NAME}-testapp`?

csrwng: That one was giving me an error

PR #29, opened 2015-06-11
Move SCL enable to /opt/openshift/etc/scl_enable

mfojtik:
See: https://github.com/openshift/sti-base/issues/38


3.3/test/run
```diff
@@ -106,6 +106,43 @@ test_docker_run_usage() {
...
+  local cid_file="$3"
+
+  info "Testing the image SCL enable"
+  out=$(docker run --rm ${IMAGE_NAME} /bin/bash -c "${run_cmd}" 2>&1)
+  if ! echo "${out}" | grep -q "${expected}"; then
+    echo "ERROR[/bin/bash -c "${run_cmd}"] Expected '${expected}', got '${out}'"
+    return 1
+  fi
+  out=$(docker run --rm -i ${IMAGE_NAME} /bin/bash -i <<< "${run_cmd}" 2>&1)
+  if ! echo "${out}" | grep -q "${expected}"; then
```
rhcarvalho: Seems like a good use case for `[[ "$out" =~ "^$expected" ]]`?

mfojtik: perl gives you multi line output and exec do some crazy stuff with ioctl so I would rather keep it with grep as that works ;-) we can fix as follow up or when we will be reworking the tests in post-3.0

3.3/.sti/bin/assemble
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+source /opt/openshift/etc/generate_container_user
```
mfojtik: @csrwng PTAL

3.3/contrib/etc/generate_container_user
```diff
@@ -0,0 +1,8 @@
+# Set current user in nss_wrapper
+PASSWD_DIR="/opt/openshift/etc"
```
mfojtik: @csrwng I moved `passwd` out of `./src` because that directory is used for the sources ;-) If you have 'passwd' file in your GIT repository then you can override the one we generate ;-)

mnagy:
LGTM

3.3/contrib/etc/generate_container_user
csrwng: makes sense

3.3/.sti/bin/assemble
csrwng: I don't think we need it for assemble... only for run. When running assemble we just use the user set in the Dockerfile.

mfojtik: @csrwng thanks, I will remove it from assemble (it seems to be harmless as the test is passing)

csrwng: Yup, it is. Just not needed.

PR #34, opened 2015-06-12
Use pip 1.5.6

rhcarvalho:
Second try...

### DIFF
--- a/3.3/.sti/bin/run
+++ b/3.3/.sti/bin/run
@@ -2,11 +2,11 @@
 source /opt/openshift/etc/generate_container_user
 
 function is_gunicorn_installed() {
-  pip show -q gunicorn
+  hash gunicorn &>/dev/null
 }
 
 function is_django_installed() {
-  pip show -q django
+  python -c "import django" &>/dev/null
 }
 
 function should_migrate() {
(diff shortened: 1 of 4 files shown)


mfojtik:
[test]

mfojtik:
@bparees @rhcarvalho is looking into this issue. In worst case I think we can live with more latest 'pip' in centos. We already have that situation with nodejs (the RHEL has 0.10.25 and Centos7 has 0.10.35).

rhcarvalho:
Looks like it was just the problem with `pip show`. Now we do not use it anymore.
@mfojtik @bparees PTAL.

mfojtik:
@rhcarvalho test passed, which is good. @bparees I will leave this merge to you :-)

3.3/.sti/bin/run
```diff
@@ -2,11 +2,11 @@
 source /opt/openshift/etc/generate_container_user
 
 function is_gunicorn_installed() {
-  pip show -q gunicorn
+  hash gunicorn &>/dev/null
```
bparees: I think "which" is the more traditional test.

rhcarvalho: http://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script/677212#677212

`hash` is a bash builtin and has its advantages over `which`. Agree with you, `which` was the first thing in my mind too.

bparees: interesting, thanks for the education!

PR #35, opened 2015-06-12
Using RHEL packages for nss_wrapper and pip

brenton:


brenton:
@mfojtik, @bparees PTAL

bparees:
thanks @brenton

PR #43, opened 2015-06-30
Fix labels and remove EPEL from rhel7 image

mfojtik:


mfojtik:
@bparees @soltysh PTAL

@brenton can you please confirm that nss_wrapper is available in OSE channel and we don't need to install EPEL anymore?

mfojtik:
@brenton @bparees currently the jenkins is failing because it is trying to install `pip` from `li-rhscl` and the `nss_wrapper` from EPEL (I think the EPEL is enabled on the Jenkins host machine, but the GPG key is not imported in container... which seems like a RHEL Docker bug ;-))

However, both packages should be installed from our channel.

mfojtik:
(just fyi, the EPEL bug reported here: https://bugzilla.redhat.com/show_bug.cgi?id=1237057)

brenton:
nss_wrapper is definitely in the OSE channels now.  Originally I had all the required packages pushed to the li-rhscl repo before they were available in the product.

soltysh:
@mfojtik what about this PR?

mfojtik:
@soltysh was pending packages available in channels.

soltysh:
According to @brenton they are, let us see:
[test]

mfojtik:
[test]

bparees:
conflict.

Ben Parees | OpenShift

mfojtik:
@bparees rebased... but want to see the tests pass because the fixing this conflict gives me more gray hairs...

bparees:
Remove the / labels from rhel

mfojtik:
@bparees I found more elegant solution for disabling unwanted repositories:

```console
[root@ip-10-45-9-83 ~]# yum-config-manager --disable epel >/dev/null
[root@ip-10-45-9-83 ~]# echo $?
0
[root@ip-10-45-9-83 ~]# yum-config-manager --disable foo >/dev/null 
[root@ip-10-45-9-83 ~]# echo $?
0
```

3.3/Dockerfile.rhel7
```diff
@@ -29,16 +35,10 @@ COPY ./.sti/bin/ /usr/local/sti
...
-    rpm -Uvh epel-release-7-5.noarch.rpm && rm epel-release-7-5.noarch.rpm && \
-    yum install -y --disablerepo=epel --setopt=tsflags=nodocs nss_wrapper && \
-    yum clean all -y && \
-    yum install -y python33-python-pip && \
-    chown -R default:default /opt/app-root && \
-    chmod -R og+rwx /opt/app-root
+# In order to drop the root user, we have to make some directories world
+# writeable as OpenShift default security model is to run the container under
+# random UID.
+RUN chown -R 1001:1001 /opt/app-root
```
bparees: shouldn't this be a chmod og+rw?

3.3/Dockerfile
```diff
@@ -22,15 +27,8 @@ COPY ./.sti/bin/ /usr/local/sti
...
-    https://www.softwarecollections.org/en/scls/rhscl/python33/epel-7-x86_64/download/rhscl-python33-epel-7-x86_64.noarch.rpm && \
-    yum install -y --setopt=tsflags=nodocs --enablerepo=centosplus python33 python33-python-devel python33-python-setuptools epel-release && \
-    yum install -y --setopt=tsflags=nodocs install nss_wrapper && \
-    yum clean all -y && \
-    # FIXME remove this line when pip will be part of python33 package
-    scl enable python33 "easy_install pip==1.5.6" && \
-    chown -R default:default /opt/app-root && \
-    chmod -R og+rwx /opt/app-root
+RUN scl enable python33 "easy_install pip==1.5.6" && \
+    chown -R 1001:1001 /opt/app-root
```
bparees: chmod og+rw?

3.3/Dockerfile.rhel7
mfojtik: it must sneaked out when I was fixing the conflict... good catch

mfojtik:
@bparees merge at your will

3.3/Dockerfile.rhel7
```diff
@@ -10,10 +10,10 @@ EXPOSE 8080
...
     PATH=.local/bin/:$PATH
 
-LABEL k8s.io/description="Platform for building and running Python 3.3 applications" \
-      k8s.io/display-name="Python 3.3" \
-      openshift.io/expose-services="8080:http" \
-      openshift.io/tags="builder,python,python33"
+LABEL io.k8s.description="Platform for building and running PHP 5.5 applications" \
+      io.k8s.display-name="Apache 2.4 with PHP 5.5" \
+      io.openshift.expose-services="8080:http" \
+      io.openshift.tags="builder,php,php55"
```
bparees: php?

bparees:
rhel labels are wrong.

PR #62, opened 2015-09-04
Setting groupID via env var

jhadvig:
@mfojtik PTAL


jhadvig:
[test]

jhadvig:
re-[test]

jhadvig:
@mfojtik @bparees here as well

PR #65, opened 2015-09-21
In run, check for app.py first

### LINKED ISSUE #56: Priority order for how to run application is wrong.
In the 'run' script, the code:

```
APP_FILE="${APP_FILE:-app.py}"
if [[ "$APP_FILE" ]]; then
  if [[ -f "$APP_FILE" ]]; then
    echo "---> Running application from Python script ($APP_FILE) ..."
    exec python "$APP_FILE"
  fi
  echo "WARNING: file '$APP_FILE' not found."
fi
```

should not be last.

Instead, it should be placed before the check for running a WSGI application with gunicorn.

This will mirror better how OpenShift 2 worked where the presence of ``app.py`` would override the default Apache/mod_wsgi server.

It is also necessary to have a check for ``app.py`` before any check for Django and trying to run Django with ``manage.py`` with ``runserver``. This is so that ``app.py`` with a customer WSGI server will still take precedence for a Django application.

Reading of the code as it stands suggests that even if ``app.py`` is supplied to use an alternate WSGI server, when using Django, it will be ignored and the Django development server always used instead.

sYnfo:
Resolves #56

### DIFF
--- a/2.7/s2i/bin/run
+++ b/2.7/s2i/bin/run
@@ -15,6 +15,13 @@ function should_migrate() {
 
 set -e
 
+APP_FILE="${APP_FILE:-app.py}"
+if [[ -f "$APP_FILE" ]]; then
+  echo "---> Running application from Python script ($APP_FILE) ..."
+  exec python "$APP_FILE"
+fi
+echo "WARNING: file '$APP_FILE' not found."
+
 # Find shallowest manage.py script, either ./manage.py or <project>/manage.py
 manage_file=$(find . -maxdepth 2 -type f -name 'manage.py' -printf '%d\t%P\n' | sort -nk1 | cut -f2 | head -1)
 
@@ -55,15 +62,6 @@ if is_django_installed; then
     echo "WARNING: seems that you're using Django, but we could not find a 'manage.py' file."
     echo "Skipped 'python manage.py runserver'."
   fi
-fi
-
-APP_FILE="${APP_FILE:-app.py}"
-if [[ "$APP_FILE" ]]; then
-  if [[ -f "$APP_FILE" ]]; then
-    echo "---> Running application from Python script ($APP_FILE) ..."
-    exec python "$APP_FILE"
-  fi
-  echo "WARNING: file '$APP_FILE' not found."
 fi
 
 >&2 echo "ERROR: don't know how to run your application."
--- a/3.3/s2i/bin/run
+++ b/3.3/s2i/bin/run
@@ -15,6 +15,13 @@ function should_migrate() {
 
 set -e
 
+APP_FILE="${APP_FILE:-app.py}"
+if [[ -f "$APP_FILE" ]]; then
+  echo "---> Running application from Python script ($APP_FILE) ..."
+  exec python "$APP_FILE"
+fi
+echo "WARNING: file '$APP_FILE' not found."
+
 # Find shallowest manage.py script, either ./manage.py or <project>/manage.py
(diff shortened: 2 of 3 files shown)


soltysh:
[test]

2.7/s2i/bin/run
```diff
@@ -15,6 +15,15 @@ function should_migrate() {
 
 set -e
 
+APP_FILE="${APP_FILE:-app.py}"
+if [[ "$APP_FILE" ]]; then
```
soltysh: This will never be false, since you're defaulting it one line before, credits to @bparees.
The same applies to all other python versions.

bparees:
pending fixing the unnecessary if-check this lgtm.

mfojtik:
LGTM as well.

soltysh:
I wouldn't say remove unnecessary if, I'd rather remove the defaulting, because we still want to check other options, if they exists.

soltysh:
Ignore my last comment, I was thinking about something totally different :wink:

bparees:
@sYnfo bump

sYnfo:
@bparees bumped

bparees:
@sYnfo can you squash the commits?

sYnfo:
@bparees squashed!

bparees:
[test]

soltysh:
re-[test]

bparees:
I haven't had a chance to look at it but it seems this change is clearly breaking the tests.

soltysh:
The base image sets scripts-url to [/usr/libexec/s2i](https://github.com/openshift/sti-base/blob/master/Dockerfile#L8) you need to update Dockerfile (all of them!) to copy s2i scripts to that location as well. I don't know why it wasn't done yet. You should be using `$STI_SCRIPTS_PATH` as the destination for those. Additionally don't forget to change the `CMD` in the last line to also point to that location. And please run the tests locally (`make test`, be sure to have s2i in path) to make sure the PR is working before submitting it here.

bparees:
@soltysh huh? The dockerfiles are using that env variable already.

bparees:
This pull probably needs to be rebased, our test logic may not be smart enough to merge the change into master and instead is just testing this branch. I don't recall at the moment.

soltysh:
Right... then for some reason it's being applied wrongly and should be rebased if so. @sYnfo can you rebase your pr, please.

bparees:
fyi yes it appears we just check out the actual PR branch/commit and do not merge it into master prior to testing:
```
# Fetch refs from Github pull requests
git fetch --quiet --tags --progress https://github.com/openshift/sti-python +refs/pull/*:refs/remotes/origin/pr/*

# switch to the desired ref
git checkout 44157475cbc2a8d725332acaa29c3f5ef2a689cd
```

@mfojtik fyi if you're looking for another itch to scratch :)

in the meantime i've cherry-picked this and am testing it locally.

bparees:
my local test passed, i'm just going to merge this and cross my fingers.  if it breaks in the push job i'll revert it.

2.7/s2i/bin/run
```diff
@@ -15,6 +15,13 @@ function should_migrate() {
 
 set -e
 
+APP_FILE="${APP_FILE:-app.py}"
+if [[ -f "$APP_FILE" ]]; then
+  echo "---> Running application from Python script ($APP_FILE) ..."
+  exec python "$APP_FILE"
+fi
+echo "WARNING: file '$APP_FILE' not found."
```
rhcarvalho: I think we don't need this warning in every run of this script... It might make users think that app.py is required.

PR #66, opened 2015-09-21
PATH uses absolute path to per user install area

### LINKED ISSUE #57: PATH to per user Python install area should not be relative.
The Dockerfile's contain:

```
ENV PYTHON_VERSION=2.7 \
	    PATH=.local/bin/:$PATH
```

Thus a relative path and not an absolute path is used to the per user Python install area used when running:

```
pip install --user -r requirements.txt
```

Presumably that relative path for the bind directory would become invalid if the web application being run changed the working directory to another directory. The consequence then would be that the web application would not then be able to find any programs which were installed as part of the Python modules installed by ``pip`` from the ``requirements.txt`` file.

Any directories in PATH should be absolute.

It should therefore use:

```
ENV PYTHON_VERSION=2.7 \
	    PATH=$HOME/.local/bin/:$PATH
```

sYnfo:
Resolves #57

### DIFF
--- a/2.7/Dockerfile
+++ b/2.7/Dockerfile
@@ -8,7 +8,7 @@ MAINTAINER SoftwareCollections.org <sclorg@redhat.com>
 EXPOSE 8080
 
 ENV PYTHON_VERSION=2.7 \
-    PATH=.local/bin/:$PATH
+    PATH=$HOME/.local/bin/:$PATH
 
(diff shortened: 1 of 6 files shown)


soltysh:
[test]

rhcarvalho:
LGTM

soltysh:
LGTM

PR #69, opened 2015-10-05
Add a Hot Deploy paragraph into README.md

mnagy:
@jhadvig please review?

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,3 +273,24 @@ following ways, in precedence order:
   in the case where you specify a path to a Python script via the `APP_FILE` environment
   variable, defaulting to a file named `app.py` if it exists. The script is
   passed to a regular Python interpreter to launch your application.
+
+Hot deploy
+---------------------
+
+If you are using Django, hot deploy will work out of the box.
+
+To enable hot deploy while using Gunicorn, make sure you have a Gunicorn
+configuration file inside your repository with the
+[`reload`](https://gunicorn-docs.readthedocs.org/en/latest/settings.html#reload)
+option set to `true`. Make sure to specify your config via the `APP_CONFIG`
+environment variable.
+
+To change your source code in running container, use Docker's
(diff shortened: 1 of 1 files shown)


mnagy:
The test failure looks like a flake. It obviously couldn't have been caused by this PR, since it's only modifying a README.md

README.md
```diff
@@ -273,3 +273,24 @@ following ways, in precedence order:
   in the case where you specify a path to a Python script via the `APP_FILE` environment
   variable, defaulting to a file named `app.py` if it exists. The script is
   passed to a regular Python interpreter to launch your application.
+
+Hot deploy
+---------------------
+
+If you are using Django, hot deploy will work out of the box.
```
jhadvig: Is the Gunicorn config file has set  `reload` option to true by default? And the `APP_CONFIG` env var is s pointing to the right config by default ?
If not nuke this line, cause thne the hot deploy is not working out of the box  
-

mnagy: No and no, but this is only applicable to running with Django, not Gunicorn. Django is only intended to run as a development web server, see: https://github.com/openshift/sti-python/blob/master/3.4/s2i/bin/run#L48 and https://docs.djangoproject.com/en/1.8/ref/django-admin/#runserver-port-or-address-port

jhadvig: But does that imply that the hot deploy works out of the box? That no action need to be done so user can modify his code and see immediately the change?

mnagy: Yes.

jhadvig: Yeah I just like the fact that first you are declaring that it's works OOTB and then then you actually tell to the user how to enable it for Gunicorn.

mnagy: Well, it says that it works OOTB *if* you use Django.

jhadvig: Well I can probably live with that

jhadvig:
LGTM

PR #70, opened 2015-10-19
Split README.md to usage and general part

hhorak:
This is similar to https://github.com/openshift/mysql/pull/99. It allows us to maintain two separate variant for every version (default values, image names, links to doc and eventually more, that may be or already is different). As a bonus, we'll have a clean `README.md` with only image-related usage information that can be used in Docker Hub or elsewhere.

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -1,0 +1,205 @@
+Python Docker image
+===================
+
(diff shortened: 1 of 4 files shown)


jhadvig:
need squash

jhadvig:
lgtm

PR #73, opened 2015-10-29
Sync Red Hat build service labels

praiskup:
Related: openshift/postgresql/pull/75


praiskup:
AFAIK we don't build rh-python27-docker, I'm not sure what should be
done with 2.7 directory.

mfojtik:
@praiskup @bparees IMHO we keep the old values for backward compatibility

bparees:
wait, SCL isn't shipping a python2.7 image?

bparees:
Just to clarify, after discussion with PM,  2.7 should be being built/shipped by SCL.

PR #79, opened 2015-12-01
Change RH required labels based on updated spec

hhorak:
For full spec, see https://github.com/projectatomic/ContainerApplicationGenericLabels/blob/master/vendor/redhat/labels.md


rhcarvalho:
@hhorak IIRC these labels are not supported yet, and we dropped this change in some other repo(s), right? If that's the case, then I think we should close this PR.

hhorak:
True, let's close it for now.

PR #97, opened 2016-02-05
Include httpd and httpd-devel in the builder image

rhcarvalho:
This supports using mod_wsgi-express instead of gunicorn in images built with the S2I Python builder.

Closes #59.

PR #100, opened 2016-02-11
Remove *-logos to reduce image sizes

rhcarvalho:
Adding httpd as a dependency brought in ~80MB, ~20 of those from `centos-logos` package, which is liked used to show a logo in the default index page.

This is a follow up to #97 aiming to reduce image sizes, while keeping httpd and httpd-devel in.

As a bonus, the first commit is removing some asymmetries that make comparing the files harder, and makes it harder to automate changing the Dockerfiles for a single Python version and applying the patch to the others.

----------------------------

For the record, here's the trick to avoid copy-paste:

```console
$ git diff -U2 2.7 | sed 's/2.7/3.3/' | git apply -
$ git diff -U2 2.7 | sed 's/2.7/3.4/' | git apply -
```

Normally, `-U2` is not needed, but I had to use it here to ignore part of the context that wouldn't allow for a clean patch (`-U2` means use only 2 lines of diff context instead of the default 3).

### DIFF
--- a/2.7/Dockerfile
+++ b/2.7/Dockerfile
@@ -15,8 +15,8 @@ LABEL io.k8s.description="Platform for building and running Python 2.7 applicati
       io.openshift.expose-services="8080:http" \
       io.openshift.tags="builder,python,python27,rh-python27"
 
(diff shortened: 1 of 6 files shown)


rhcarvalho:
@bparees @mfojtik PTAL
@GrahamDumpleton FYI

mfojtik:
LGTM

bparees:
lgtm

PR #102, opened 2016-02-13
Corrected pkgs libjpeg-turbo libjpeg-turbo-devel

donnydavis:
Added the correct packages 
libjpeg-turbo libjpeg-turbo-devel

### DIFF
--- a/2.7/Dockerfile
+++ b/2.7/Dockerfile
@@ -16,7 +16,7 @@ LABEL io.k8s.description="Platform for building and running Python 2.7 applicati
       io.openshift.tags="builder,python,python27,rh-python27"
 
 RUN yum install -y centos-release-scl && \
-    INSTALL_PKGS="python27 python27-python-devel python27-python-setuptools python27-python-pip epel-release nss_wrapper httpd httpd-devel" && \
+    INSTALL_PKGS="libjpeg-turbo libjpeg-turbo-devel python27 python27-python-devel python27-python-setuptools python27-python-pip epel-release nss_wrapper httpd httpd-devel" && \
     yum install -y --setopt=tsflags=nodocs --enablerepo=centosplus $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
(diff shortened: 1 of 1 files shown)


bufke:
Could someone comment on what is the best way to add various package support? There are a large number of packages that may need installed to support various pypi packages that compile c code. Is the intention that a end user would fork sti-python and add whatever they need? Or should sti-python attempt to support most pypi packages out of box - which I think means adding quite a lot of packages to the base Docker image. For instance many cryptography and image python packages have C dependencies which require additional rpms to be installed.

donnydavis:
This patch is to support most django related projects which depend on jpeg. I know this could get crazy, but this is quite common.

rhcarvalho:
@bparees PTAL

bparees:
@bufke it's pretty case by case so far, we don't have a hard and fast rule.  you've outlined the two options perfectly :)

basically i'd say the process is:
1) if you need it now, create your own image based on sti-python
2) if you have some time, open a pull and we'll discuss the merits of including those packages for everyone

rhcarvalho:
@bparees I think this needs to be post-reviewed :)

- It's unclear what apps or Python libraries that depend on those libs. Perhaps it doesn't belong in the Python layer, but in sti-base? Or in a specialized derivative of sti-python?
- The patch was applied only to the 2.7 image (we could update 3.3 and 3.4 in a follow up)
- The increase in image size is of ~3 MB

```console
$ docker images | grep jpeg
python-27-jpeg                                      latest              15639fcae25f        29 seconds ago      657.7 MB
python-27-no-jpeg                                   latest              804931d10c74        2 days ago          654.8 MB
```

Something else to consider?

bparees:
@rhcarvalho sure.  @donnydavis can you elaborate on why you consider this to be a common dependency?

rhcarvalho:
@donnydavis @bparees let's please move the discussion to a new issue to facilitate tracking of status and completion of necessary follow ups:
https://github.com/openshift/sti-python/issues/109

bparees:
@rhcarvalho +1

PR #110, opened 2016-04-01
Replace .sti -> .s2i in README.md

### LINKED ISSUE #106: Update for .s2i from .sti
when using .sti/ folder, a deprecation warning is thrown during assembly.

if you put your files in .s2i then no warning.

Docs should be updated accordingly, 2.7, 3.3, and 3.4

I can make a PR if preferred

rhcarvalho:
@bparees @soltysh are we ready to point users to use `.s2i` instead of the deprecated `.sti`? I guess so. PTAL.

Fixes #106.

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -89,8 +89,8 @@ Repository organization
 Environment variables
 ---------------------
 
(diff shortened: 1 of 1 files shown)


bparees:
lgtm

PR #111, opened 2016-04-01
Replace .sti -> .s2i in README.md of 3.x

rhcarvalho:
Follow up to #110.

@bparees annnnnd it passed us both...! See why I don't like the level of copy-paste we have here?


bparees:
to be fair, i noticed you'd only changed one file, but i just assumed it didn't need to be fixed in the other files or you would have done it in the first place :)  only decided to double check after.

lgtm

PR #114, opened 2016-05-12
Add container image for python 3.5

### LINKED ISSUE #75: Please add Python 3.5
Python 3.5 [has been released](https://www.python.org/downloads/release/python-350/) and it wuold be nice to have it available :)

hhorak:
Add container image for python 3.5
Python 3.5 is now available in RHSCL Beta only
Closes #75.


mnagy:
@rhcarvalho you going to have time for this?

rhcarvalho:
@hhorak do we want to ship it with the rpm coming from `rhel-server-rhscl-7-beta-rpms`? Or wait till beta is over to drop the last commit?

[test]

rhcarvalho:
@hhorak could you please update the description to use [one of the keywords that will close the issue](https://help.github.com/articles/closing-issues-via-commit-messages/#keywords-for-closing-issues) when this merges?

Including something like `Closes #75.` will do.

rhcarvalho:
It's the 4th time we're copy-pasting the same tree to add another image version... isn't it time yet to refactor it and get rid of all the duplication?

hhorak:
Squashed and comment changed to reference #75 properly.
@rhcarvalho we already ship images in RH registry with Beta packages coming from `rhel-server-rhscl-7-beta-rpms`, so I guess we want to have it here as well.

hhorak:
As for the de-duplication, I'd love to do it but have no a good idea how yet. Anyway, it cannot be coincidence, that I hear this topic for the 3rd time today, so let's brainstorm how it can be done.

3.5/contrib/etc/passwd.template
```diff
@@ -0,0 +1,14 @@
...
+lp:x:4:7:lp:/var/spool/lpd:/sbin/nologin
+sync:x:5:0:sync:/sbin:/bin/sync
+shutdown:x:6:0:shutdown:/sbin:/sbin/shutdown
+halt:x:7:0:halt:/sbin:/sbin/halt
+mail:x:8:12:mail:/var/spool/mail:/sbin/nologin
+operator:x:11:0:operator:/root:/sbin/nologin
+games:x:12:100:games:/usr/games:/sbin/nologin
+ftp:x:14:50:FTP User:/var/ftp:/sbin/nologin
+nobody:x:99:99:Nobody:/:/sbin/nologin
+default:x:${USER_ID}:${GROUP_ID}:Default Application User:${HOME}:/sbin/nologin
```
rhcarvalho: @hhorak why does 3.5 have this file and 2.7, 3.3 and 3.4 don't?

rhcarvalho:
@hhorak thanks, changes look good compared to other versions! And thank you so much for scheduling the de-duplication brainstorming.

LGTM pending just the question about the extra file (maybe we need it in other image versions as well?)

@bparees could you please have a look in the test output? We might have some infra problems.

omron93:
@rhcarvalho This is the same issue as for MongoDB https://github.com/openshift/mongodb/pull/127

3.5/contrib/etc/passwd.template
hhorak: Correct, I'm not sure how this file got to the git, but it doesn't seem to be used, so removing..

bparees:
@hhorak can you squash this and i'll merge it?

hhorak:
Squashed.

PR #123, opened 2016-06-08
Fix wrong namespace for Python 3.3 image

dinhxuanvu:
Python 3.3 image is currently listed under "openshift" namespace
which is deprecated. The correct namespace is "centos".

### DIFF
--- a/3.3/s2i/bin/usage
+++ b/3.3/s2i/bin/usage
@@ -1,6 +1,8 @@
 #!/bin/sh
 
 DISTRO=`cat /etc/*-release | grep ^ID= | grep -Po '".*?"' | tr -d '"'`
+NAMESPACE=centos
+[[ $DISTRO =~ rhel* ]] && NAMESPACE=rhscl
 
 cat <<EOF
 This is a S2I python-3.3 ${DISTRO} base image:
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
@bparees @soltysh Please review. Thanks.

soltysh:
I'll let @bparees speak up, I don't know the details of where is what at this point.

bparees:
python-33 is still under the openshift namespace:
https://github.com/openshift/s2i-python/blob/master/Makefile#L4

```
docker pull centos/python-33-centos7
Using default tag: latest
Trying to pull repository docker.io/centos/python-33-centos7 ... not found
Error: image centos/python-33-centos7:latest not found
```

only the other versions(2.7, 3.4, 3.5) are in the centos namespace.

dinhxuanvu:
@bparees Hey Ben, thanks for the info. I actually wasn't aware of this. Is there a reason we keep Python 3.3 under namespace "openshift" while the rest is in "centos"?

bparees:
all the images the devex team originally created reside in the openshift namespace (python 3.3, ruby 2.0, nodejs 0.10, perl 5.16, php 5.5, etc), newer versions (python 3.4, 3.5, ruby 2.2, etc) were created by the SCL team and as such reside in the namespace of their choosing (centos).  python 2.7, though older than 3.3, was also a new image created by the SCL team, so it's in the centos namespace.

So we're keeping the images originally created (and still maintained by) the openshift team in the openshift namespace, the newer stuff is going in the SCL namespaces (centos for centos and rhscl for rhel).

soltysh:
@bparees I might know the answer but I will ask, is there a chance to be consistent and keep them all in one namespace? Even though we are maintaining them?

bparees:
answered on irc but the short answer is no, we won't be doing that.

dinhxuanvu:
@bparees Thanks, Ben. Will close this PR since we won't be able to fix this.

PR #124, opened 2016-06-09
Add PIP_INDEX_URL env var to support custom index URL/mirror

dinhxuanvu:
Currently, Python S2I images use default pypi.python.org to get all
required packages during deployment process. However, there is a need
to allow users to use custom index URL (mirror) to get the packages.
This option can speed up deployment process.

The assemble code is modified to check for PYPI_MIRROR env var and
then passes the env var into "pip install" command if it exists.

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -130,6 +130,11 @@ a `.s2i/environment` file inside your source code repository.
     Set this variable to a non-empty value to inhibit the execution of 'manage.py migrate'
     when the produced image is run. This only affects Django projects.
 
+* **PIP_INDEX_URL**
+
+    Set this variable to use a custom index URL or mirror to download required packages
+    during build process. This only affects packages listed in requirements.txt.
+
 Source repository layout
 ------------------------
 
--- a/3.3/README.md
+++ b/3.3/README.md
@@ -130,6 +130,11 @@ file inside your source code repository.
     Set this variable to a non-empty value to inhibit the execution of 'manage.py migrate'
     when the produced image is run. This only affects Django projects.
 
+* **PIP_INDEX_URL**
+
+    Set this variable to use a custom index URL or mirror to download required packages
+    during build process. This only affects packages listed in requirements.txt.
+
 Source repository layout
 ------------------------
 
--- a/3.4/README.md
+++ b/3.4/README.md
@@ -130,6 +130,11 @@ file inside your source code repository.
     Set this variable to a non-empty value to inhibit the execution of 'manage.py migrate'
     when the produced image is run. This only affects Django projects.
 
+* **PIP_INDEX_URL**
+
+    Set this variable to use a custom index URL or mirror to download required packages
+    during build process. This only affects packages listed in requirements.txt.
+
 Source repository layout
 ------------------------
 
--- a/3.5/README.md
+++ b/3.5/README.md
@@ -130,6 +130,11 @@ file inside your source code repository.
     Set this variable to a non-empty value to inhibit the execution of 'manage.py migrate'
     when the produced image is run. This only affects Django projects.
 
+* **PIP_INDEX_URL**
+
+    Set this variable to use a custom index URL or mirror to download required packages
+    during build process. This only affects packages listed in requirements.txt.
+
 Source repository layout
 ------------------------
 


dinhxuanvu:
@bparees Please review. Thanks :)

bparees:
looks reasonable to me...can you document the new env variable in the repo readme?

@rhcarvalho seem ok to you?

fyi, knowingly or otherwise, you've started implementing this card:
https://trello.com/c/pVziEaO1/911-make-npm-rubygems-sources-configurable

dinhxuanvu:
@bparees Thanks. Will take care of the docs as you advised.

rhcarvalho:
I'm -1 for this, because it's already supported by the `requirements.txt` format understood by `pip`:

https://pip.readthedocs.io/en/1.1/requirements.html#indexes-find-links

Users that need to specify an alternative or additional mirror should add a line with `--index-url` or `--extra-index-url` to their requirements file.

dinhxuanvu:
@rhcarvalho This PR simply adds another way to utilize our UI on OpenShift Online where you can add the env var during the build process (included in the BuildConfig file) without having to modifying the requirements.txt. The users can always do what you suggested but they don't have to. Also, this comes handy in the future for us to add the env var to use the PyPi mirror that is operated by Ops without having to ask customers to add the URL to requirements.txt as we can inject that to BuildConfig by default if needed.

rhcarvalho:
@dinhxuanvu makes sense in that context, thanks for clarifying!

Have you checked:
https://pip.pypa.io/en/stable/user_guide/#environment-variables

Possibly you can achieve the same without introducing a new environment variable.

rhcarvalho:
In particular, I just tested `PIP_INDEX_URL` and it works for me  :-)

rhcarvalho:
@dinhxuanvu if you're fine with using `PIP_INDEX_URL`, then we could just document it as a supported feature, and perhaps add a test case. WDYT?

rhcarvalho:
Update: `PIP_INDEX_URL` works even with the rather old `pip 1.5.6` version that we have in our images, tested in a fresh container running the `openshift/python-33-centos7` image.

jorgemoralespou:
Name of variable should be consistent with the rest of images

bparees:
> Name of variable should be consistent with the rest of images

not if it's a variable dictated by the underlying framework.  I don't want to get in the business of defining a common "mirror" env variable that we then can't necessarily support for all frameworks, or which takes different formats for different frameworks.  We already tried that with DEV_MODE and failed.

jorgemoralespou:
@bparees Sounds fair, but I will note it down just in case I have to remind you the same for other languages :-P

dinhxuanvu:
@rhcarvalho Hey Rodolfo. I agree with your suggestion. It's good to reuse a familiar env var that is used for pip. I have modified the PR to use the PIP_INDEX_URL var instead. Please review it when you have a chance. Thanks :)

2.7/s2i/bin/assemble
```diff
@@ -14,6 +14,9 @@ echo "---> Copying application source ..."
 cp -Rf /tmp/src/. ./
 
 if [[ -f requirements.txt ]]; then
+  if [ -n "$PIP_INDEX_URL" ]; then
+    export PIP_INDEX_URL=${PIP_INDEX_URL}
```
bparees: what's the point of this export, given that the value is already set and you're then setting it to the same value again?

dinhxuanvu: Hey Ben. My bad on this. Originally, I tested to inject PIP_INDEX_URL directly via env var and it didn't work for some reasons. I thought it was a out-of-the-scope issue so I added those codes this PR. Then, I retested and it was working fine so I removed the code. This PR should be docs change to reflect the env var availability. I will fix this ASAP.

bparees: :+1:

dinhxuanvu:
@rhcarvalho Sorry for double-tag you :D. The PR is now a doc change so code review is no longer needed. Please let me know if the wording is good :). Thanks.

rhcarvalho:
Thanks @dinhxuanvu! I'm fine with the wording. LGTM.

mnagy:
Should this also be documented in openshift-docs repository?

dinhxuanvu:
@mnagy I have a PR for the doc change on openshift-docs as well. Please review when you have a chance. Thanks
https://github.com/sclorg/s2i-python-container/pull/124

wewang58:
@dinhxuanvu ,add cases for the pr: goo.gl/B4brTS , have a question,because container is running without root ,so I cannot use pip list command in pod, how can I check the package is installed by pip? build logs did not show the info

rhcarvalho:
@wewang58 using the `openshift/django-ex` example repository, I could use `pip list` just fine:

```console
$ pip list                                                                                                                                                                                                                       
You are using pip version 7.1.0, however version 8.1.2 is available.                                                                                                                                                                   
You should consider upgrading via the 'pip install --upgrade pip' command.                                                                                                                                                             
Django (1.8.14)                                                                                                                                                                                                                        
django-debug-toolbar (1.4)                                                                                                                                                                                                             
gunicorn (19.4.5)                                                                                                                                                                                                                      
pip (7.1.0)                                                                                                                                                                                                                            
psycopg2 (2.6.1)                                                                                                                                                                                                                       
setuptools (18.0.1)                                                                                                                                                                                                                    
sqlparse (0.2.1)                                                                                                                                                                                                                       
virtualenv (13.1.2)                                                                                                                                                                                                                    
whitenoise (3.0)   
```

You can also try `pip list --user`, to list only the packages installed in the user-site:

```console
$ pip list --user                                                                                                                                                                                                                
You are using pip version 7.1.0, however version 8.1.2 is available.                                                                                                                                                                   
You should consider upgrading via the 'pip install --upgrade pip' command.                                                                                                                                                             
Django (1.8.14)                                                                                                                                                                                                                        
django-debug-toolbar (1.4)                                                                                                                                                                                                             
gunicorn (19.4.5)                                                                                                                                                                                                                      
psycopg2 (2.6.1)                                                                                                                                                                                                                       
sqlparse (0.2.1)                                                                                                                                                                                                                       
whitenoise (3.0)
```


However, I think this doesn't help telling where the package comes from :)

Looking at the test cases, they should be pointing at a PyPI mirror, not a specific package.


One way to check that `PIP_INDEX_URL` is in effect is by setting it to an invalid value and expecting an error. For example:

```console
$ PIP_INDEX_URL=http://not/a/valid/index pip install --user simplejson                                                                                                                                                           
You are using pip version 7.1.0, however version 8.1.2 is available.                                                                                                                                                                   
You should consider upgrading via the 'pip install --upgrade pip' command.                                                                                                                                                             
Collecting simplejson                                                                                                                                                                                                                  
  Retrying (Retry(total=4, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /a/valid/index/simplejson/                           
  Retrying (Retry(total=3, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /a/valid/index/simplejson/                           
  Retrying (Retry(total=2, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /a/valid/index/simplejson/                           
  Retrying (Retry(total=1, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /a/valid/index/simplejson/                           
  Retrying (Retry(total=0, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /a/valid/index/simplejson/                           
  DEPRECATION: Failed to find 'simplejson' at http://not/a/valid/index/simplejson/. It is suggested to upgrade your index to support normalized names as the name in /simple/{name}.                                                   
  Retrying (Retry(total=4, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /a/valid/index/                                      
  Retrying (Retry(total=3, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /a/valid/index/                                      
  Retrying (Retry(total=2, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /a/valid/index/                                      
  Retrying (Retry(total=1, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /a/valid/index/                                      
  Retrying (Retry(total=0, connect=None, read=None, redirect=None)) after connection broken by 'ProtocolError('Connection aborted.', gaierror(-2, 'Name or service not known'))': /a/valid/index/                                      
  Cannot fetch index base URL http://not/a/valid/index/                                                                                                                                                                                
  The repository located at not is not a trusted or secure host and is being ignored. If this repository is available via HTTPS it is recommended to use HTTPS instead, otherwise you may silence this warning and allow it anyways wit
h '--trusted-host not'.                                                                                                                                                                                                                
  Could not find a version that satisfies the requirement simplejson (from versions: )                                                                                                                                                 
No matching distribution found for simplejson        
```

wewang58:
@rhcarvalho ,updated cases with PIP_INDEX_URL=https://pypi.python.org/pypi and add cases with invalid url, pls help to review again:goo.gl/B4brTS 
and met issue when implement :
1. oc new-build openshift/python:3.3~https://github.com/openshift/django-ex PIP_INDEX_URL=https://pypi.python.org/pypi
2 . ---> Installing dependencies via https://pypi.python.org/pypi ...
Downloading/unpacking django>=1.8,<1.9 (from -r requirements.txt (line 1))
Could not find any downloads that satisfy the requirement django>=1.8,<1.9 (from -r requirements.txt (line 1))
I think it cannot recognize django>=1.8,<1.9 format in requirements.txt ,do you think it's a issue?  if I use https://github.com/wewang58/django-ex/blob/master/requirements.txt (Django==1.8.1), it works.

rhcarvalho:
@wewang58 你好！

> I think it cannot recognize django>=1.8,<1.9 format in requirements.txt ,do you think it's a issue?

This seems to be a problem with PyPI... though I could not find good evidence to explain it. I tried a different URL and it seems to work:

```console
$ PIP_INDEX_URL=https://pypi.python.org/simple pip install --force-reinstall --no-cache -r /tmp/requirements.txt
Collecting Django<1.9,>=1.8 (from -r /tmp/requirements.txt (line 1))
  Downloading Django-1.8.14-py2.py3-none-any.whl (6.2MB)
    100% |████████████████████████████████| 6.2MB 56.1MB/s 
Collecting django-debug-toolbar==1.4 (from -r /tmp/requirements.txt (line 2))
  Downloading django_debug_toolbar-1.4-py2.py3-none-any.whl (212kB)
    100% |████████████████████████████████| 215kB 66.5MB/s 
Collecting gunicorn==19.4.5 (from -r /tmp/requirements.txt (line 3))
  Downloading gunicorn-19.4.5-py2.py3-none-any.whl (112kB)
    100% |████████████████████████████████| 122kB 105.0MB/s 
...
```

Instead of `https://pypi.python.org/pypi`, using `https://pypi.python.org/simple`.

---

Note that if we set PIP_INDEX_URL to a valid mirror, especially the default mirror, that has the same effect as not setting the variable.
Therefore if the build is successful, we can't tell if it worked because the mirror was used or ignored (in case the variable was not set or ignored for some reason).

IMHO, what we need to test is that `pip` uses the value of `PIP_INDEX_URL` that we set. In that case, setting it to a bad value and seeing the error seems to be the most effective test.

wewang58:
@rhcarvalho ,yes using https://pypi.python.org/simple ,it works now ,and can you see my cases again: goo.gl/B4brTS , should update or ok? thanks

rhcarvalho:
@wewang58 they are okay.

As I noted, PIP_INDEX_URL=https://pypi.python.org/simple doesn't help us much (if we removed this PR, or even made a typo in the var name or removed the ability to pass env vars from s2i, that test would pass anyway...)

wewang58:
@rhcarvalho  oh, got your mean now ,just do negative test and already updated cases :goo.gl/B4brTS, thanks

PR #127, opened 2016-06-17
Use OSE 3.2 repos for RHEL images

hhorak:
This is just change in version, so we use the most recent repositories for nss_wrapper package.


bparees:
lgtm

PR #133, opened 2016-07-25
reduce max worker calculations

bparees:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1351065

previous calculations were too aggressive, resulting in too many workers and overloading the container.

@rhcarvalho ptal


bparees:
@hhorak ptal

soltysh:
LGTM

PR #135, opened 2016-07-27
Comment between LABEL lines breaks OSBS builds

hhorak:
Let's use a bit slower but safer way to define OSBS labels.


bparees:
you could just put the comment at the top, before the LABEL block, and reorder the labels so the osbs labels are first.

hhorak:
Right, let me do it so we save some osbs time :)

PR #137, opened 2016-08-10
Allow to pull base image during build

omron93:
This PR adds option to update base image during build. So user can add "UPDATE_BASE=1" to `make` command and `docker build` will try to pull latest base image - see "--pull" option in https://docs.docker.com/engine/reference/commandline/build/

@hhorak Please merge - same PR as sclorg/mariadb-container#10


hhorak:
LGTM.

PR #140, opened 2016-08-19
Display custom PIP_INDEX_URL info

dinhxuanvu:
The current PIP version (1.5.6) in Python 3.3 and 3.4 doesn't display
the repository link for non-PyPi during "pip install" step in the log
which may make debugging difficult.

The fix is added to display the non-PyPi repository link that is set
by users via PIP_INDEX_URL env var.

Bug 1357794
Link https://bugzilla.redhat.com/show_bug.cgi?id=1357794


dinhxuanvu:
@bparees PTAL. Thanks

Issue #86, opened 2016-01-11
Detection of app.py and warnings if missing.


GrahamDumpleton:
Is the code in the ``run`` script of:

```
APP_FILE="${APP_FILE:-app.py}"
if [[ -f "$APP_FILE" ]]; then
  echo "---> Running application from Python script ($APP_FILE) ..."
  exec python "$APP_FILE"
fi
echo "WARNING: file '$APP_FILE' not found."
```

really the best way of doing this?

This will output a warning message of:

```
WARNING: file 'app.py' not found.
```

every time the script is run when ``APP_FILE`` was not set by the user.

Further, if someone supplies the ``APP_FILE`` environment variable explicitly but gets it wrong, with it pointing to a directory or a non existent file, should it really be falling through to later options?

It would be better to tell them explicitly that the ``APP_FILE`` they provided was wrong and exit.

The last thing they want is it silently matching a latter possibility and so not running what they intended, with it being hard to possibly understand why.

PR #143, opened 2016-09-14
Fail with an error message when the APP_FILE variable is set incorrectly

pkubatrh:
This PR changes the APP_FILE variable check to immediately fail when it is set incorrectly instead of trying other options.

Resolves: #86


bparees:
lgtm, add to the other python image versions?
[test]

pkubatrh:
Changes added to other versions

bparees:
[test]

PR #144, opened 2016-09-16
Add numpy/scipy support + test

pkubatrh:
This PR adds packages needed to succesfully install numpy and scipy in the image.

A simple test case has also been added. I have opted for testing numpy only, since scipy takes a long time to install.


bparees:
[test]

Issue #134, opened 2016-07-27
Detection of Django manage.py not predictable.


GrahamDumpleton:
The `assemble` script works out what is the Django `manage.py` file by using:

```
    # Find shallowest manage.py script, either ./manage.py or <project>/manage.py
    manage_file=$(find . -maxdepth 2 -type f -name 'manage.py' -printf '%d\t%P\n' | sort -nk1 | cut -f2 | head -1)
```

This is problematic when it could match multiple files because it takes the first in sorted order.

Someone could have `manage.py` in the root of their repository and then have a Django application directory called `hello`. They might for some reason create a file called `hello/manage.py` as part of their Django application code. The `find` command will pick up the `manage.py` in the sub directory rather than giving precedence to the `manage.py` in the root directory, because 'h' comes before 'm'.

The rule should be that if there is a `manage.py` in the root directory that it takes precedence.

If there is no `manage.py` in the root directory and you do resort to searching the immediate subdirectories, then it shouldn't blindly take the first one it finds as that still isn't predictable. It might initially use the right one, but then if someone later creates a new sub directory that comes earlier in sort order, then it will start using that new one when it shouldn't.

The rule should be that if more than one `manage.py` file is found when searching the immediate sub directories that it should issue a warning/error and then skip any attempt to interpret one as a Django `manage.py` file.

If for some reason a user really needs duplicates, but only one in a sub directory is the real `manage.py` file, then there would need to be an environment variable that can be set to say which it is.

Is arguable that only the `manage.py` in the root directory should be detected and if one in a sub directory is needed, then required that user specify environment variable indicating the location of the Django project by giving path to the `manage.py` in the sub directory.

There is also the issue mentioned in prior issues about `collectstatic` that there is no attempt to verify that the `manage.py` is actually for Django. The `Flask-Scripts` package also uses a `manage.py` and so could be incorrectly interpreted that is Django when it isn't.

bparees:
i'd almost rather see us just look for it in root and if you want it somewhere else, we give you an env variable to tell us which manage.py file you want to use.  how does that sound?

GrahamDumpleton:
Rather than environment variable giving path to manage.py in sub directory, my current thinking is that you give it the directory name for the Django project if not in root. The scripts would then look for manage.py within that directory itself rather than user calling it out explicitly. Thus instead of:

```
DJANGO_MANAGE_SCRIPT=hello/manage.py
```

would use:

```
[...]

PR #145, opened 2016-09-19
Add APP_HOME variable to set the application's location when needed

pkubatrh:
This PR adds a new variable (APP_HOME) that can be used to provide the `run` and `assemble` scripts with the location of the application to be run (sub-directory containing `wsgi.py` and `manage.py` files for Gunicorn and Django, respectively) in case the application is not located directly in the root.
The `assembly` and `run` scripts have also been changed so as to only look in the directory provided (or the root) instead of doing a deep search for `wsgi.py` or `manage.py`, which can result in wrong findings.
Additionaly, this PR also adds a simple test case to verify this change for Gunicorn.

Resolves: #134

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -115,6 +115,14 @@ a `.s2i/environment` file inside your source code repository.
     can be read from there. For an example, see
     [setup-test-app](https://github.com/openshift/s2i-python/tree/master/2.7/test/setup-test-app).
 
+* **APP_HOME**
+
+    This variable can be used to specify a sub-directory in which the application to be run is contained.
+    The directory pointed to by this variable needs to contain `wsgi.py` (for Gunicorn) or `manage.py` (for Django).
+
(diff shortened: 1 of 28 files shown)


bparees:
@grahamdumpleton ptal.

hhorak:
[test]

hhorak:
I think this is good idea, but rather re-testing after few weeks. If tests still pass, LGTM.

omron93:
Rebase needed to get RHEL7 CI passing..

pkubatrh:
Rebased. Thanks for the notification.

hhorak:
LGTM.

coreydaley:
Looks like this pr broke the tests in origin for the django-ex repo.  Is there a specific reason that the maxdepth was changed to 1 inside of the is_gunicorn_installed functions for all python versions?

bparees:
@coreydaley I think the reason is summarized in the issue: https://github.com/sclorg/s2i-python-container/issues/134

can we fix our test/sample app?

PR #148, opened 2016-10-04
Fix getting container names of versions in openshift namespace

omron93:
This PR fixes problem when version in openshift namespace  is not the lesser one (example of s2i-python-container). In this case, `NAMESPACE` variable is not set back to `openshift/` and created versions have wrong name.

@hhorak @bparees @praiskup Please take a look. I will create PRs for other repos after merge.


praiskup:
Looks reasonable.  Thanks, worth syncing in other containers, or maybe moving postgresql-container/pull/138 forward ...

bparees:
lgtm

PR #149, opened 2016-10-06
Added cccp.yml file for centos container pipeline.

mohammedzee1000:
This is for the purpose of building these containers
on CentOS Container Pipeline.


hhorak:
It does not influence image functionality, so no need to test. LGTM.

PR #150, opened 2016-10-12
don't install the epel repo

bparees:


bparees:
@hhorak @praiskup ptal

bparees:
@praiskup you ok w/ this one too?

praiskup:
Yes, thanks!

praiskup:
FWIW, I epel-release should be installed before installing nss_wrapper, not in the same yum transaction; so the previous usage was probably broken anyway.

PR #151, opened 2016-10-19
Fix the wrong condition check for PIP_INDEX_URL in Python 3.4

dinhxuanvu:
The condition check for PIP_INDEX_URL in Python 3.4 is placed in a
wrong if-statement. It should be placed in requirements.txt if-statement,
not set.up one.

PR: Display custom PIP_INDEX_URL info (140)

### DIFF
--- a/3.4/s2i/bin/assemble
+++ b/3.4/s2i/bin/assemble
@@ -15,16 +15,16 @@ echo "---> Installing application source ..."
 mv /tmp/src/* ./
 
 if [[ -f requirements.txt ]]; then
-  echo "---> Installing dependencies ..."
-  pip install --user -r requirements.txt
-fi
-
-if [[ -f setup.py ]]; then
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
@bparees PTAL :)

bparees:
@dinhxuanvu it looks like 3.5 is missing this print statement entirely, can you fix that too please?  (and maybe sanity check the rest of the assemble scripts)

dinhxuanvu:
@bparees Hey Ben. I don't think Python 3.5 has this issue. I only added this print statement due to Python 3.3 and 3.4 use an older version of pip which doesn't print out the URL where the packages are downloaded. Python 3.5 uses a newer pip so it doesn't have the issue.

bparees:
@dinhxuanvu ok cool, thanks.

Issue #158, opened 2016-11-28
pip version is too old and can't install Linux wheels


GrahamDumpleton:
You need at least pip version 8.1.0 in order to be able to install Linux wheels.

Some people (e.g. Sentry server), are now only making available some of their software packages as Linux wheels. This means it is impossible to install their software. This could get worse over time.

The SCL pip version should be updated.

There should also be an easy way for a user to say that a newer version of pip should be installed as part of the build before installing packages packages from requirements.txt. This will likely necessitate needing to add use of a Python virtual environment first, as you can't upgrade pip in a per user Python site-packages as it will fail trying to uninstall the system one first.

pkubatrh:
Closing via #159

Issue #117, opened 2016-05-31
Building 2.7 app with Gunicorn fails with syntax error


mnagy:
```
$ s2i build git://github.com/mnagy/stress_test centos/python-27-centos7 python-sample-app  --context-dir=sti-python/test-app
I0531 08:20:51.853106 30506 clone.go:31] Downloading "git://github.com/mnagy/stress_test" ("sti-python/test-app") ...
I0531 08:20:52.638103 30506 install.go:251] Using "assemble" installed from "image:///usr/libexec/s2i/assemble"
I0531 08:20:52.638170 30506 install.go:251] Using "run" installed from "image:///usr/libexec/s2i/run"
I0531 08:20:52.638211 30506 install.go:251] Using "save-artifacts" installed from "image:///usr/libexec/s2i/save-artifacts"
---> Copying application source ...
---> Installing dependencies ...
Downloading/unpacking gunicorn (from -r requirements.txt (line 1))
Installing collected packages: gunicorn
Compiling /tmp/pip-build-izJkqx/gunicorn/gunicorn/workers/_gaiohttp.py ...
File "/tmp/pip-build-izJkqx/gunicorn/gunicorn/workers/_gaiohttp.py", line 84
yield from self.wsgi.close()
^
SyntaxError: invalid syntax

Successfully installed gunicorn
Cleaning up...
```

soltysh:
See [this comment](https://github.com/benoitc/gunicorn/issues/788#issuecomment-92877751), but the general take-away is we need newer version of pip, which ignores that error. I've just tried it and it works with pip 8.1.2. @hhorak is update of pip possible in scl? @mnagy have you tried other versions/images, does it apply there as well?

mnagy:
@soltysh nice, thanks for finding that! Yes, I tried and they work well. And actually, I realized later that this error message does not fail building the image, so it actually works ok. Would be nice to get rid of it, however.

soltysh:
:+1: for getting rid of it.

hhorak:
We'll track pip update in this BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1344674

pkubatrh:
Closing via #159

Issue #90, opened 2016-01-13
Relying on per user site-packages equivalent causes problems.


GrahamDumpleton:
The way the image is set up at present means that no Python virtual environment is used. Instead the installation of Python packages from the user `requirements.txt` file is done into the users home directory using the `--user` option to `pip`. Similarly for `setup.py`.

```
if [[ -f requirements.txt ]]; then
  echo "---> Installing dependencies ..."
  pip install --user -r requirements.txt
fi

if [[ -f setup.py ]]; then
  echo "---> Installing application ..."
  python setup.py develop --user
fi
```

Use of no Python virtual environment will ultimately cause problems and confusion for a user.

This is because when the SCL package for Python, e.g.: `python2.7`, is installed, it is dragging in as dependencies a bunch of Python packages from operating system package repository.

```
Dependencies Resolved

================================================================================
 Package                     Arch    Version              Repository       Size
================================================================================
Installing:
 python27                    x86_64  1.1-20.el7           centos-sclo-rh  4.8 k
Installing for dependencies:
 dwz                         x86_64  0.11-3.el7           base             99 k
 iso-codes                   noarch  3.46-2.el7           base            2.7 M
 perl-srpm-macros            noarch  1-8.el7              base            4.6 k
 python27-python             x86_64  2.7.8-3.el7          centos-sclo-rh   81 k
 python27-python-babel       noarch  0.9.6-8.el7          centos-sclo-rh  1.4 M
 python27-python-devel       x86_64  2.7.8-3.el7          centos-sclo-rh  384 k
 python27-python-docutils    noarch  0.11-1.el7           centos-sclo-rh  1.5 M
 python27-python-jinja2      noarch  2.6-11.el7           centos-sclo-rh  518 k
 python27-python-libs        x86_64  2.7.8-3.el7          centos-sclo-rh  5.6 M
 python27-python-markupsafe  x86_64  0.11-11.el7          centos-sclo-rh   25 k
 python27-python-nose        noarch  1.3.0-2.el7          centos-sclo-rh  274 k
 python27-python-pip         noarch  1.5.6-5.el7          centos-sclo-rh  1.3 M
 python27-python-pygments    noarch  1.5-2.el7            centos-sclo-rh  774 k
 python27-python-setuptools  noarch  0.9.8-5.el7          centos-sclo-rh  400 k
 python27-python-simplejson  x86_64  3.2.0-3.el7          centos-sclo-rh  173 k
[...]

Issue #58, opened 2015-08-28
Per user Python install area bin directory not in PATH when pip being run.


GrahamDumpleton:
May relate to #57, but when `pip` is run to install modules, the bin directory for the per user Python install area, of `${HOME}/.local/bin` is not in the `PATH`. This will cause problems where `requirements.txt` for `pip` lists two modules which have an ordering relationship and where the latter relies on being able to find a program installed as a side effect of the first module to be installed.

Therefore, in `assemble` scripts, prior to:

```
if [[ -f requirements.txt ]]; then
  echo "---> Installing dependencies ..."
  pip install --user -r requirements.txt
fi

if [[ -f setup.py ]]; then
  echo "---> Installing application ..."
  python setup.py develop --user
fi
```

what is needed is:

```
PATH=$HOME/.local/bin:$PATH
```

This may not be necessary if the problem is due to the current working directory not being `${HOME}` and as such fixing #57 would resolve the problem.

GrahamDumpleton:
Confirmed that this is fixed by fixing #57.

To test this, one can run:

```
sti build https://github.com/GrahamDumpleton/openshift-sti-python-sample.git openshift/python-27-centos7 python-sample-app
```

The repo being pointed at has a ``requirements.txt`` that installs two special modules for test. When run it yields:

```
W0828 16:30:11.868852 04863 docker.go:299] BuilderImage uses deprecated environment variable STI_SCRIPTS_URL, please migrate it to io.openshift.s2i.scripts-url label instead!
W0828 16:30:12.143298 04863 sti.go:127] Clean build will be performed because of error saving previous build artifacts
I0828 16:30:12.536852 04863 sti.go:393] ---> Copying application source ...
I0828 16:30:12.544565 04863 sti.go:393] ---> Installing dependencies ...
I0828 16:30:13.016062 04863 sti.go:393] Downloading/unpacking https://dl.dropboxusercontent.com/u/22571016/module1-1.0.tar.gz (from -r requirements.txt (line 1))
I0828 16:30:14.970469 04863 sti.go:393]   Downloading module1-1.0.tar.gz
I0828 16:30:14.984304 04863 sti.go:393]   Running setup.py (path:/tmp/pip-tMGbPB-build/setup.py) egg_info for package from https://dl.dropboxusercontent.com/u/22571016/module1-1.0.tar.gz
I0828 16:30:15.105179 04863 sti.go:393]
[...]

PR #159, opened 2016-11-29
Add support for Python virtual environment and optional pip upgrade.

GrahamDumpleton:
This set of changes addresses #90 by adding in use of a Python virtual environment. It means #58, #117 and #158 can be closed out as well since change provides way for user to indicate they want the latest ``pip`` version installed, overriding whatever version comes with SCL or which the image was explicitly installing to match RHEL.

I have checked this for CentOS images for each Python version by using OpenShift to do a Docker build of the respective directives and then deploying a test application etc. I have not checked this for RHEL variants of the images or used the top level makefile build system.

A number of points to look at when evaluating this change.

* Execution of:

```
fix-permissions /opt/app-root
```

at end of s2i/bin/assemble generates lot of warnings.

```
chgrp: changing group of '/opt/app-root/include/python2.7': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/UserDict.py': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/_abcoll.py': Operation not permitted
...
chmod: changing permissions of '/opt/app-root/include/python2.7': Operation not permitted
chmod: changing permissions of '/opt/app-root/lib/python2.7/UserDict.py': Operation not permitted
chmod: changing permissions of '/opt/app-root/lib/python2.7/_abcoll.py': Operation not permitted
chmod: changing permissions of '/opt/app-root/lib/python2.7/_weakrefset.py': Operation not permitted
```

This is because the Python virtual environment creates symbolic links back to the main Python installation.

```
(app-root)sh-4.2$ pwd
/opt/app-root
(app-root)sh-4.2$ ls -las include
total 0
0 drwxrwxr-x. 2 default root 22 Nov 28 06:02 .
0 drwxrwxr-x. 7 default root 97 Nov 28 06:04 ..
0 lrwxrwxrwx. 1 default root 43 Nov 28 06:02 python2.7 -> /opt/rh/python27/root/usr/include/python2.7
```

The fault here is not the use of a Python virtual environment or where it is placed. The problem here is the ``fix-permissions`` script. That script has:

```
#!/bin/sh
# Fix permissions on the given directory to allow group read/write of
# regular files and execute of directories.
find $1 -exec chgrp 0 {} \;
find $1 -exec chmod g+rw {} \;
find $1 -type d -exec chmod g+x {} +
```

If any file created under the directory for which permissions are being changed is actually a symbolic link, then it will attempt to change the target of the symbolic link. If the symbolic link points to a file elsewhere in the file system which the user has no privileges to modify, you will get these warnings. As ``chmod`` on Linux has no way to say change the permissions on the link rather than the target, the way to get rid of all the warnings would be to use the ``-f`` option to ``chgrp`` and ``chmod``. This at least suppresses the warnings. The other alternative is to use a fancy condition with ``find`` so that it doesn't exec the command when a file is a symbolic link.

Either way, this is an issue for the command base image that ``fix-permissions`` is added in.

* The ``pip`` cache directory must be disabled.

There is no real reason to have the cache anyway as it just takes up extra space in the image and is not needed by the running application.

Leaving the ``pip`` cache directory in place causes other problems because the directory permissions get changed. As a consequence, ``pip`` will try and set them back to what it wants if ``pip`` were run in a subsequent layer. Changing permissions of a directory in a higher layer can cause the AUFS permissions bug and the directory then becomes inaccessible and could then cause ``pip`` to fail later.

To ensure the ``pip`` cache is always disabled, including for when ``pip`` is run explicitly in a derived image ``Dockerfile``, the ``Dockerfile`` for these images should set:

```
PIP_NO_CACHE_DIR=off
```

It has to have a value of ``off`` because of the strange parsing that ``pip`` does on argument values.

* Included optional upgrading of pip to latest version via environment variable. This is to address #158 where ``pip`` which is a part of SCL is too old to be able to install Linux wheels, plus other reported issues. Some packages are now only supplying binary wheels for Linux and no source packages. Upgrading enabled by setting ``UPGRADE_PIP_TO_LATEST`` to non empty value:

```
UPGRADE_PIP_TO_LATEST=on
```

as build time environment variable. Could also be supplied in the ``.s2i/environment`` file.

Making this change in same PR as adding Python virtual environment as it requires use of Python virtual environment. It is not possible to install a newer ``pip`` version into a per user ``site-packages`` directory as it can't first uninstall the ``pip`` version installed in system wide Python installation.

* No longer need ``$HOME/.local/bin`` in ``$PATH`` so is removed. This should not cause any problems as was only used by Python when per user ``site-packages`` was being used. The ``--user`` option to ``pip`` has also been removed though as that is incompatible with Python virtual environments. That is, cannot use ``pip install --user`` when using a Python virtual environment.

There is a very low risk that someone had created a derived Docker image, or a custom ``assemble`` script which ran ``pip install --user`` for some reason, but that is not likely. So not expecting this change would really cause any real world impacts on existing users. If a user was affected, they would need to remove the ``--user`` option, and if running it from a ``Dockerfile`` and also activate the Python virtual environment first, in addition to enabling SCL packages as they would already have had to be doing.

* As the Python virtual environment needs to be manually activated, if someone uses ``oc exec pod command`` or ``oc rsh pod command``, then the Python virtual environment will not be activated. This is not seen as a problem because they also would have needed to enable the SCL Python package themselves as well. The latter would have meant they had to wrap the command in a shell script for it to work and rely on the ``BASH_ENV`` hack. So long as that is done then the Python virtual environment should be activated fine, as activation was added to the ``contrib/etc/scl_enable`` script invoked as a side effect of ``BASH_ENV``.

* In CentOS image for Python 3.3, still install `pip` 1.5.6 to match what was provided in RHEL image, but now only installed within the Python virtual environment and not in the base SCL Python installation. Also install it with ``pip`` rather than ``easy_install`` to ensure that it replaces the ``pip`` version in the Python virtual environment correctly.


GrahamDumpleton:
May want to tweak the messages output when running pip due to:

* https://github.com/sclorg/s2i-python-container/commit/46c95d15071e4a4f013ca2ead0aa9cb3a951cca9

and that can now indicate you want newer pip installed.

To be consistent all images should likely use:

```
if [[ -f requirements.txt ]]; then
  if [ -n "$PIP_INDEX_URL" ]; then
    echo "---> Installing dependencies via $PIP_INDEX_URL ..."
  else
    echo "---> Installing dependencies ..."
  fi
  pip install -r requirements.txt
fi
```

There will be duplicate information in logging when new enough ``pip`` version is available, but that is less of a problem than the fact that the ``assemble`` scripts in 2.7 and 3.5 differ to the others. For the purposes of making maintenance easy, they should all be the same and not be concerned if the progress log message from the ``assemble`` script is showing the index URL as well as ``pip``. In the case where the URL is invalid and causes ``pip`` to fail on starting, could be said that more useful to always display in progress log message anyway in case the error message from ``pip`` is cryptic in the way it displays an invalid index URL.

hhorak:
[test]

GrahamDumpleton:
Note that of all the PRs, this one should be regarded as highest priority.

hhorak:
[test]

pkubatrh:
There should also be some test for this change present. Something along the lines of setting the env var to upgrade pip to latest and then trying to install a package that is available only as a wheel, but that can be added later. I have verified the new functionality manually for the time being.
LGTM, trying out new merge powers

PR #160, opened 2016-11-29
Do not publish a container's port(s) to the host in tests

omron93:
This may lead to conflicts with tests for different repository.

@hhorak @bparees Please take a look.


bparees:
[test]

bparees:
not sure how this works since we later try to access the container on $test_port, but if it passes i'm fine with it.

bparees:
lgtm

Issue #164, opened 2016-12-05
Tests for each Python version should include uWSGI test.


GrahamDumpleton:
As uWSGI is one of the major WSGI servers, should include a test for uWSGI against each version.

GrahamDumpleton:
The uWSGI tests were merged in 682afd6ba8212dabd80ccef4d707409e5467865a.

PR #162, opened 2016-12-05
Use SCL httpd24 packages instead of operating system httpd package.

GrahamDumpleton:
As per #161 use httpd24 packages from Software Collections Library instead of the operating system httpd package.

This becomes important for corporate customers who need to integrate their Python WSGI applications into an SSO infrastructure. By using the SCL httpd24 package, along with either mod_ldap, mod_auth_kerb or mod_session, also installed from SCL in this change, they can when using mod_wsgi-express use the authentication and authorisation handlers of Apache httpd server, backed by the providers in these modules to implement their SSO solution.

This enables them to use a supported SSO mechanism and they do not need to implement a custom SSO solution within their own Python WSGI application.

These changes were tested on CentOS for Python 2.7, 3.3, 3.4 and 3.5 using the application:

* https://gitlab.com/osevg/python-flask-modwsgi

For example, using new versions of modules:

```
oc new-app python:2.7~https://gitlab.com/osevg/python-flask-modwsgi --name app27
```

This validated that mod_wsgi-express correctly found the Apache installation at build time and could compile itself, with it then starting successfully when deployed.

The test didn't check the operation of mod_ldap, mod_auth_kerb or mod_session as these were pulled from SCL and it is expected that are complete and will work when required.

Testing of RHEL variants of images not done at this point, however changes are still made to the RHEL versions of the Dockerfile for each Python version.

Issue #126, opened 2016-06-14
Support running of an app.sh script.


GrahamDumpleton:
First raised as comment against separate issue:
- https://github.com/sclorg/s2i-python-container/issues/92#issuecomment-206080573

but breaking out as independent issue.

The ability to provide an `app.sh` file makes things easier when wishing to run an alternate WSGI server such as uWSGI or Waitress where there may not be a way to start it up by importing a Python module and executing some API.

Right now you would have to do the following.

1 - Create a `wsgi.py` file with your WSGI application.

2 - Create an `app.py` file containing:

```
import os

os.execl('/opt/app-root/src/app.sh', 'uwsgi')
```

3 - Create an `app.sh` file containing:

```
#!/bin/bash

exec uwsgi \
    --http-socket :8080 \
    --die-on-term \
    --master \
    --single-interpreter \
    --enable-threads \
    --threads=5 \
    --thunder-lock \
    --module wsgi
```

The reason for using the `app.sh` in this case is so that `PATH` searching for `uwsgi` program can be relied on. If you try and launch `uwsgi` from `app.py`, because `os.exec` calls require fill path, you have to either hardwire the path, the location of which would change if a switch is made to a Python virtual environment rather than per user `site-packages`, or you have to add a Python function to search `PATH` yourself.

Far simpler to add support for `app.sh`. So something like the following, which would go before existing check and execution of `app.py`.

```
APP_SCRIPT="${APP_SCRIPT:-app.sh}"
if [[ -x "$APP_SCRIPT" ]]; then
  echo "---> Running application from script ($APP_SCRIPT) ..."
  exec "$APP_SCRIPT"
fi
```

In general, easier to use shell script rather than Python code to do special stuff prior to executing command line based server.

bparees:
@GrahamDumpleton you want to submit a PR updating the run script(s)?

pkubatrh:
Implemented by #174

PR #165, opened 2016-12-05
Add tests for using uWSGI as the WSGI server. #164

GrahamDumpleton:
These are tests for running uWSGI as WSGI server.

This is set up to jump from ``app.py`` to ``app.sh``, which then runs ``uwsgi``. When #126 is implemented, this will also serve as test of having ``app.sh`` and the ``app.py`` file will not be needed.

A merge conflict will occur for the changes to ``run`` scripts with #162. One or the other will need to resolve the conflict depending on order PR is merged.

### DIFF
--- a/2.7/test/run
+++ b/2.7/test/run
@@ -8,7 +8,7 @@
 #
 IMAGE_NAME=${IMAGE_NAME:-openshift/python-27-centos7-candidate}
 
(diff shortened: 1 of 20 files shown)


pkubatrh:
[test]

pkubatrh:
LGTM
@hhorak please merge if you agree as I would like to use these tests for #126 in jenkins.

hhorak:
Merging.

Issue #156, opened 2016-11-21
Default lang/locale should be UTF-8, not ASCII.


GrahamDumpleton:
People are used to the default lang/locale of an operating system when using the command line being a variant of UTF-8. The consequence of this is that people often unknowingly use or try to output strings, using ``print()`` from a Python web application or module which contains character strings that are valid UTF-8, but which will fail if the default lang/locale is ASCII.

When they try and deploy any code which does this to OpenShift it will fail when using recommended gunicorn, or if they roll their own web application as an ``app.py``. For example:

```
[2016-11-21 02:28:10 +0000] [1] [INFO] Starting gunicorn 19.6.0
[2016-11-21 02:28:10 +0000] [1] [INFO] Listening at: http://0.0.0.0:8080 (1)
[2016-11-21 02:28:10 +0000] [1] [INFO] Using worker: sync
[2016-11-21 02:28:10 +0000] [28] [INFO] Booting worker with pid: 28
[2016-11-21 02:28:10 +0000] [29] [INFO] Booting worker with pid: 29
[2016-11-21 02:28:10 +0000] [30] [INFO] Booting worker with pid: 30
Internal Server Error: /
Traceback (most recent call last):
  File "/opt/app-root/src/.local/lib/python3.5/site-packages/django/core/handlers/exception.py", line 39, in inner
    response = get_response(request)
  File "/opt/app-root/src/.local/lib/python3.5/site-packages/django/core/handlers/base.py", line 249, in _legacy_get_response
    response = self._get_response(request)
  File "/opt/app-root/src/.local/lib/python3.5/site-packages/django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/opt/app-root/src/.local/lib/python3.5/site-packages/django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/app-root/src/demo/views.py", line 4, in index
    print(u'\u292e')
UnicodeEncodeError: 'ascii' codec can't encode character '\u292e' in position 0: ordinal not in range(128)
10.1.7.1 - - [21/Nov/2016:02:28:13 +0000] "GET / HTTP/1.1" 500 27 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/54.0.2840.98 Safari/537.36"
```

[...]

PR #167, opened 2016-12-07
Set defaults for lang/locale and I/O encoding.

GrahamDumpleton:
This addresses issues raised in #156 whereby various Python web applications and modules might assume that a default locale off UTF-8 is used. If such an application or module is used on OpenShift it will fail. It is not necessarily realistic to expect that people will be able to fix their code to handle bytes/unicode encoding correctly and so sensible defaults should be used.

To do this three environment variables should be set. These are ``LANG``, ``LC_ALL`` and ``PYTHONIOENCODING``.

Technically you might be able to get away with ``LANG`` and not ``LC_ALL`` for Python applications as Python will look for both when initialising. It is though safer to set both as the behaviour of separate non Python applications executed from the web application may not be as tolerant and so require both to handle different situations.

The ``PYTHONIOENCODING`` is also needed to ensure that when a terminal is not associated with the running application, which is default for application run from ``CMD``, or where an application is run and input/output passes through a pipe, that a suitable encoding is also set otherwise it will fallback to ASCII for Python 2 applications. Python 3 applications will possibly fallback to UTF-8, but better to set it regardless of Python version to be consistent.


pkubatrh:
[test]

hhorak:
The RHEL7 test fails with "Network timed out while trying to connect to https://index.docker.io/v1/repositories/rhscl/s2i-base-rhel7/images. You may want to check your internet connection or if you are behind a proxy." I guess it is some temporary random issue only.

pkubatrh:
>The RHEL7 test fails with "Network timed out while trying to connect to https://index.docker.io/v1/repositories/rhscl/s2i-base-rhel7/images. You may want to check your internet connection or if you are behind a proxy." I guess it is some temporary random issue only.

Actually failing because of:
Get https://registry.access.redhat.com/v2/rhscl/s2i-base-rhel7/manifests/latest: net/http: TLS handshake timeout

Still seems like a temporary issue so will try re-running tests later

pkubatrh:
[test]

pkubatrh:
RHEL7 tests failing because of:

File "/opt/app-root/src/wsgi.py", line 11, in hello
    assert os.environ['LC_ALL'] == 'en_US.UTF-8'
AssertionError

Looking at the log this might be the cause:
/bin/bash: warning: setlocale: LC_ALL: cannot change locale (en_AU.UTF-8)
/bin/bash: warning: setlocale: LC_ALL: cannot change locale (en_AU.UTF-8)

Running the tests manually on a rhel 7.3 virtual machine didn't fail on me.

pkubatrh:
[test]

pvalena:
@hhorak, I do not know of any Ruby application that could be broken by different LANG/LC_ALL setting. And when some Ruby app does break, I think it should be fixed in the app itself (i.e. not hard-coded).
I do not think there is any use-case for this in case of Ruby :)

hhorak:
Since there is no feedback that such change would be beneficial in other languages, let's do it in Python for now. We can change this decision and move it to s2i-base any time later.

@GrahamDumpleton can you, please, rebase?

hhorak:
@GrahamDumpleton #159 changed same piece, one more rebase is needed.

hhorak:
[test]

hhorak:
I've tried github feature and it seems I was able to resolve the conflicts instead of you, let's see whether the tests still pass.

PR #168, opened 2016-12-07
Pass image ID for testing

omron93:
When running test/run use image ID instead of IMAGE_NAME. This allows parallel testing of images (for example parallel testing of PRs in Jenkins).

This is similar PR as for sclorg/mariadb-container#14. @hhorak Please take a look.

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -13,6 +13,9 @@
 # OPENSHIFT_NAMESPACES - Which of available versions (subdirectories) should be
 #       put into openshift/ namespace.
 
+set -o errexit
+set -o pipefail
+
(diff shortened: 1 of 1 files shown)


omron93:
Approach used in this PR is not possible for s2i based images (running `s2i build` in tests fails - using image ID instead of tag is not supported). So am closing this PR (and similar PRs for other container images) to keep building scripts same s2i based and non-s2i based images.

As mentioned in sclorg/postgresql-container#155, different approach could be used, so I will try to prepare something. But it would be helpful to have shared build scripts as mentioned in sclorg/postgresql-container#138.

Issue #157, opened 2016-11-21
Consider setting PYTHONUNBUFFERED.


GrahamDumpleton:
If a Python web application uses ``stdout`` for generating log messages, then the result can be buffered in memory and not output promptly such that the messages can be captured by Docker and OpenShift. The consequence of this is that there is potential for loss of log messages if the application crashes for some reason making it much harder to work out what the application was doing. The messages may also simply be delayed and not come out straight away.

It is not uncommon for people to set up the Python logging module to use ``stdout`` rather than ``stderr`` which exacerbates this problem.

Consideration should be given to setting the ``PYTHONUNBUFFERED`` environment variable as a default.

* https://docs.python.org/2/using/cmdline.html#envvar-PYTHONUNBUFFERED
* https://docs.python.org/2/using/cmdline.html#cmdoption-u

FWIW, Heroku sets ``PYTHONUNBUFFERED`` as a default as they likely have determined it is too confusing for users when they don't see their logged messages.

Note that this issue mainly affects gunicorn and hand rolled applications provided by ``app.py``. When using ``mod_wsgi-express`` it treats ``stdout`` the same as ``stderr`` and ensures that messages are always flushed when an EOL is encountered, with anything still remaining buffered at the end of a request which wasn't EOL terminated, being flushed automatically at the end of a request.

GrahamDumpleton:
Note that it would be possible to undo this as the default by a user setting in the ``.s2i/environment`` file:

```
PYTHONUNBUFFERED=
```

That is, in the ``Dockerfile`` use:

```
ENV PYTHONUNBUFFERED=1
```

A user can undo it by setting the same environment variable, but to an empty string, thus overriding the default. This works as the Python interpreter code is:

```
    if (!saw_unbuffered_flag &&
        (p = Py_GETENV("PYTHONUNBUFFERED")) && *p != '\0')
        Py_UnbufferedStdioFlag = 1;
```

so only enables unbuffered ``stdio`` if the environment variable is non empty.

ncoghlan:
For https://github.com/ncoghlan/srpminfo I'm seeing the problem where no logging output from the Python workers appears in the OpenShift console window, but setting `PYTHONUNBUFFERED=1` in the S2I environment didn't appear to make any difference.

GrahamDumpleton:
[...]

PR #169, opened 2016-12-08
Disable buffering of output from applications so logging appears immediately.

GrahamDumpleton:
This addresses #157 and ensures that any output from a Python web application will be captured in logs immediately. Without this option, the output being logged can be buffered in process, meaning that if the application crashes, log output can be lost. The delay in logging messages even in normal case can be confusing where user is expecting to see debug messages from an application.

This change affects ``gunicorn`` and ``waitress`` WSGI servers, Django development server, and any other custom Python web application using ``app.py``. It is not relevant to ``mod_wsgi-express`` or ``uWSGI`` as they already set up capture of ``stdout`` and ``stderr`` in a way that results in messages being output immediately with no buffering.

Note that there are no specific tests as is can't easily be validated with the test infrastructure. One would need to deploy the updated standalone test application with existing S2I builder and then new version. Then hit the application with siege. With the existing S2I builder the log messages are delayed and show up in batches as flushing of buffered output occurs. With changes, the log messages appear straight away.


pkubatrh:
LGTM
We should retest however as the last run seems to have failed on pulling OS images. I wonder if I can run some of the tests...

pkubatrh:
[test]

GrahamDumpleton:
[test]

hhorak:
Sorry for delay, can you rebase, please?

GrahamDumpleton:
[test]

GrahamDumpleton:
[test]

GrahamDumpleton:
@rhcarvalho Can you trigger test on this one as well.

omron93:
[test]

torsava:
Looks good to me! I'd say this is overdue for a merge.

Issue #166, opened 2016-12-06
The project/__init__.py file should be removed from app-home-test-app test.


GrahamDumpleton:
The ``app-home-test-app`` test is for testing ``APP_HOME`` environment variable. The idea is that the specified directory then becomes the root from which the application source code is picked up. In using ``APP_HOME`` and for the purposes of the test, the ``project`` directory doesn't need to be a Python package itself. The ``project/__init__.py`` file should be removed as not needed.

pkubatrh:
The way `APP_HOME` is implemented the current working directory is left unchanged and `project` is used by gunicorn as a package by running `gunicorn project.wsgi`, same as before. Hence the need for the `__init__.py` file. The only thing that really changed was that the `find` command only looks in the directory provided by `APP_HOME` (or the current working directory if unset) for the presence of `wsgi.py`

Is there any added value were we to change the working directory before executing an app instead?

GrahamDumpleton:
If that is what it doing, that wasn't the intent of the original request for adding the feature. The intent was that it would change directory. If all it is doing is using ``project.wsgi``, then you could have just set ``APP_MODULE`` to ``project.wsgi`` and didn't need to add a new variable. It isn't particular useful in the current form and doesn't solve the reason for asking for it. Will check properly tomorrow then to see what was done. I never validated what it was doing.

GrahamDumpleton:
Original comment where it was described what feature should do was https://github.com/sclorg/s2i-python-container/issues/134#issuecomment-235767928.

pkubatrh:
>If that is what it doing, that wasn't the intent of the original request for adding the feature. The intent was that it would change directory. If all it is doing is using project.wsgi, then you could have just set APP_MODULE to project.wsgi and didn't need to add a new variable.

The new variable was needed for Django, as I would not call its `manage.py` a module (and point to it using `APP_MODULE`) and `APP_FILE` would be picked up as a regular python script. But you have a point that for gunicorn it would make more sense to just use `APP_MODULE`.

Also must have missed the mention of changing the cwd in the original issue - thanks for pointing me in the right direction. Will look into modifying the code to change the cwd for both Django and gunicorn (and modifying the tests).

frenzymadness:
Is this issue still relevant? It seems that `APP_HOME` is used as an application home and working directory for years and there is no `__init__.py` file in `app-home-test-app/project` which this issue was created for.

PR #170, opened 2016-12-19
WIP: APP_HOME now changes working directory before running app

pkubatrh:
This PR modifies the APP_HOME variable feature in a number of ways.
Currently, the APP_HOME variable was only used as a point from which to do checks specific to assemble/run scripts. With this PR the scripts are now changing their current working directory to what is provided inside APP_HOME (+ appending old working directory to PYTHONPATH) before calling their respective applications in order to allow checks run from the appliaction to be done against APP_HOME.

As for the work in progress tag - I am at loss on how this should work with other application-path-modifying variables like APP_FILE or APP_MODULE. For APP_FILE, for the time being I have set them to be mutually exclusive as that made the most sense for me, but for APP_MODULE is a little bit different as it can be also set via `setup.py`. Taking this into account the script is now setting APP_HOME to its default value if APP_MODULE is found via `setup.py`. This seems however a bit counterinuitive to me so I would welcome some other suggestions.

@grahamdumpleton PTAL

Related: #134 #166

### DIFF
--- a/3.5/s2i/bin/run
+++ b/3.5/s2i/bin/run
@@ -15,6 +15,12 @@ function should_migrate() {
   is_django_installed && [[ -z "$DISABLE_MIGRATE" ]]
 }
 
+# Utility function to update PYTHONPATH and change cwd
+function change_cwd() {
+  export PYTHONPATH=${PYTHONPATH:-}:`pwd`
+  cd $1
+}
+
 # Guess the number of workers according to the number of cores
 function get_default_web_concurrency() {
   limit_vars=$(cgroup-limits)
@@ -32,17 +38,19 @@ function get_default_web_concurrency() {
   echo $default
 }
 
+APP_HOME=${APP_HOME:-.}
+
 app_file_check="${APP_FILE-}"
-APP_FILE="${APP_FILE:-app.py}"
+APP_FILE="${APP_FILE:-${APP_HOME}/app.py}"
 if [[ -f "$APP_FILE" ]]; then
   echo "---> Running application from Python script ($APP_FILE) ..."
-  exec python "$APP_FILE"
+  change_cwd $APP_HOME
+  exec python $(basename $APP_FILE)
 else
   test -n "$app_file_check" && (>&2 echo "ERROR: file '$app_file_check' not found.") && exit 1
 fi
 
-APP_HOME=${APP_HOME:-.}
-# Look for 'manage.py' in the directory specified by APP_HOME, or the current direcotry
+# Look for 'manage.py' in the directory specified by APP_HOME, or the current directory
 manage_file=$APP_HOME/manage.py
 
 if should_migrate; then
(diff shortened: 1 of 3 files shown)


GrahamDumpleton:
I can't look at this for over a week as about to go on holiday. I would like to have a close look at it though to make sure is what I was originally after.

3.5/s2i/bin/run
```diff
@@ -15,6 +15,12 @@ function should_migrate() {
   is_django_installed && [[ -z "$DISABLE_MIGRATE" ]]
 }
 
+# Utility function to update PYTHONPATH and change cwd
+function change_cwd() {
+  export PYTHONPATH=${PYTHONPATH:-}:`pwd`
```
hhorak: If not set, the result would begin with colon, is this expected? Should there be rather ```export PYTHONPATH=${PYTHONPATH:+$PYTHONPATH:}:`pwd` ```?

pkubatrh: Right, thanks for spotting this. Although the problematic colon found your way into your version as well. ``` export PYTHONPATH=${PYTHONPATH:+$PYTHONPATH:}`pwd` ``` seems to be what we want.

pkubatrh:
@hhorak Yes I am planning to merge the changes with all of the versions once all is reviewed/updated.

GrahamDumpleton:
When I did this in my own S2I builder, all I did was to change the working directory once before any of the checks were made, plus add ``/opt/app-root/src`` to ``PYTHONPATH``. 

Why do you delay changing the directory until the point of starting the WSGI server or application?

One not just do that around where you have:

```
APP_HOME=${APP_HOME:-.}
```

Thus instead of that and everything else you added to ``run``, just do at that point something like:

```
if [[ "$APP_HOME" != "" ]]; then
    PYTHONPATH=/opt/app-root/src:$PYTHONPATH
    cd $APP_HOME
fi
```

We also need to get together a proper check list of application examples that this needs to be able to work with that are commonly seen in the real world so can work out whether catered.

pkubatrh:
>Why do you delay changing the directory until the point of starting the WSGI server or application?

I opted for this because of the `setup.py` check since right now we are expecting `setup.py` to be located in the repository's root.
I guess we could do this one check against the root even when in APP_HOME, but it might be (again) a little bit unintuitive. There has also been some mentions of completely removing the `setup.py` check in one of the issues but wouldn't that be too much of a behavioral change?

Btw thanks for looking at this during your time off.

3.5/s2i/bin/run
hhorak: Ah, yes.

GrahamDumpleton:
The setup.py and requirements.txt should always be from top level directory regardless of what APP_HOME is. If they were also in a sub directory they should use contextDir in build configuration.

GrahamDumpleton:
Just to clarify, not talking about changing directory in assemble, only in run at point I suggested. The manage.py check for collectstatic can have special case of looking in APP_HOME as was already done.

Anyway am soon to be without Internet for a week at worst, so will need to look when get back.

pkubatrh:
>The setup.py and requirements.txt should always be from top level directory regardless of what APP_HOME is. 

Indeed. What I am proposing is not putting those files into whatever APP_HOME will be set to, but (after changing the working directory to APP_HOME) popping back to the root directory to check for `setup.py` and returning to APP_HOME if not found.

GrahamDumpleton:
Are you talking about in run or assemble? Are you talking about the querying of the module name for run only?

Anyway, my ship has literally sailed so will loose Internet soon as will be too far out to sea. :-)

pkubatrh:
Only the module query in run. I don't think there is anything that needs changing in assemble at this point.

pkubatrh:
Closing in favour of #228

PR #174, opened 2017-01-03
Add support for app.sh script so have easy control over how application launched.

GrahamDumpleton:
This address issue https://github.com/sclorg/s2i-python-container/issues/126.

Note that you should merge change https://github.com/sclorg/s2i-python-container/pull/165 first as this depends on the test case added in that for uWSGI, which uses an ``app.sh`` file executed from an ``app.py`` file. When the change here is then merged, the ``app.sh`` file in the test will take precedence and the ``app.sh`` will be executed directly with ``app.py`` being bypassed.

Note that this change doesn't do anything about ``APP_HOME`` as existing code didn't do anything with ``APP_HOME`` for ``app.py`` either. Arguably it should, but this is being dealt with in separate issues and PR.

Also note that although the change uses ``APP_SCRIPT`` and refers to it as application script file, encouraging it as being used for a shell script, technically it could be any executable script file, or even a compiled application. A generic name for the environment variable could be used such as ``APP_PROGRAM`` if really want to reflect that, but probably better to encourage it being just used for a shell script which then in turn can execute anything.

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -214,7 +214,7 @@ following ways, in precedence order:
   This is the most general way of executing your application. It will be
   used in the case where you specify a path to an executable script file
   via the `APP_SCRIPT` environment variable, defaulting to a file named
-  `app.sh` if it exists. The script is executed directory to launch your
+  `app.sh` if it exists. The script is executed directly to launch your
   application.
 
 Hot deploy
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -92,6 +92,12 @@ Environment variables
 To set these environment variables, you can place them as a key value pair into
 a `.s2i/environment` file inside your source code repository.
 
+* **APP_SCRIPT**
+
+    Used to run the application from a script file.
+    This should be a path to a script file (defaults to `app.sh`) that will be
+    run to start the application.
+
 * **APP_FILE**
 
     Used to run the application from a Python script.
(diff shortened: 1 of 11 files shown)


pkubatrh:
#165 has been merged, please rebase so that we can use the tests in CI

pkubatrh:
[test]

2.7/README.md
```diff
@@ -197,10 +203,19 @@ following ways, in precedence order:
...
+  Python script via the `APP_FILE` environment variable, defaulting to a
+  file named `app.py` if it exists. The script is passed to a regular
+  Python interpreter to launch your application.
+
+* **Application script file**
+
+  This is the most general way of executing your application. It will be
+  used in the case where you specify a path to an executable script file
+  via the `APP_SCRIPT` environment variable, defaulting to a file named
+  `app.sh` if it exists. The script is executed directory to launch your
```
pkubatrh: I guess directory -> directly?

pkubatrh:
One more thing - this part should also get changed to accommodate for the new APP_SCRIPT variable
https://github.com/GrahamDumpleton/s2i-python-container/blob/7d09b2ba22c60a1c012fed00b5f11cc13375e7f6/2.7/s2i/bin/run#L102

GrahamDumpleton:
Added ``APP_SCRIPT`` environment variable to final error message.

hhorak:
LGTM.

hhorak:
[test]

Issue #122, opened 2016-06-08
Lack of libffi causes some Python packages to fail to be installed.


GrahamDumpleton:
This appears to affect at least `ndg-httpsclient` a dependency required by Lektor site generator.

```
Running setup.py install for ndg-httpsclient
Skipping installation of /opt/app-root/src/.local/lib/python3.4/site-packages/ndg/__init__.py (namespace package)

Installing /opt/app-root/src/.local/lib/python3.4/site-packages/ndg_httpsclient-0.4.1-py3.4-nspkg.pth
Installing ndg_httpclient script to /opt/app-root/src/.local/bin
Running setup.py install for cryptography
Package libffi was not found in the pkg-config search path.
Perhaps you should add the directory containing `libffi.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libffi' found
Package libffi was not found in the pkg-config search path.
Perhaps you should add the directory containing `libffi.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libffi' found
Package libffi was not found in the pkg-config search path.
Perhaps you should add the directory containing `libffi.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libffi' found
Package libffi was not found in the pkg-config search path.
Perhaps you should add the directory containing `libffi.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libffi' found
Package libffi was not found in the pkg-config search path.
Perhaps you should add the directory containing `libffi.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libffi' found
c/_cffi_backend.c:15:17: fatal error: ffi.h: No such file or directory
#include <ffi.h>
^
compilation terminated.
Traceback (most recent call last):
File "/opt/rh/rh-python34/root/usr/lib64/python3.4/distutils/unixccompiler.py", line 126, in _compile
extra_postargs)
File "/opt/rh/rh-python34/root/usr/lib64/python3.4/distutils/ccompiler.py", line 909, in spawn
spawn(cmd, dry_run=self.dry_run)
File "/opt/rh/rh-python34/root/usr/lib64/python3.4/distutils/spawn.py", line 36, in spawn
_spawn_posix(cmd, search_path, dry_run=dry_run)
File "/opt/rh/rh-python34/root/usr/lib64/python3.4/distutils/spawn.py", line 162, in _spawn_posix
% (cmd, exit_status))
distutils.errors.DistutilsExecError: command 'gcc' failed with exit status 1

[...]

PR #175, opened 2017-01-10
Add required packages for SAML2

mczp:
pySAML2 requires several packages, one of them (python-cryptography) requires openssl-devel, libffi (cffi) and xmlsec1 (which depends on the missing libltdl).
To my knowledge there are other packages requiring cffi and cryptography, thus this change should benefit everyone using them, while adding only a few Mb to the image.

### DIFF
--- a/2.7/Dockerfile
+++ b/2.7/Dockerfile
@@ -16,7 +16,7 @@ LABEL io.k8s.description="Platform for building and running Python 2.7 applicati
 
 RUN yum install -y centos-release-scl && \
     INSTALL_PKGS="libjpeg-turbo libjpeg-turbo-devel python27 python27-python-devel python27-python-setuptools 
-	python27-python-pip nss_wrapper httpd httpd-devel atlas-devel gcc-gfortran libffi-devel libtool-ltdl" && \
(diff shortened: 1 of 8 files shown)


mczp:
This should solve #122 as well

pkubatrh:
[test]

pkubatrh:
From what I can see, the openssl-devel package is already present in each of the python images.

hhorak:
openssl-devel is already installed in s2i-base (https://github.com/sclorg/s2i-base-container/blob/master/Dockerfile), so there is no need to list it in python specifically.

hhorak:
@mczp, cau you be more specific about how many MBs the difference is, please?

pkubatrh:
@hhorak From what I have checked yesterday the difference amounted to about 100kB (libffi + libltdl).

mczp:
@hhorak yes the final squashed image size should be about 100 Kb bigger, in my latest commit I removed openssl-devel

hhorak:
[test]

hhorak:
I think it's worth it then, let's do it. LGTM after tests pass.

Issue #178, opened 2017-01-27
Build with the latest image generates "Operation not permitted" errors in assemble


tmckayus:
The fix-permissions calls in the assemble script generate warnings on the python files.
 
Here is the command I ran, and the output from the build script (could be any python source):

  oc new-app --name testbuild docker.io/centos/python-27-centos7~https://github.com/tmckayus/play.git

Build logs:

Cloning "https://github.com/tmckayus/play.git" ...
	Commit:	bf8668eef2718c42339bcfae4a42f41a1342f1c2 (Add a spark-conf subdir to play with custom spark config)
	Author:	Trevor McKay <tmckay@redhat.com>
	Date:	Wed Jan 25 11:46:54 2017 -0500

---> Installing application source ...
chgrp: changing group of '/opt/app-root/include/python2.7': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/UserDict.py': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/_abcoll.py': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/_weakrefset.py': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/abc.py': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/codecs.py': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/config': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/copy_reg.py': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/encodings': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/fnmatch.py': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/genericpath.py': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/lib-dynload': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/linecache.py': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/locale.py': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/ntpath.py': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/os.py': Operation not permitted
chgrp: changing group of '/opt/app-root/lib/python2.7/posixpath.py': Operation not permitted
[...]

PR #181, opened 2017-02-09
Enable incremental builds

luiscoms:


3.5/test/run
```diff
@@ -168,7 +168,7 @@ test_application() {
 
 # Since we built the candidate image locally, we don't want S2I attempt to pull
 # it from Docker hub
-s2i_args="--force-pull=false"
+s2i_args="--incremental=true --force-pull=false"
```
bparees: i don't think this is sufficient to test this new functionality.  it really needs to:

1) build the app image
2) build the app image again passing --incremental and confirm (via the log output, perhaps) that the artifacts from (1) were reused successfully instead of being re-downloaded.

bparees:
@GrahamDumpleton you are probably the best person to review this from an implementation perspective.

GrahamDumpleton:
Will this even work with latest?

Now that using a Python virtual environment, the per user site-packages directory under ``$HOME/.local`` isn't even used any more. The better way to handle incremental builds in my experiments has been to generate packages as wheels when installing them and then the wheelhouse directory is what is copied across as an artefact. That means on subsequent install, it will use packages from the wheelhouse when installing.

I will look over changes, but also go back and find how I was doing the wheelhouse support that I have in my own S2I builder.

One benefit of the wheelhouse approach is that you can also copy in the wheelhouse directory in a build from an image source. You can thus have a separate build, the sole purpose of which is to construct a wheelhouse directory to be used for other builds.

I wrote blog posts about this stuff at:

* https://blog.openshift.com/using-image-source-reduce-build-times/
* http://blog.dscpl.com.au/2016/03/speeding-up-docker-build-times-for.html

BTW, I thought I heard that incremental builds were possibly going to be deprecated.

bparees:
> BTW, I thought I heard that incremental builds were possibly going to be deprecated.

not that i'm aware of.  extended builds, perhaps.

GrahamDumpleton:
Okay, may well be extended builds since that appears to still be made available as technology preview only.

luiscoms:
I just test this code on my child image https://github.com/luiscoms/s2i-python-35

luiscoms:
I ran `make test TARGET=centos7 VERSION=3.5` and on some point of build I got

```
Installed /opt/app-root/lib/python3.5/site-packages/gunicorn-19.6.0-py3.5.egg
Finished processing dependencies for testapp==0.1
ERROR: chgrp: changing group of ‘/opt/app-root/lib/python3.5/tokenize.py’: Operation not permitted
ERROR: chgrp: changing group of ‘/opt/app-root/lib/python3.5/random.py’: Operation not permitted
ERROR: chgrp: changing group of ‘/opt/app-root/lib/python3.5/shutil.py’: Operation not permitted
ERROR: chgrp: changing group of ‘/opt/app-root/lib/python3.5/re.py’: Operation not permitted
```

GrahamDumpleton:
See #178.

Known issue and was reported that would occur when switched to virtual environments. It doesn't cause any problems. No resolution on whether base image should change, or whether just do fix ups to permissions in Python image differently to avoid the problem due to how base image function does it.

ibotty:
Are there any pointers regarding deprecation of extended builds? Is there a (proposed) replacement to have a minimal runtime image but a big builder image?

bparees:
> Are there any pointers regarding deprecation of extended builds? Is there a (proposed) replacement to have a minimal runtime image but a big builder image?

the suggested way to separate your build-time deps from your runtime deps is to use chained builds as described here:
https://docs.openshift.org/latest/dev_guide/builds/advanced_build_operations.html#dev-guide-chaining-builds

and lest anyone be confused, incremental builds are not the same thing as extended builds and incremental builds are NOT being deprecated.

frenzymadness:
Is there anybody who is willing to rebase the patch?

GrahamDumpleton:
There is still an outstanding question as to whether this even does what is required. The `.local` directory would not be used as far as I know, so not sure what the point of preserving it is. Also, when packages are installed, `pip` is told not to cache packages, so there is no cache to preserve either. I sort of think this should be closed and if there really is interest a new PR created which properly outlines what is required and shows how changes address it.

frenzymadness:
@GrahamDumpleton thank you for the comment. I agree.

PR #184, opened 2017-03-06
RFC: Add npm for front-end development

jsvgoncalves:
* Following the commit on [redhat-developer/s2i-dotnetcore](https://github.com/redhat-developer/s2i-dotnetcore/commit/b8992565ca6f2724103cfdfbc049a1806929bfe4),  adding npm to the images allows for modern web applications to rely on javascript tools to build the app.

Currently it's hard (if not impossible) to pull in node and npm during build time, it takes too long and often times out. However, typically, CSS and JS files need pre-deployment processing such bundling, minifying, etc. 

In line with redhat-developer/s2i-dotnetcore#35, it would be better to avoid referring new developers to "Create Your Own Image" guides and instead provide npm directly.

This PR is serves both as an RFC and as a starting point for the proposal.

### DIFF
--- a/2.7/Dockerfile
+++ b/2.7/Dockerfile
@@ -7,7 +7,7 @@ MAINTAINER SoftwareCollections.org <sclorg@redhat.com>
 EXPOSE 8080
 
 ENV PYTHON_VERSION=2.7 \
-    PATH=$HOME/.local/bin/:$PATH \
+    PATH=$HOME/.local/bin/:/opt/app-root/node_modules/.bin:$PATH \
     PYTHONIOENCODING=UTF-8 \
     LC_ALL=en_US.UTF-8 \
     LANG=en_US.UTF-8 \
@@ -21,7 +21,7 @@ LABEL io.k8s.description="Platform for building and running Python 2.7 applicati
 
 RUN yum install -y centos-release-scl && \
     INSTALL_PKGS="libjpeg-turbo libjpeg-turbo-devel python27 python27-python-devel python27-python-setuptools \ 
-	python27-python-pip nss_wrapper httpd httpd-devel atlas-devel gcc-gfortran libffi-devel libtool-ltdl" && \
+	python27-python-pip nss_wrapper httpd httpd-devel atlas-devel gcc-gfortran libffi-devel libtool-ltdl rh-nodejs4-npm" && \
     yum install -y --setopt=tsflags=nodocs --enablerepo=centosplus $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     # Remove centos-logos (httpd dependency, ~20MB of graphics) to keep image
--- a/2.7/test/run
+++ b/2.7/test/run
@@ -8,7 +8,7 @@
 #
 IMAGE_NAME=${IMAGE_NAME:-openshift/python-27-centos7-candidate}
 
-declare -a WEB_APPS=({standalone,setup,django,numpy,app-home,uwsgi,locale}-test-app)
+declare -a WEB_APPS=({standalone,setup,django,numpy,app-home,uwsgi,locale,npm}-test-app)
 
 # TODO: Make command compatible for Mac users
 test_dir="$(readlink -zf $(dirname "${BASH_SOURCE[0]}"))"
@@ -163,6 +163,20 @@ test_connection() {
     sleep $sleep_time
   done
   return $result
+}
+
+# !TODO: This still needs to be called
+test_npm_application() {
+  # Verify that NPM packages from package.json are installed
+  local_module_name="local_module"
+  npm_packages="$(npm list | grep $local_module_name)"
+
+  if [[ $npm_packages != *"$local_module_name"* ]]; then
+    echo "ERROR[Npm modules not installed]"
+    return 1
+  else
+    return 0
+  fi
 }
 
 test_application() {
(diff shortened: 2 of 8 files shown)


rhcarvalho:
[test]

omron93:
[test]

omron93:
[test]

2.7/test/run
```diff
@@ -8,7 +8,7 @@
 #
 IMAGE_NAME=${IMAGE_NAME:-openshift/python-27-centos7-candidate}
 
-declare -a WEB_APPS=({standalone,setup,django,numpy,app-home,uwsgi,locale}-test-app)
+declare -a WEB_APPS=({standalone,setup,django,numpy,app-home,uwsgi,locale, npm}-test-app)
```
omron93: Space before 'npm' can't be there.

jsvgoncalves:
@omron93 nice catch, thanks

2.7/test/run
jsvgoncalves: @omron93 nice catch, thanks

jsvgoncalves:
@omron93 @hhorak RFC. Let me know what you think of the change (can we proceed?) before I apply it to `3.x` and fix the tests.

hhorak:
We chatted with @jsvgoncalves and I'm now pretty convinced that NodeJS is more and more popular between even python developers. On the other hand, installing npm will bring additional 43MB to the image, which is quite a lot. So one thing I'm thinking about is having a common layer for nodejs and python containers, so we can at least share those 40+MB across more images..

@bparees @torsava what are your opinions?

bparees:
@hhorak it sounds ok to me.

2.7/Dockerfile
```diff
@@ -7,7 +7,7 @@ MAINTAINER SoftwareCollections.org <sclorg@redhat.com>
 EXPOSE 8080
 
 ENV PYTHON_VERSION=2.7 \
-    PATH=$HOME/.local/bin/:$PATH \
+    PATH=$HOME/.local/bin/:/opt/app-root/node_modules/.bin:$PATH \
```
torsava: Is there a reason to use `/opt/app-root/node_modules/.bin` instead of `$HOME/node_modules/.bin/:$HOME/.npm-global/bin/` as it is used in [s2i-nodejs-container](https://github.com/sclorg/s2i-nodejs-container/blob/master/4/Dockerfile#L18)?

torsava:
@hhorak I agree, a common layer would be good.

2.7/Dockerfile
jsvgoncalves: @torsava no reason other than taking it from  [redhat-developer/s2i-dotnetcore](https://github.com/redhat-developer/s2i-dotnetcore/commit/b8992565ca6f2724103cfdfbc049a1806929bfe4), but I will dig deeper into this if I get green light to actually add npm into this image

pkubatrh:
@hhorak I think creating another layer with npm is a much better solution than just adding it to the images with such a size increase. This would make the rebuilding of the images a bit more complex but nothing we cannot manage.
Another alternative would be to add npm to the underlying s2i base image without adding another layer on top of it. The question in this case is, however, whether the other images built upon s2i-base would also profit from it.

jsvgoncalves:
@pkubatrh I agree that it would be even better to have it as a part of s2i base, as indeed it is a fairly common tool and pretty much independent of whatever the back-end stack is (Java, PHP, .NET, Ruby...).

hhorak:
> Another alternative would be to add npm to the underlying s2i base image without adding another layer on top of it. The question in this case is, however, whether the other images built upon s2i-base would also profit from it.

I think this would be unused in case of images like nginx, passenger, httpd, etc.. but in those cases even the devel packages from s2i-base are not very usable, so we might think about one more layer without the devel packages..

jsvgoncalves:
> I think this would be unused in case of images like nginx, passenger, httpd, etc..

I didn't know such images were also using `s2i-base`. But as they do, I agree with you that it makes no sense to add `npm` to them, and it would be better to create a shared `s2i-npm-layer` or the likes instead.

GrahamDumpleton:
FWIW, I am +1 on adding npm. I am already having to ignore the default Python S2I builder for some work related to running JupyterHub on OpenShift because it lacks npm.

torsava:
So it looks there's no opposition to the common npm layer in the base container: sclorg/s2i-base-container/issues/115. 

@jsvgoncalves could you make the PR against [s2i-base-container](https://github.com/sclorg/s2i-base-container)?

torsava:
This was merged into the s2i-base-container in PR: https://github.com/sclorg/s2i-base-container/pull/132, and into this container in PR: https://github.com/sclorg/s2i-python-container/pull/209.

GrahamDumpleton:
One note I will make about this which is a slight pain is that you can't install npm packages globally as see some people want because global directory for npm hasn't been made writable to non root user for purposes of the S2I builder to make things more convenient. It means you have to set environment variable in ``.s2i/environment`` to override location. I used:

```
NPM_CONFIG_PREFIX=/opt/app-root
```
because ``/opt/app-root/bin`` is already in the ``PATH`` and so programs installed using ``npm`` would already be found without having to fiddle with ``PATH`` in custom ``assemble`` or ``run`` script wrappers.

If don't want to make global directories writable, consideration could be made to setting ``NPM_CONFIG_PREFIX`` to some different value permanently and ensure the ``bin`` directory in it is added to ``PATH`` as part of image.

Anyway, leave this here so people can find details about it and perhaps create separate new issue if keeps coming up as a problem.

PR #185, opened 2017-03-13
Add Fedora Dockerfile

hhorak:
Do not test last part of the version, it is compatible and updated in the same image

### DIFF
--- a/3.5/Dockerfile.fedora
+++ b/3.5/Dockerfile.fedora
@@ -5,7 +5,7 @@ FROM 25/s2i-base
 LABEL MAINTAINER SoftwareCollections.org <sclorg@redhat.com>
 
 ENV NAME=python \
-    VERSION=3.5 \
+    VERSION=0 \
     RELEASE=1 \
     ARCH=x86_64
 
@@ -13,18 +13,20 @@ LABEL BZComponent="$NAME" \
       Name="$FGC/$NAME" \
       Version="$VERSION" \
       Release="$RELEASE.$DISTTAG" \
-      Architecture="$ARCH"
+      Architecture="$ARCH" \
+      Usage="s2i build file:///your/app 25/python your-app" \
+      Help="help.1"
 
 EXPOSE 8080
 
-ENV PYTHON_VERSION=$VERSION \
+ENV PYTHON_VERSION=3.5 \
     PATH=$HOME/.local/bin/:$PATH \
     PYTHONIOENCODING=UTF-8 \
     LC_ALL=en_US.UTF-8 \
     LANG=en_US.UTF-8 \
     PIP_NO_CACHE_DIR=off
 
-ENV SUMMARY="Platform for building and running Python PYTHON_VERSION applications" \
+ENV SUMMARY="Platform for building and running Python $PYTHON_VERSION applications" \
     DESCRIPTION="Python $PYTHON_VERSION available as docker container, is a base platform for \
 building and running various Python $PYTHON_VERSION applications and frameworks. \
 Python is an easy to learn, powerful programming language. It has efficient high-level \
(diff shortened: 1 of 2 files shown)


hhorak:
@torsava What do you think?

torsava:
I sent a PR to your PR with modified `hack/` scripts to allow building of the fedora Dockerfile.

The Dockerfile itself looks good to me. Nevertheless, as there's not yet a Fedora-based s2i-base image (correct me if I'm wrong), I wasn't able to test it. Should I look into creating one?

pkubatrh:
>Nevertheless, as there's not yet a Fedora-based s2i-base image (correct me if I'm wrong), I wasn't able to test it. Should I look into creating one?

Seems like the Fedora s2i-base image is available in the candidate registry at least, should be ok for testing imo.
`docker pull candidate-registry.fedoraproject.org/f25/s2i-base`

torsava:
@hhorak: I've created a PR on your repo which adds some commits to this PR: hhorak/s2i-python-container/pull/1:
 - Modifies the `hack/` scripts to allow `TARGET=fedora`,
 - moves the man help page to `/usr/share/man/man1` so it can be opened by running `man help`,
 - and fixes virtualenv call.

I think it would be better to have the source code for the `help.1` file in this repo and compile it during the build of the container, so it can be modified if necessary.

hhorak:
@torsava Thank you Tomas! I agree with all suggestions except the man page changes, it's more explained in https://github.com/hhorak/s2i-python-container/pull/1... I'm OK with other suggestions.

torsava:
Looks good to me!

omron93:
I have one question about changes in "build scripts". @hhorak Will we build fedora based images?

If not, it seems confusing to me, that we provide a way to do it easily.

Otherwise, image can be built:
`pushd 3.5; docker build -t whatever_name . -f Dockerfile.fedora; popd`

and tested:
`IMAGE_NAME=whatever_name 3.5/test/run`

torsava:
> I have one question about changes in "build scripts". @hhorak Will we build fedora based images?

I believe we do want to build Fedora images—the Fedora-based s2i-base-container is already available:

    $ docker pull registry.fedoraproject.org/f25/s2i-base:latest

Or am I misunderstanding you, @omron93?

hhorak:
I think we should build them for testing PRs, so we make sure the PR does not break Fedora dockerfile.. But production images will be built in koji, using OSBS, so Fedora dockerfiles should be IMO handled similarly as RHEL dockerfiles.. Does that make sense?

omron93:
>  think we should build them for testing PRs, so we make sure the PR does not break Fedora dockerfile.

OK.

>Or am I misunderstanding you, @omron93?

As @hhorak said, fedora images are built in koji.

hhorak:
Ha, I've realized that a better way to add the fedora-build part would be adding to https://github.com/sclorg/container-common-scripts, and then include the common directory similar as it is done in https://github.com/sclorg/postgresql-container..
So, please, do not merge this PR yet.

omron93:
Support for Fedora images building is already in container-common-scripts + see PR for using these build scripts for python image - #193

@hhorak Please remove build scripts changes from this PR. Thanks.

hhorak:
Rebased and changes in build scripts removed from this PR. Thanks for getting it done in the container-common-scripts!

torsava:
I've added a commit with the regenerated help.1 file containing latest changes, and also put the command that generates it into the comment in the Fedora Dockerfile.

torsava:
I have two proposals (and volunteer for their eventual execution):

1. Why not add the help.1 file to the CentOS and RHEL images as well (in a separate PR)?

2.  Relating to the comment in the Dockerfile:

        # Add help.1 file, that is generated using go-md2man manually (at this point)
        # TODO: we should make it transparent for the contributors

How about we start a `contributor-README.md` file (or similar) add add a note about rebuilding the help file there?

omron93:
> Why not add the help.1 file to the CentOS and RHEL images as well (in a separate PR)?

In "official" RHEL based images `help.1` is generated and added automatically by OSBS build system.

So possible solution would be to generate `help.1` from `README.md` automatically by build scripts too. See https://github.com/sclorg/container-common-scripts/pull/7 . This affects/fixes all images using common scripts.

torsava:
Code from this pull request was already merged as part of pull request #215, therefore I'm closing this!

PR #191, opened 2017-04-19
Skip default assignment to APP_FILE if it is set to null

### LINKED ISSUE #190: No way to not run app.py if it exists
When trying to run a simple flask app through gunicorn, I encountered unusual behaviour where my app was running on the flask development server instead of following anything in the wsgi.py file. It was caused because I had unimaginatively called my app "app.py". I couldn't find any way to avoid this behaviour other than to rename the file to "flaskapp.py" instead.

This seems to be a side-effect of resolving https://github.com/sclorg/s2i-python-container/issues/56 which rules out the simplest solution of reversing the order. Bearing this in mind, I do not know if there even is an ideal solution without giving up on the automagic XKCD1654 style approach.

The code added by https://github.com/sclorg/s2i-python-container/issues/86 unfortunately removes the workaround of setting the $APP_FILE to something nonexistent.

torsava:
If APP_FILE is set to any value including null ( APP_FILE= ) then the default
assignment to value of `app.py` is skipped.

This will allow people to have a script named `app.py` but run it in a
different manner (e.g. through gunicorn—fixes #190) by setting APP_FILE to null.

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -95,13 +95,13 @@ a `.s2i/environment` file inside your source code repository.
 * **APP_SCRIPT**
 
     Used to run the application from a script file.
-    This should be a path to a script file (defaults to `app.sh`) that will be
+    This should be a path to a script file (defaults to `app.sh` unless set to null) that will be
     run to start the application.
 
 * **APP_FILE**
 
     Used to run the application from a Python script.
-    This should be a path to a Python file (defaults to `app.py`) that will be
+    This should be a path to a Python file (defaults to `app.py` unless set to null) that will be
     passed to the Python interpreter to start the application.
 
 * **APP_MODULE**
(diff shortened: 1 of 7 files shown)


hhorak:
[test]

torsava:
This PR fixes issue #190, I asked the issuer to test whether it resolves the problem an they confirmed it does.

For the sake of consistency *and* functionality, I think `APP_SCRIPT` would merit a similar change as it is used in the same manner as `APP_FILE`. **Would anyone be opposed to this?**

pkubatrh:
>For the sake of consistency and functionality, I think APP_SCRIPT would merit a similar change as it is used in the same manner as APP_FILE. 

Sounds reasonable. Additionally both cases should be documented.

torsava:
Fixed `APP_SCRIPT` as well and updated the `README.md`.

torsava:
[test]

pkubatrh:
lgtm, merged

Issue #177, opened 2017-01-25
Tests needed for python's virtual environment added in a recent PR


pkubatrh:
The use of Python's virtual environment has been added in #159 alas without any tests included. It would be nice to have some at least.

The PR has also added an option to upgrade pip to the latest version.
I have proposed adding a test that would try to install a Python wheel after doing a pip upgrade, however we will be getting pip rebased to version 8.1.2 (https://bugzilla.redhat.com/show_bug.cgi?id=1344674) which should make the default pip able to install wheels without the upgrade so the test would not help with verifying the change.

GrahamDumpleton:
Ignore the test for updating ``pip``, which didn't know how to add with way current test harness work, the use of the virtual environment is tested indirectly by all existing tests that install packages. If there was a significant issue, the existing tests would work. One could add extra checks to validate ``which python`` returns Python from the virtual environment but how do you add it, using ``.s2i/bin`` override scripts, since no concept of action hooks to allow easy insertion of scripts at any step.

torsava:
I've looked into adding the tests for the virtualenv and pip upgrade.

Please take a look: PR #192.

torsava:
Tests were added in PR #192, closing.

PR #192, opened 2017-04-24
Add virtualenv and UPGRADE_PIP_TO_LATEST tests

torsava:
I've taken a look at issue #177 and tried to come up with a way to test the virtual env and UPGRADE_PIP_TO_LATEST functionality.

Due to the way the tests are handled in s2i, they needs to be made into a test-app that will then be able to respond to HTML requests. That part of the test would be useless for testing the above mentioned and thus I opted for bundling the test with an already existing test-app. What do you think?


torsava:
@pkubatrh @GrahamDumpleton Could you please take a look?

pkubatrh:
Seems reasonable. Could you add the test case changes to other python versions as well?

torsava:
@pkubatrh Thanks for looking into it! I've amended the commits to test all the three active versions.

pkubatrh:
Thanks, merging.

PR #195, opened 2017-05-26
Update Dockerfiles

omron93:
Provide same labels for all base image variants (CentOS, RHEL, Fedora)
 - follow recomended labels from Project Atomic Container best practices
 - remove architecture label for RHEL based images (this label is set in RHEL image - no need to define it again)
 - use label 'maintainer' (MAINTAINER instruction is deprecated)
                                           
(don't use separate instruction for each label)

@hhorak @praiskup @pkubatrh Please take a look and merge.

### DIFF
--- a/2.7/Dockerfile
+++ b/2.7/Dockerfile
@@ -2,8 +2,6 @@
 # applications.
 FROM centos/s2i-base-centos7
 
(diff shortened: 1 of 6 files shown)


torsava:
LGTM.

PR #196, opened 2017-05-30
Delete old unnecessary instructions in assemble

### LINKED ISSUE #182: Is /tmp/pip_default_default directory still used.
The ``assemble`` scripts remove ``/tmp/pip_build_default`` but is that directory even created by anything now. Can't see any other mention of it, such as in an environment variable for ``pip``, so not sure why it would use it.

```
# remove pip temporary directory
rm -rf /tmp/pip_build_default
```

torsava:
The temporary build directory at this location is unused since this container
switched to using virtualenv. The new default build directory is at
`/opt/app-root/build/` and it's removed automatically by pip.

Resolves #182.

### DIFF
--- a/2.7/s2i/bin/assemble
+++ b/2.7/s2i/bin/assemble
@@ -59,6 +59,3 @@ fi
 fix-permissions ./
 # set permissions for any installed artifacts
 fix-permissions /opt/app-root
-
-# remove pip temporary directory
-rm -rf /tmp/pip_build_default
--- a/3.4/s2i/bin/assemble
+++ b/3.4/s2i/bin/assemble
@@ -63,6 +63,3 @@ fi
 fix-permissions ./
 # set permissions for any installed artifacts
 fix-permissions /opt/app-root
-
-# remove pip temporary directory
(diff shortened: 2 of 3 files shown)


pkubatrh:
I would not change 3.3 as it is deprecated, otherwise lgtm.

torsava:
> I would not change 3.3 as it is deprecated, otherwise lgtm.

Thanks!

To the 3.3: Can we remove its directory in a new PR? I don't recall reading why it was kept.

pkubatrh:
Most likely for archival purposes? @hhorak ^

hhorak:
Exactly, we kept the directories for EOL collections for anyone who would be fine to use them even as unsupported. But we don't need to update them any more. Those should stay untouched basically.. we may even remove them at some point, but I'd give it more time..
@bparees, any objections for removing the directories for EOL in the future? (say a year after particular collection was made EOL?)

torsava:
Ah, thanks for the explanation! I amended the commit to remove the 3.3 change.

pkubatrh:
Thanks! Merging

Issue #198, opened 2017-06-23
Changing permissions on src twice.


GrahamDumpleton:
The ``assemble`` script does:

```
# Fix source directory permissions
fix-permissions ./
# set permissions for any installed artifacts
fix-permissions /opt/app-root
```

The current working directory at the time is always ``/opt/app-root/src``. Thus this is equivalent to:

```
# Fix source directory permissions
fix-permissions /opt/app-root/src
# set permissions for any installed artifacts
fix-permissions /opt/app-root
```

Why not just use:

```
# set permissions for any installed artifacts
fix-permissions /opt/app-root
```

and avoid doing it twice for ``/opt/app-root/src``.

torsava:
I agree, makes sense.

PR #200, opened 2017-06-30
Prevent changing permissions on src twice

mcyprian:
The current working directory is always /opt/app-root/src when fix-permissions functions is called in assemble script. It is enough to call it once with the parent directory /opt/app-root as an argument.

Resolves issue #198.

### DIFF
--- a/2.7/s2i/bin/assemble
+++ b/2.7/s2i/bin/assemble
@@ -55,7 +55,5 @@ if should_collectstatic; then
   )
 fi
 
-# Fix source directory permissions
-fix-permissions ./
 # set permissions for any installed artifacts
 fix-permissions /opt/app-root
--- a/3.3/s2i/bin/assemble
+++ b/3.3/s2i/bin/assemble
@@ -59,8 +59,6 @@ if should_collectstatic; then
   )
 fi
 
-# Fix source directory permissions
-fix-permissions ./
 # set permissions for any installed artifacts
 fix-permissions /opt/app-root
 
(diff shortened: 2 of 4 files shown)


pkubatrh:
[test]

omron93:
Thanks for the fix.

LGTM + tests are passing + same solution as suggested in mentioned issue. So merging.

jfmatth:
Is this supposed to eliminate the issue below for django projects?

```
chgrp: changing group of ‘/opt/app-root/lib64/python3.5/operator.py’: Operation not permitted
chgrp: changing group of ‘/opt/app-root/lib64/python3.5/os.py’: Operation not permitted
chgrp: changing group of ‘/opt/app-root/lib64/python3.5/_collections_abc.py’: Operation not permitted
chgrp: changing group of ‘/opt/app-root/lib64/python3.5/warnings.py’: Operation not permitted
chgrp: changing group of ‘/opt/app-root/lib64/python3.5/lib-dynload’: Operation not permitted
chgrp: changing group of ‘/opt/app-root/lib64/python3.5/plat-linux’: Operation not permitted
chmod: changing permissions of ‘/opt/app-root/include/python3.5m’: Operation not permitted
chmod: changing permissions of ‘/opt/app-root/lib/python3.5/copy.py’: Operation not permitted
chmod: changing permissions of ‘/opt/app-root/lib/python3.5/token.py’: Operation not permitted
chmod: changing permissions of ‘/opt/app-root/lib/python3.5/rlcompleter.py’: Operation not permitted
chmod: changing permissions of ‘/opt/app-root/lib/python3.5/collections’: Operation not permitted
chmod: changing permissions of ‘/opt/app-root/lib/python3.5/encodings’: Operation not permitted
```

I see that the issue isn't replicating itself, but i think it's still occuring?

Unless i'm not picking up the newest S2I build?

PR #203, opened 2017-07-21
prototype of new README.md file for s2i repositories

ficap:
according to new rules
@hhorak please review new stuff

### DIFF
--- a/3.5/README.md
+++ b/3.5/README.md
@@ -261,7 +261,6 @@ to `/opt/app-root/src`, where the source code is located.
 
 See also
 --------
-Base Python environment Dockerfile and other sources for building the final image 
-with your Python application are available on https://github.com/sclorg/s2i-python-container.
+Dockerfile and other sources are available on https://github.com/sclorg/s2i-python-container.
 In that repository you also can find another versions of Python environment Dockerfiles.
 Dockerfile for CentOS is called Dockerfile, Dockerfile for RHEL is called Dockerfile.rhel7.
--- a/3.5/README.md
+++ b/3.5/README.md
@@ -1,12 +1,13 @@
 Python 3.5 Docker image
 ===================
 
-This repository contains the source for building various versions of
-the Python application as a reproducible Docker image using
-[source-to-image](https://github.com/openshift/source-to-image).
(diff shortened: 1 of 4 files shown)


ficap:
and how about troubleshooting article. does it also logs to term so it appears in docker logs or developer of the application can have his/her own way to log the application for example logging to special file stored somewhere based on developer.

3.5/README.md
```diff
@@ -244,3 +255,11 @@ docker exec -it <CONTAINER_ID> /bin/bash
 
 After you enter into the running container, your current directory is set
 to `/opt/app-root/src`, where the source code is located.
+
+
+See also
+--------
+Base Python environment Dockerfile and other sources for building the final image 
+with your Python application are available on https://github.com/sclorg/s2i-python-container.
```
ficap: containing your Python application - 'd be better?

pkubatrh: Neither. You do not need to have the sources for s2i-python to be able to build an image containing your python application. Source-to-image does all the magic for you.
The plain "Dockerfile and other sources for this container image are available on" would be better.

torsava:
Looks good to me.

pkubatrh:
The description seems like it has been directly taken from Python's documentation. As I am not familiar with Python's license, are we allowed to do so without attribution?

In general I think it would be better to just use something we already have, like what is available in Fedora rpms:
https://admin.fedoraproject.org/pkgdb/package/rpms/python/

ficap:
I used description from Dockerfile DESC, so if there's any licence problem there is also.

pkubatrh:
>I used description from Dockerfile DESC, so if there's any licence problem there is also.

It might be just me over-thinking things in that case. Feel free to add changes to other versions as well.

ficap:
Should I squash these commits into one?

omron93:
@ficap Github allows squashing of PR when merging. But you can.

```
 Squash and merge
The 3 commits from this branch will be combined into one commit in the base branch. 
```

ficap:
@hhorak please review latest changes

hhorak:
@ficap the 3.5 version looks good to me, but the last changes are not done in 2.7 and 3.4 versions. The PR also will need to be rebased.

hhorak:
Otherwise no issues spotted.

hhorak:
This looks good to me, let's merge it and thanks!

PR #204, opened 2017-07-21
Process setup.py only if requirements.txt is not present

### LINKED ISSUE #125: Environment variable to disable setup.py processing.
There should be an environment variable to disable processing of `setup.py`.

Best practice for use of a `setup,py` as outlined by main developer who manages the Python package index in:
- https://caremad.io/2013/07/setup-vs-requirement/

is not to use `setup.py` directly, but to trigger its processing from the `requirements.txt` file using:

```
-e .
```

In case where people follow this practice, then processing of `setup.py` would be done twice.

One could actually argue that the decision to have `setup.py` be processed should always be opt in by using the `requirements.txt` file and so should not ever be processed, but too late to change it to the better use case. At the least therefore, there should be an environment variable to disable processing `setup.py` so doesn't cause problems by being run more than once, or even if the user never desired it to be used in the first place.

mcyprian:
If source contains requirements.txt file, direct processing of setup.py should be skipped. This will prevent processing it twice in case where people use -e . in requirements.txt.

- setup.py is processed directly only if requirements.txt is missing
- test case
- small note in README

I skipped version 3.3 that is deprecated according to its README. This PR intends to resolve #125.


pkubatrh:
[test]

torsava:
Looks good to me, merging. Thanks, Michal!

PR #208, opened 2017-08-20
Introduce version 3.6

hhorak:
This PR should hopefully keep the history in 3.6 directory, so blame works.. Plus there is also a fedora dockerfile added.


jfmatth:
Looks like your dockerfile.fedora references Python 3.5 not 3.6?

hhorak:
> Looks like your dockerfile.fedora references Python 3.5 not 3.6?

Thanks for spotting this, I've just fixed that in the comment, any other places where I did use 3.5?

jfmatth:
Not that I saw, just was glancing through and noticed that obvious one.

pkubatrh:
@hhorak From the PR it seems like you are expecting to build and test the new version on some of the platforms right at start. In that case please add the new version string to Makefile so that CI picks it up.

You can also pull in newest container-common-scripts to be able to blacklist specific version-platform combinations from being built (see https://github.com/sclorg/container-common-scripts/pull/20/commits/c31f4975b32abec9e88b065933044bdb9cde1655)

3.6/Dockerfile.rhel7.tbd
```diff
@@ -0,0 +1,74 @@
+# This image provides a Python 3.6 environment you can use to run your Python
```
torsava: As @pkubatrh mentioned, it would be better to [update container-common-scripts](https://github.com/sclorg/container-common-scripts/pull/20/commits/c31f4975b32abec9e88b065933044bdb9cde1655) and add a file `.exclude-rhel7` into the versioned directory instead of renaming the Dockerfile.

3.6/Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+
+LABEL summary="$SUMMARY" \
+      description="$DESCRIPTION" \
+      io.k8s.description="$DESCRIPTION" \
+      io.k8s.display-name="Python 3.6" \
+      io.openshift.expose-services="8080:http" \
+      io.openshift.tags="builder,python,python36,rh-python36" \
+      com.redhat.component="python36-docker" \
+      name="centos/python-36-centos7" \
+      version="3.6" \
```
torsava: Can we set the version to 1 in the CentOS specfile like it is in the RHEL one? I think we could, since it's a new image and its name already contains the Python version.

ikke-t:
Any reason this is pending?

pkubatrh:
@hhorak ^
Also need rebasing against master due to Dockerfile changes.

PR #209, opened 2017-08-24
Add NPM support and tests

torsava:
In light of the [recent addition of npm to s2i-base-container](https://github.com/sclorg/s2i-base-container/pull/132), I've enabled npm in this image and added some tests for it as well.

What do you think, @omron93, @pkubatrh, @hhorak ?

WIP because I'd like to wait for some feedback before committing to the other Python versions as well.


omron93:
@torsava I don't see any problem, thanks.

Only there are some problems with yum in rhel7.4, so please add this workaround (https://github.com/sclorg/mongodb-container/pull/253/commits/40322ebd4b61b7a1039a07581a9517eb6c406056) to let RHEL CI pass.

hhorak:
Sounds good to me. There is one more thing though -- the s2i base image with nodejs is not yet released, so the testing might fail even after fixing the 7.4 work-around. I'd probably go with it and merge it anyway, since waiting for next rebuild doesn't seem worth the PASS sign.

torsava:
I've propagated the changes to all active versions and added the rhel7.4 workaround, thanks @omron93 !

However, the RHEL build now fails with `error: option --user not recognized`, which seems like some problem with the build machinery. Would anyone know what that might be?

pkubatrh:
>However, the RHEL build now fails with error: option --user not recognized, which seems like some problem with the build machinery. Would anyone know what that might be?

This looks like some issue with the jenkins builder (my PRs are hitting this all across the board) so @omron93 might know more.

pkubatrh:
@torsava Likely a side-effect of https://github.com/sclorg/container-common-scripts/commit/55d1108ed56058756665e5fe893c1e07f73bdce7. Rebasing to latest `common` seems to have fixed the issue in my PRs.

torsava:
Thank you @pkubatrh, it indeed helped. Now the RHEL test fails due to rh-nodejs6 not being available in the s2i-base-container as @hhorak predicted.

torsava:
Rebased onto the master so there are no conflicts.

hhorak:
[test]

hhorak:
@torsava There are new conflicts now, please, rebase.

torsava:
@hhorak Rebased.

PR #211, opened 2017-08-28
Dockerfile changes

pkubatrh:


3.4/Dockerfile.rhel7
```diff
@@ -35,9 +35,10 @@ LABEL summary="$SUMMARY" \
 # To use subscription inside container yum command has to be run first (before yum-config-manager)
 # https://access.redhat.com/solutions/1443553
 RUN yum repolist > /dev/null && \
+    yum-config-manager --disable \* &> /dev/null && \
+    yum-config-manager --enable rhel-7-server-rpms && \
     yum-config-manager --enable rhel-server-rhscl-7-rpms && \
```
torsava: In the version 3.4 the ordering of the repositories is difeerent than in 2.7 and 3.5. It might be worth fixing that so the files are as similar as possible.

3.4/Dockerfile.rhel7
```diff
@@ -32,13 +32,14 @@ LABEL summary="$SUMMARY" \
...
 
-# To use subscription inside container yum command has to be run first (before yum-config-manager)
-# https://access.redhat.com/solutions/1443553
+# We need to call 2 (!) yum commands before being able to enable repositories properly
+# This is a workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1479388
 RUN yum repolist > /dev/null && \
+    yum install -y yum-utils && \
+    yum-config-manager --disable \* &> /dev/null && \
+    yum-config-manager --enable rhel-7-server-rpms && \
     yum-config-manager --enable rhel-server-rhscl-7-rpms && \
```
torsava: In version 3.4 the ordering of the enabled repositories is different than in 2.7 and 3.5. It might be worth fixing that so the files are as similar as possible.

torsava:
Besides that one note I'm ready to merge this.

pkubatrh:
Thanks for review, should be fixed now.

torsava:
Merged, thanks!

Issue #201, opened 2017-07-11
UPGRADE_PIP_TO_LATEST should also perhaps update wheel package as well.


GrahamDumpleton:
``UPGRADE_PIP_TO_LATEST`` environment variable when set will update to latest version of ``pip``, but it should also perhaps update to latest version of ``wheel`` at the same time. If you don't you can hit problems when doing stuff with wheels still, such as:

* https://bitbucket.org/pypa/wheel/issues/146/wheel-building-fails-on-cpython-350b3

In short, these packages should really be upgraded together to ensure newest packaging tools are being used.

GrahamDumpleton:
Also should pull in latest ``setuptools`` to be safe.

torsava:
Agreed, pip, setuptools and wheel are highly interconnected and we should update all three.

PR #213, opened 2017-09-01
Upgrade also wheel and setuptools when UPGRADE_PIP_TO_LATEST is set

mcyprian:
Pip setuptools and wheel are highly interconnected and should be upgraded together as it is described in issue #201.


pkubatrh:
[test]

torsava:
Looks good to me, merging, thanks!

torsava:
Looks good to me! However, please amend the relevant README section as well, before we merge this.

PR #222, opened 2017-10-24
Sync downstream changes

pkubatrh:


omron93:
What about adding `rpm-file-permissions` into last `RUN` statement?

pkubatrh:
>What about adding rpm-file-permissions into last RUN statement?

Makes sense, must have missed it while adding downstream. Thanks!

pkubatrh:
Modified the RUN command so that it does not introduce another layer.

omron93:
Tests are passing. So merging.

Issue #207, opened 2017-08-09
add enchant package to the builder image


akostadinov:
Description of problem:
Trying to create a python app using sti build but library is missing.

Version-Release number of selected component (if applicable):
2.7 and 3.5

How reproducible:
always

Steps to Reproduce:
1. oc new-app python:2.7~https://github.com/sopel-irc/sopel.git
2. oc new-app python:3.5~https://github.com/sopel-irc/sopel.git
3. oc new-app python:latest~https://github.com/sopel-irc/sopel.git

4. oc logs -f bc/sopel

Actual results:

  Sti build fails like:
```
> ---> Installing dependencies ...
> You are using pip version 7.1.0, however version 9.0.1 is available.
> You should consider upgrading via the 'pip install --upgrade pip' command.
> Collecting xmltodict (from -r requirements.txt (line 1))
> Downloading xmltodict-0.11.0-py2.py3-none-any.whl
> Collecting pytz (from -r requirements.txt (line 2))
> Downloading pytz-2017.2-py2.py3-none-any.whl (484kB)
> Collecting praw (from -r requirements.txt (line 3))
> Downloading praw-5.0.1-py2.py3-none-any.whl (86kB)
> Collecting pyenchant (from -r requirements.txt (line 4))
> Downloading pyenchant-1.6.8.tar.gz (63kB)
> Complete output from command python setup.py egg_info:
> Traceback (most recent call last):
> File "<string>", line 20, in <module>
> File "/tmp/pip-build-qSAnGK/pyenchant/setup.py", line 210, in <module>
> import enchant
> File "enchant/__init__.py", line 92, in <module>
> from enchant import _enchant as _e
> File "enchant/_enchant.py", line 145, in <module>
> raise ImportError(msg)
> ImportError: The 'enchant' C library was not found. Please install it via your > OS package manager, or use a pre-built binary wheel from PyPI.
> ----------------------------------------
> Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-qSAnGK/pyenchant
```
Expected results:
build succeeds

Additional info:

Please add `enchant` package to the python builder image. Initially I requested a mechanism to add random dependencies but this feature has been rejected by upstream s2i project [1]. That's why I'm requesting a specific fix for people to be able to deploy the sopel bot project.

[1] openshift/source-to-image/issues/786

torsava:
[...]

PR #224, opened 2017-11-08
Install the enchant rpm package into the image

torsava:
It is used for spell-checking and was requested here:
https://github.com/sclorg/s2i-python-container/issues/207

The size of the image increased by less than 1 MB, so less than 1%.

### DIFF
--- a/2.7/Dockerfile
+++ b/2.7/Dockerfile
@@ -36,7 +36,7 @@ LABEL summary="$SUMMARY" \
 RUN yum install -y centos-release-scl && \
     INSTALL_PKGS="libjpeg-turbo libjpeg-turbo-devel python27 python27-python-devel python27-python-setuptools \ 
 	python27-python-pip nss_wrapper httpd24 httpd24-httpd-devel httpd24-mod_ssl httpd24-mod_auth_kerb \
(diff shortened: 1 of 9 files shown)


3.3/Dockerfile
```diff
@@ -22,7 +22,7 @@ LABEL io.k8s.description="Platform for building and running Python 3.3 applicati
 RUN yum install -y centos-release-scl && \
     INSTALL_PKGS="python33 python33-python-devel python33-python-setuptools nss_wrapper httpd24 \
         httpd24-httpd-devel httpd24-mod_ssl httpd24-mod_auth_kerb httpd24-mod_ldap httpd24-mod_session \
-	atlas-devel gcc-gfortran libffi-devel libtool-ltdl" && \
+	atlas-devel gcc-gfortran libffi-devel libtool-ltdl enchant" && \
```
mcyprian: The Python 3.3 image is deprecated, is there any specific reason to apply this change?

torsava: My apologies, I did `vi */Dockerfile*` and inadvertently fixed 3.3 as well, thanks!

torsava:
Thanks for review, @mcyprian! I've amended the commit to not change 3.3.

torsava:
Tests are passing now, merging.

PR #225, opened 2017-11-10
Limit number of gunicorn workers to 12 at maximum. Fixes #199

### LINKED ISSUE #199: Automatic determination of gunicorn processes when no resource limits.
Saw this behaviour sometime back but possibly forgot to raise it. I can't revalidate it right now, but should be investigated to see if still an issue.

The issue is that if there are no resource limits set on a deployment, nor enforced on it by being set on a project and inherited, then it appears the whole amount of resources available to the node is used in the calculation of the number of process to run gunicorn with.

The consequence of this is that if the node had 100+ GB of memory, then a ridiculous number of gunicorn processes get allocated, way more than is reasonable. If there are no limits set on the deployment configuration, the amount of processes gunicorn is configured to use should be capped at some reasonable level.

This almost caused issues when using a single node cluster with lots of memory in some training, as the gunicorn example unexpectedly consumed huge amounts of memory. Luckily not everyone doing the training did the deployment because if they did, it would have overwhelmed the node.

bkabrda:
As I understand it, the `3.3` image is not maintained any more, so I didn't fix values for it.

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -117,7 +117,8 @@ a `.s2i/environment` file inside your source code repository.
 
     Set this to change the default setting for the number of
     [workers](http://docs.gunicorn.org/en/stable/settings.html#workers). By
-    default, this is set to the number of available cores times 4.
+    default, this is set to the number of available cores times 2, capped
+    at 12.
 
 Source repository layout
 ------------------------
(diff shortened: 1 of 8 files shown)


pkubatrh:
[test]

pkubatrh:
@omron93 Any idea why the rhel7 CI fails?

omron93:
[test]

@pkubatrh Server for allocating RHEL VMs was down (no sign of failure). It should work now.

pkubatrh:
Thanks! Tests are passing now. LGTM, merging.

PR #228, opened 2017-11-15
Change the working directory in run script to APP_HOME if set

mcyprian:
#170 was not updated for quite some time, so I went through originally proposed changes and the discussion. I tried to make up something that resolves problem described in #166, doesn't break any of checks implemented in run script and simplify logic of the script. I would like to hear your opinion on this @GrahamDumpleton @pkubatrh.

### DIFF
--- a/2.7/s2i/bin/run
+++ b/2.7/s2i/bin/run
@@ -35,6 +35,13 @@ function get_default_web_concurrency() {
   echo $default
 }
 
+APP_HOME=$(readlink -f ${APP_HOME:-.})
+# Change the working directory to APP_HOME if set
+if [[ "$APP_HOME" != "." ]]; then
+  PYTHONPATH=$(pwd):$PYTHONPATH
+  cd "$APP_HOME"
+fi
+
 app_script_check="${APP_SCRIPT-}"
 APP_SCRIPT="${APP_SCRIPT-app.sh}"
 if [[ -f "$APP_SCRIPT" ]]; then
@@ -56,9 +63,8 @@ else
   test -n "$app_file_check" && (>&2 echo "ERROR: file '$app_file_check' not found.") && exit 1
 fi
 
-APP_HOME=${APP_HOME:-.}
-# Look for 'manage.py' in the directory specified by APP_HOME, or the current direcotry
-manage_file=$APP_HOME/manage.py
+# Look for 'manage.py' in the current directory
+manage_file=./manage.py
 
 if should_migrate; then
   if [[ -f "$manage_file" ]]; then
@@ -71,14 +77,12 @@ if should_migrate; then
 fi
 
 if is_gunicorn_installed; then
-  if [[ -z "$APP_MODULE" ]]; then
-    # Look only in the directory specified by APP_HOME, or the current directory
-    # replace all "/" with ".", remove leading "." and ".py" suffix
-    APP_MODULE=$(find $APP_HOME -maxdepth 1 -type f -name 'wsgi.py' | sed 's:/:.:g;s:^\.\+::;s:\.py$::')
-  fi
-
-  if [[ -z "$APP_MODULE" && -f setup.py ]]; then
-    APP_MODULE="$(python setup.py --name)"
+  setup_py=$(find $HOME -maxdepth 2 -type f -name 'setup.py' -print -quit)
+  # Look for wsgi module in the current directory
+  if [[ -z "$APP_MODULE" && -f "./wsgi.py" ]]; then
+    APP_MODULE=wsgi
+  elif [[ -z "$APP_MODULE" && -f $setup_py ]]; then
+    APP_MODULE="$(python $setup_py --name)"
   fi
 
   if [[ "$APP_MODULE" ]]; then
--- a/2.7/s2i/bin/run
+++ b/2.7/s2i/bin/run
@@ -35,7 +35,7 @@ function get_default_web_concurrency() {
   echo $default
 }
 
-APP_HOME=$(readlink -f ${APP_HOME:-.})
+APP_HOME=$(readlink -f "${APP_HOME:-.}")
 # Change the working directory to APP_HOME if set
 if [[ "$APP_HOME" != "." ]]; then
   PYTHONPATH=$(pwd):$PYTHONPATH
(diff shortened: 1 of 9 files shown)


mcyprian:
[test]

omron93:
[test]

mcyprian:
Rebased against master branch after merge of #234 .

pkubatrh:
[test]

2.7/s2i/bin/run
```diff
@@ -35,6 +35,13 @@ function get_default_web_concurrency() {
   echo $default
 }
 
+APP_HOME=$(readlink -f ${APP_HOME:-.})
```
pkubatrh: Double quotes needed around APP_HOME

GrahamDumpleton:
It will be after next week before I can look at this properly. I will be finally released then from major thing I had to get done and will have free time again. :-)

mcyprian:
I  rebased this PR to use templates after merge of https://github.com/sclorg/s2i-python-container/pull/259. I am not sure if this change is necessary, but we should either finish the work here and merge it or close this PR and https://github.com/sclorg/s2i-python-container/pull/170 as well. What do you think @pkubatrh?

mcyprian:
[test-openshift]

src/s2i/bin/run
```diff
@@ -35,6 +35,13 @@ function get_default_web_concurrency() {
   echo $default
 }
 
+APP_HOME=$(readlink -f ${APP_HOME:-.})
```
pkubatrh: `APP_HOME` in the readlink call needs to be double-quoted to avoid word splitting

src/s2i/bin/run
```diff
@@ -71,14 +77,12 @@ if should_migrate; then
...
-  fi
-
-  if [[ -z "$APP_MODULE" && -f setup.py ]]; then
-    APP_MODULE="$(python setup.py --name)"
+  setup_py=$(find $HOME -maxdepth 2 -type f -name 'setup.py' -print -quit)
+  # Look for wsgi module in the current directory
+  if [[ -z "$APP_MODULE" && -f "./wsgi.py" ]]; then
+    APP_MODULE=wsgi
+  elif [[ -z "$APP_MODULE" && -f $setup_py ]]; then
+    APP_MODULE="$(python $setup_py --name)"
```
pkubatrh: Same for `$setup_py` and `HOME` in this block.

pkubatrh:
Otherwise lgtm, I think its worth it to get this merged in.

GrahamDumpleton:
Can you wait until I have had another look before merging. I recollect that originally this change wasn't even doing things in the way that I had originally asked for and felt was required. I am on holidays right now, but will try and look later today or tomorrow.

2.7/s2i/bin/run
mcyprian: When trying this APP_HOME=$(readlink -f ${"APP_HOME":-.}) it ends up with an error bash: ${"APP_HOME":-.}: bad substitution, maybe I don't understand your point here correctly

omron93: I guess `APP_HOME=$(readlink -f "${APP_HOME:-.}")`

GrahamDumpleton:
For ``gunicorn`` case, does ``examples/app-home-test-app/project/__init__.py`` need to exist in the example for this test? The aim is that things should work without it if using method where ``cd`` into ``APP_HOME`` directory. Unless can explain why it needs to be there, should remove it to ensure that latest changes are not dependent on it existing, which wasn't what was desired.

For Django case, are there any possible risks with ``python manage.py collectstatic`` being run from top level directory with path to ``manage.py``, rather than ``cd`` to ``APP_HOME`` to do it? When do ``python manage.py runserver``, a ``cd`` is done. Am concerned that someone has written something to depend on the current working directory, with assumption that ``manage.py`` is always run from the directory where it is located. If use path to ``manage.py``, the current working directory will be ``/app/app-root/src`` still and not ``APP_HOME``.

mcyprian:
Thanks for your comments @GrahamDumpleton. The `examples/app-home-test-app/project/__init__.py` doesn't need to exist, I removed it in 046feced85c083c03fd4b90.

I am not aware of any issues caused by `python manage.py collectstatic
` being run from different location, but I added cd $APP_HOME (50ccd6d3cace383efa609a45f) to the run script as well to keep the two scripts aligned.

pkubatrh:
Rebased onto current master for latest bits

[test-openshift]

pkubatrh:
Removed the check for `.` since afaik `readlink -f` will never return `.`, otherwise lgtm.

@GrahamDumpleton @torsava Would either one of you guys have some spare time to review as well?

omron93:
[test-openshift]

2.7/s2i/bin/run
```diff
@@ -71,14 +75,12 @@ if should_migrate; then
...
 if is_gunicorn_installed; then
-  if [[ -z "$APP_MODULE" ]]; then
-    # Look only in the directory specified by APP_HOME, or the current directory
-    # replace all "/" with ".", remove leading "." and ".py" suffix
-    APP_MODULE=$(find $APP_HOME -maxdepth 1 -type f -name 'wsgi.py' | sed 's:/:.:g;s:^\.\+::;s:\.py$::')
-  fi
-
-  if [[ -z "$APP_MODULE" && -f setup.py ]]; then
-    APP_MODULE="$(python setup.py --name)"
+  setup_py=$(find "$HOME" -maxdepth 2 -type f -name 'setup.py' -print -quit)
```
torsava: Why `$HOME` instead of `$APP_HOME`?

torsava: Right, sources are copied to the $HOME directory so setup.py should be there. The `-maxdepth 2` is just for safety I presume?

2.7/s2i/bin/run
```diff
@@ -35,6 +35,11 @@ function get_default_web_concurrency() {
   echo $default
 }
 
+APP_HOME=$(readlink -f "${APP_HOME:-.}")
+# Change the working directory to APP_HOME
+PYTHONPATH=$(pwd):$PYTHONPATH
```
torsava: Unless we're sure that `PYTHONPATH` is set and non-empty at this point, this might introduce a trailing colon, which is not safe with PYTHONPATH as far as I know.

torsava:
I've pointed some things in the comments. Besides that it looks good!

pkubatrh:
New changes pushed

[test-openshift]

3.4/s2i/bin/run
```diff
@@ -35,6 +35,13 @@ function get_default_web_concurrency() {
   echo $default
 }
 
+APP_HOME=$(readlink -f "${APP_HOME:-.}")
+# Change the working directory to APP_HOME if set
+if [[ "$APP_HOME" != "." ]]; then
+  PYTHONPATH=$(pwd):$PYTHONPATH
+  cd "$APP_HOME"
+fi
```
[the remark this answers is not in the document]
omron93: @torsava Or what about removing generated changes from `master` and for people that need the code use automatically generated `generated` branch ?

2.7/s2i/bin/run
torsava: I realize it's just a detail, but for readability's sake:
Since we're using $HOME in this place, could we also use it when we move the scripts there?

https://github.com/sclorg/s2i-python-container/blob/0b8fa29097d5e179dde6e2267c44f2e1d5f6fa5b/2.7/s2i/bin/assemble#L38

Right now it says `mv /tmp/src/* ./`, which is quite cryptic unless you know where you're currently located, using `$HOME/` would make it much clearer for anyone trying to understand it.

pkubatrh: Makes sense to me. Changed in the latest commit.

pkubatrh:
[test-openshift]

PR #230, opened 2017-11-20
Add note on running DB migrations. Fixes #95

### LINKED ISSUE #95: Django manage.py migrate will cause corruption to databases.
Having Django database migration enabled to be run by default inside of the container when the web application is run is not best practice and could be dangerous. Yes, there is an option to optionally disable it, but this isn't of much use when an unknowing user has already corrupted their database when they scaled up to multiple replicas.

Right now with the current Django version the risk of a corrupted database many be minimal, but with the next version of Django this could all change if what I have heard is correct.

In the current version of Django if deploying with multiple replicas, then there would be a race condition on the determination of whether database migration steps actually need to be run. The database migration itself is still done within a database transaction though, so when migration steps do end up running in multiple instances, then the transaction lock should mean that in the second, it would when it subsequently gets its transaction and checks again, realise the steps have already run and skip them.

The problem now is that I have heard that a future Django version will have transaction-less database migration. If I understand what that implies, that means that there will not be a transaction to protect an attempt to do a migration from multiple replicas. This means that multiple attempts at the same time will interfere with each other and potentially corrupt the database.

In short, running database migration as a deploy step is a very bad idea and not something we want to encourage or even allow. Yes, this does mean extra separate steps will be required but that is better than us corrupting a users databases. So the solution is to disable automatic migration as a default and provide documentation on what is best practice for Python as to how to manage database migration in the context of OpenShift.

bkabrda:
Hi, I think this would be the best solution to #95, since there seems to be no "one size fits all" solution and the linked blogpost seems to describe two good ways of achieving DB migrations in OpenShift. If this seems ok, I'll also amend the other README files (didn't do it for now as I want to hear comments first).

CC @GrahamDumpleton @torsava @hhorak

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -91,15 +91,19 @@ a `.s2i/environment` file inside your source code repository.
     Path to a valid Python file with a
     [Gunicorn configuration](http://docs.gunicorn.org/en/latest/configure.html#configuration-file) file.
 
+* **DISABLE_MIGRATE**
+
+    Set this variable to a non-empty value to inhibit the execution of 'manage.py migrate'
+    when the produced image is run. This only affects Django projects. See
+    "Handling Database Migrations" section of [Django blogpost on OpenShift blog](
+    https://blog.openshift.com/migrating-django-applications-openshift-3/) on suggestions
+    how/when to run DB migrations in OpenShift environment. Most importantly,
+    note that running DB migrations from two or more pods might corrupt your database.
+
 * **DISABLE_COLLECTSTATIC**
 
     Set this variable to a non-empty value to inhibit the execution of
     'manage.py collectstatic' during the build. This only affects Django projects.
-
-* **DISABLE_MIGRATE**
-
-    Set this variable to a non-empty value to inhibit the execution of 'manage.py migrate'
-    when the produced image is run. This only affects Django projects.
 
 * **PIP_INDEX_URL**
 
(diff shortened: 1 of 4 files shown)


mcyprian:
[test]

torsava:
@bkabrda So in your opinion, it's best to keep the default setting to migrate?

bkabrda:
@torsava I'm not 100 % sure about that, but I tend to say it makes sense to keep this as it is. The reason is that when starting to develop your application initially, you'll most likely want to start with a simple setup which achieves working deployment quickly. You're then going to want to improve on this and the migrations are going to want to be a part of that, but I think initially this makes sense. YMMV ;)

torsava:
So, I've reread the long issue discussion, and I agree with your proposal. I do have 2 suggestions:

1. I would mention outright that automatic migration done by more than one pod at the same time might corrupt your database.
2. I would move the `DISABLE_MIGRATE` paragraph above the `DISABLE_COLLECTSTATIC` so it's in a more prominent place.

If you're running out of time in your sprint, @bkabrda, I can do the changes.

bkabrda:
Thanks for the review! I agree with your suggestions and I'll definitely do them myself, but probably will get to it tomorrow, as I'm buried deep in another issue and don't want to get myself distracted ATM.

bkabrda:
@torsava I amended the commit and it is now ready for re-review and merge.

torsava:
@bkabrda LGTM, thank you!

PR #231, opened 2017-11-20
Add .travis.yml file

mcyprian:
I set up Travis CI builds to run test suit for CentOS images after each commit in my fork. I think It would be nice to have .travis.yml file in the sources, even in case Travis won't be enabled for this repository. Some other contributors might find the possibility of alternative CI service helpful.


torsava:
I like the idea! However, could you please put a comment in there that it's not being used in this repo, but it's for people to use at their will?

mcyprian:
I added a short description to top of the file.

torsava:
Thanks!

PR #232, opened 2017-11-21
Process setup.py script by default

mcyprian:
At this moment we don't process setup.py by default in case `requirements.txt/Pipfile` is present and try to encourage users to use `-e .` in `requirements.txt` to trigger its processing.
This PR changes the default behaviour to following:
- `requirements.txt/Pipfile` is processed first if present
- `setup.py is processed` (`pip install .` command)

Processing of `setup.py` script can be disabled via `DISABLE_SETUP_PY_PROCESSING` env variable, which was also an original proposal of https://github.com/sclorg/s2i-python-container/issues/125. In case user doesn't disable setup.py processing and it contains the same dependency that was listed in `requirements.txt/Pipfile` the result is "Requirement already satisfied: ...".


pkubatrh:
[test]

mcyprian:
Rebased against master branch after merge of #234.

mcyprian:
I  rebased this PR to use templates after merge of https://github.com/sclorg/s2i-python-container/pull/259. It has been opened for a while, these changes are needed. In my opinion we should at least replace `python setup.py develop` with `pip install .` to be sure packages versions are picked properly. Can you please take a look at this @torsava?

mcyprian:
[test-openshift]

src/README.md
```diff
@@ -127,13 +127,20 @@ file inside your source code repository.
     Pipenv.){% endif %}
 
 
+* **DISABLE_SETUP_PY_PROCESSING**
```
torsava: AFAIK the list of envvars used to be alphabetized and still is besides this and ENABLE_PIPENV. Could you please put these two in the alphabetical order?

src/s2i/bin/assemble
```diff
@@ -56,9 +56,11 @@ if [[ ! -z "$ENABLE_PIPENV" ]]; then
 elif [[ -f requirements.txt ]]; then
   echo "---> Installing dependencies ..."
   pip install -r requirements.txt
-elif [[ -f setup.py ]]; then
+fi
+
+if [[ -f setup.py && -z "$DISABLE_SETUP_PY_PROCESSING" ]]; then
   echo "---> Installing application ..."
-  python setup.py develop
+  pip install .
```
torsava: Why this change?

torsava:
So if Pipfile or requirements.txt contains `-e .` then setup.py is run. And then you do `pip install .` and it automatically detects that setup.py has already been run before, is that correct?

torsava:
I'm a little worried that this will change behaviour for existing users that did not want to run setup.py, do you foresee any breakage due to running setup.py unexpectadly?

mcyprian:
That's right, if someone uses `-e . ` the environment variable should be used to disable processing of setup.py as it is described in README. In case this variable is not set the result is a bunch of warnings "Requirement already satisfied: foo in /opt/app-root/lib/python3.6/site-packages".

mcyprian:
[test-openshift]

mcyprian:
[test]

mcyprian:
All the CI checks are green, changes were approved by @torsava in personal conversation. Merging.

mcyprian:
[test]

PR #233, opened 2017-11-23
Implement init wrapper for reaping zombie processes. Fixes #94

### LINKED ISSUE #94: No zombie process reaping protection built in.
The builder image doesn't provide its own mechanism for protecting against process ID '1' zombie reaping problems.

The issue is outlined in:
- http://blog.dscpl.com.au/2015/12/issues-with-running-as-pid-1-in-docker.html

The result is that the requirement for reaping zombie processes is put onto the user, but many WSGI servers, ASYNC web servers or other Python web servers do not meet the requirements of what an application running as process ID '1' needs to do. This could result in exhaustion of kernel resources in the worst case as the process table fills up with dead entries.

The rather abysmal score card for Python WSGI, ASYNC and web servers is:
- django (runserver) - FAIL
- flask (builtin) - FAIL
- gunicorn - PASS
- Apache/mod_wsgi - PASS
- tornado (async) - FAIL
- tornado (wsgi) - FAIL
- uWSGI - FAIL
- uWSGI (master) - PASS
- waitress - FAIL
- wsgiref - FAIL

So it isn't an isolated issue.

The problem now is that because of the requirement for any third party packages to be packaged in SCL, you can't use a tool such as `tini`.
- https://github.com/krallin/tini

or even the `pip` installable `dumb-init`:
- https://pypi.python.org/pypi/dumb-init/1.0.0
- https://github.com/Yelp/dumb-init

The only options would be to see if one can at least get away with avoiding some of the issues by using the shell script solution in:
- http://blog.dscpl.com.au/2015/12/issues-with-running-as-pid-1-in-docker.html

of:

```
#!/bin/sh
trap 'kill -TERM $PID' TERM INT
python server_wsgiref.py &
PID=$!
wait $PID
trap - TERM INT
wait $PID
STATUS=$?
exit $STATUS
```

This likely isn't going to be entirely reliable, but, still better than nothing.

Or, instead learn from the Python script mini init originally described by Phusion and implement our own to suit our requirements, add it to the repo and bundle it with the builder.
- https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/
- https://github.com/phusion/baseimage-docker/blob/rel-0.9.16/image/bin/my_init

Note that a proper solution would also resolve other related problems with running as process ID '1' described in #77. Note that only reason accepted that #77 was closed was because it would be impossible to fix the default builder to do this properly and so that anyone would always ignore the default builder for everything anyway. If you actually want to people to use the default builder, this issue does need to be addressed and can't be ignored.

bkabrda:


bkabrda:
I'm not 100 % sure, but it seems to me that the "rhel7" build failure is not related to this PR and is a random Jenkins setup issue.

pkubatrh:
>I'm not 100 % sure, but it seems to me that the "rhel7" build failure is not related to this PR and is a random Jenkins setup issue.

That seems to be the case.

@omron93 @praiskup Can you take a look?

praiskup:
[test]

praiskup:
I started the resalloc daemon, for some reason it was inactive.

praiskup:
[test] again (script timeouted..?)

2.7/s2i/bin/run
```diff
@@ -35,14 +35,34 @@ function get_default_web_concurrency() {
   echo $default
 }
 
+function maybe_run_in_init_wrapper() {
+  if [[ -z "$ENABLE_INIT_WRAPPER" ]]; then
+    exec $@
+  else
+    local wrapper=`mktemp -p .`
```
omron93: I would prefer to remove wrapper file after tests or create it under `/tmp/`. Any reason why to create it in tests directory?

bkabrda: I'm not sure how this relates to tests... You create the wrapper when starting the container and the only point when removing the wrapper would make sense is when the container is shutting down, at which point you don't care that there's a leftover file, do you?

omron93: Oh, my fault. Sorry, you are right.

omron93:
@pkubatrh If you are fine with this, please merge.

pkubatrh:
@bkabrda 
I see you run the option enabled against `test-application` but it would be better to have a test that looks at whether or not the zombie processes are actually reaped. Could such a test case be added?

2.7/s2i/bin/run
```diff
@@ -35,14 +35,34 @@ function get_default_web_concurrency() {
   echo $default
 }
 
+function maybe_run_in_init_wrapper() {
+  if [[ -z "$ENABLE_INIT_WRAPPER" ]]; then
+    exec $@
```
praiskup: Please double-quote.

3.4/s2i/bin/run
```diff
@@ -35,14 +35,34 @@ function get_default_web_concurrency() {
...
 
+function maybe_run_in_init_wrapper() {
+  if [[ -z "$ENABLE_INIT_WRAPPER" ]]; then
+    exec $@
+  else
+    local wrapper=`mktemp -p .`
+    chmod u+x $wrapper
+    cat <<EOF > $wrapper
+trap 'kill -TERM \$PID' TERM INT
+$@
```
praiskup: Missing double quotes too.

This is rather a preference, but some precisely shaped init wrapper could live directly in e.g. shareable '/bin/s2i-init-wrapper' (could live in s2i-core).  Basically, at least I would prefer not to have temporary file for that; alternative could be `/bin/sh -c 'kill ... ; "$@" ; ...' -- "$@"`

praiskup: ... and the `$ENABLE_INIT_WRAPPER` (or something else) could have the wrapper inside.

bkabrda:
Thanks all for the comments. I currently have quite a lot of work, but I'll try to address them during next week.

pkubatrh:
@bkabrda ping

frenzymadness:
Hello @bkabrda. Do you have a plan to finish this or somebody should take it over?

torsava:
I believe @bkabrda won't have time to work on this anymore. We have just discussed with @frenzymadness that we'd try to address the issues with this PR and then deploy it only on the Python 3.7 version. It'll hopefully get tested on Fedora before being deployed on RHEL/CentOS. And we won't be breaking existing usage for users.

frenzymadness:
Closing this one in favour of https://github.com/sclorg/s2i-python-container/pull/322

PR #234, opened 2017-11-27
Tests deduplication

mcyprian:
Sometimes I find it a bit annoying to copy the same code into four different version directories. I believe we can reduce amount of such work. The nice first step could be deduplication of test applications. Majority of containers  maintained by sclorg organization have already done something similar.

After playing with it for a while, I've made up an implementation matching following objectives:
- What varies can be specified separately for each version (`pipenv-test-app` at this moment)
- What is the same can be symlinked from the shared space (not sure if `s2i-python-container/examples` or `s2i-python-container/tests` is the most appropriate location for this)
- `run` script can be also shared after small adjustment

This approach is inspired by [tests of nginx-container](https://github.com/sclorg/nginx-container/tree/master/1.12/test).


pkubatrh:
Tried to hook this up to some of my tooling:

```
OSError: [Errno 40] Too many levels of symbolic links: 'python/2.7/test/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app/django-test-app'

**Comment by omron93** [comment]
LGTM, thanks.

@pkubatrh Please take a look if this matches all needs for downstream sync process.

[test]

File: test/run [hunk 1]
```diff
@@ -0,0 +1,205 @@
...
+# TODO: Make command compatible for Mac users
+test_dir="$(readlink -zf $(dirname "${BASH_SOURCE[0]}"))"
+image_dir=$(readlink -zf ${test_dir}/..)
+
+if [[ -z $VERSION ]]; then
+    echo "ERROR: The VERSION variable must be set."
+    exit 1
+fi
+
+IMAGE_NAME=${IMAGE_NAME:-sclorg/python-${VERSION//./}-centos7-candidate}
```
pkubatrh: Would rather go with this as default:

```
IMAGE_NAME=${IMAGE_NAME:-centos/python-${VERSION//./}-centos7}
```

as I do not think candidate tags are used anymore and centos images live in centos namespace.

pkubatrh:
After removing the problematic symlinks in my local copy the tooling works ok and tests complete without any issues even in an isolated environment so no further changes other than those mentioned should be needed.

examples/app-home-test-app/app-home-test-app
```diff
@@ -0,0 +1 @@
+../../examples/app-home-test-app
```
pkubatrh: Still present

examples/django-test-app/django-test-app
```diff
@@ -0,0 +1 @@
+../../examples/django-test-app
```
pkubatrh: ditto

mcyprian:
I haven't realised presence of these extra symlinks before, it should be fixed now, thanks @pkubatrh.

test/run
```diff
@@ -17,7 +17,7 @@ if [[ -z $VERSION ]]; then
     exit 1
 fi
 
-IMAGE_NAME=${IMAGE_NAME:-sclorg/python-${VERSION//./}-centos7-candidate}
+IMAGE_NAME=${IMAGE_NAME:-sclorg/python-${VERSION//./}-centos7}
```
pkubatrh: Still in the `sclorg` namespace instead of `centos`

pkubatrh:
[test]

pkubatrh:
Thanks! lgtm now

PR #245, opened 2018-02-08
Use -P option of find command in fix-permissions script

mcyprian:
Stop following symlinks when running fix-permissions script under root to prevent undesirable permission changes of software collection files. Resolves https://bugzilla.redhat.com/show_bug.cgi?id=1536129.


omron93:
LGTM, thanks.

Please update also Fedora Dockerfile... (fedora s2i-core container should container latest changes from upstream sclorg/)

omron93:
[test][test-openshift]

PR #248, opened 2018-02-26
add imagestreams to image repo

bparees:


bparees:
@praiskup @hhorak @pkubatrh @pvalena ptal

pkubatrh:
@bparees Is it necessary to keep the deprecated images in the imagestream as well?

pkubatrh:
Will merge for now since it does not matter much anyway but it would be good to know whether we can remove them in future without causing harm.

bparees:
> @bparees Is it necessary to keep the deprecated images in the imagestream as well?

yes for backwards compatibility we need to keep deprecated images in the imagestream, but we tag them as "hidden"

PR #249, opened 2018-02-27
Remove Python 3.3

mhutter:
As per the README, only 2.7, 3.4, 3.5 and 3.6 are supported so let's get rid of 3.3.


pkubatrh:
Thanks for the PR, unfortunately I will have to close this.

We are keeping the deprecated versions in git for archival purposes for now but are looking into more ways of being able to keep the archives available while also having the repository uncluttered.

PR #250, opened 2018-02-27
 Bring s2i scripts in line between versions

mhutter:
It's hard to introduce changes to the s2i scripts due to differences between the individual versions. This change aims to reduce the differences and increase maintainability.

- add `make sync` to copy `s2i/bin/{run,assemble}` from 3.6 to the other versions. Can of course be improved to copy mor files and be more flexible
- update `s2i/bin/{run,assemble}` to be identical for all supported versions

On long term it may make sense to have some kind of "template", similar to other repositories and helper scripts to generate the files for each version.

### DIFF
--- a/2.7/s2i/bin/assemble
+++ b/2.7/s2i/bin/assemble
@@ -15,6 +15,7 @@ function should_collectstatic() {
 # (pip script installer).
 function install_pipenv() {
   echo "---> Installing pipenv packaging tool ..."
+  pip install -U virtualenv
   VENV_DIR=$HOME/.local/venvs/pipenv
   virtualenv $VENV_DIR
   $VENV_DIR/bin/pip --isolated install -U pipenv
--- a/3.6/s2i/bin/assemble
+++ b/3.6/s2i/bin/assemble
@@ -15,6 +15,7 @@ function should_collectstatic() {
 # (pip script installer).
 function install_pipenv() {
   echo "---> Installing pipenv packaging tool ..."
+  pip install -U virtualenv
   VENV_DIR=$HOME/.local/venvs/pipenv
   virtualenv $VENV_DIR
   $VENV_DIR/bin/pip --isolated install -U pipenv
@@ -28,7 +29,7 @@ shopt -s dotglob
 echo "---> Installing application source ..."
 mv /tmp/src/* ./
 
-if [[ ! -z "$UPGRADE_PIP_TO_LATEST" ]]; then
+if [[ ! -z "$UPGRADE_PIP_TO_LATEST" || ! -z "$ENABLE_PIPENV" ]]; then
   echo "---> Upgrading pip to latest version ..."
   pip install -U pip setuptools wheel
 fi
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,8 @@ var/
 *.egg-info/
 .installed.cfg
 *.egg
+.image-id
+.image-id.raw
 
 # PyInstaller
 #  Usually these files are written by a python script from a template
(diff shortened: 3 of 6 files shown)


omron93:
@mhutter Thanks for the PR.

We were thinking about this problem too and came with approach of generating files from templates. See https://github.com/sclorg/postgresql-container/pull/193 for more information.

2.7/s2i/bin/assemble
```diff
@@ -15,6 +15,7 @@ function should_collectstatic() {
 # (pip script installer).
 function install_pipenv() {
   echo "---> Installing pipenv packaging tool ..."
+  pip install -U virtualenv
```
mcyprian: Each of the images contains different Python version and different software collection, this is the reason why also s2i scripts have to be different in some cases. 
**python-34-centos7** image originally contains ancient versions of pip and virtualenv:
```
docker run -it centos/python-34-centos7 bash
(app-root)pip --version
pip 1.5.6 from /opt/app-root/lib/python3.4/site-packages (python 3.4)
(app-root)virtualenv --version
1.11.6
```
while **centos/python-35-centos7** image contains **pip 7.1.2** and **virtualenv 13.1.2**.
The reason why virtualenv package is updated from PyPI in 34 image is that old versions of pip and virtualenv are incompatible with code that follows. In 2.7, 3.5 and 3.6  overwriting of the virtualenv binary from collection by package from PyPI is not necessary to make Pipenv work.

3.6/s2i/bin/assemble
```diff
@@ -28,7 +29,7 @@ shopt -s dotglob
 echo "---> Installing application source ..."
 mv /tmp/src/* ./
 
-if [[ ! -z "$UPGRADE_PIP_TO_LATEST" ]]; then
+if [[ ! -z "$UPGRADE_PIP_TO_LATEST" || ! -z "$ENABLE_PIPENV" ]]; then
```
mcyprian: This is similar case to `pip install -U virtualenv` above. Update of pip from PyPI is not needed for 3.6.

2.7/s2i/bin/assemble
mhutter: > In 2.7, 3.5 and 3.6 overwriting of the virtualenv binary from collection by package from PyPI is not necessary to make Pipenv work.

But it doesn't hurt either, and that way all containers have the same, up-to-date versions of pip and virtualenv (both of which are outdated in all other containers)

The point of this PR is to bring the versions in line  with each other. Because I'm certainly NOT going to create PR's in a repo where I have to do the _same_ change in _four_ different places _by hand_.

mcyprian: I see your point, but containers should use scl content in case there is not good reason to use something else, what do you thing @hhorak?

mcyprian:
@omron93 as far as I know distgen templates are used to generate the Dockerfiles not assemble/run scripts. Distgen templates for Python container are ["under construction"]( https://github.com/mcyprian/s2i-python-container/commit/bc0e241b308c1a5f5221156f1873b3b9e886488f).

pkubatrh:
@mcyprian With the most recent changes in common-scripts you should be able to template any kind of file you want.

Check https://github.com/sclorg/postgresql-container/pull/203 for an example.

mcyprian:
@pkubatrh alright, good to know, I will try to include templating of s2i script in my distgen support PR.

mhutter:
That 'd be awesome, with distgen this PR would be obsolete!

mcyprian:
All the version specific files are now generated from templates https://github.com/sclorg/s2i-python-container/pull/259, closing this.

PR #251, opened 2018-02-27
Use a variable for NodeJS SCL name

hhorak:
This will allow us to update to newer nodejs smoothly.


hhorak:
Simple enough, hope tests passing is good enough review.

PR #256, opened 2018-03-19
use local reference policy for tags

bparees:


omron93:
[test]

bparees:
ci failure seems unrelated to this change

omron93:
You are right.
@hroncok @torsava Please, could you take a look? Newer pipenv is being installed, so the failure is probably somehow connected...

hroncok:
> Please, could you take a look?

On what? I'm not following this discussion, sorry.

omron93:
> On what? I'm not following this discussion, sorry.

Oh, sorry:-( It should be @mcyprian instead...

hroncok:
No problem :) Edited. Clutter removed.

mcyprian:
I am looking into it.

PR #257, opened 2018-03-21
Add content_sets.yml

pkubatrh:


mcyprian:
When I run yum list installed inside the container there are some base packages marked as installed from \@anaconda/X.Y repo. Is this excluded from content_sets on purpose?

pkubatrh:
I expect these come from the base rhel image and should also be present in the `rhel-7-server-rpms` repository.

PR #258, opened 2018-03-21
Make sure Pipenv is installed into updated virtualenv on 2.7 and 3.5

mcyprian:
[The latest release of Pipenv](https://github.com/pypa/pipenv/releases/tag/v11.9.0)  uses advanced syntax to define install requiremets in setup.py script. Pip version bundled inside virtualenv present in rh-python35 and python27 collections is too old to handle this syntax, which causes CI failures. Virtualenv has to be updated to the latest version in order to install Pipenv successfully.

### DIFF
--- a/2.7/s2i/bin/assemble
+++ b/2.7/s2i/bin/assemble
@@ -15,6 +15,7 @@ function should_collectstatic() {
 # (pip script installer).
 function install_pipenv() {
   echo "---> Installing pipenv packaging tool ..."
+  pip install -U virtualenv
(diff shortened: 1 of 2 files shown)


mcyprian:
[test]

omron93:
Thanks. @mcyprian Versions 3.4 and 3.6 are fine?

[test][test-openshift]

mcyprian:
3.6 is fine, 3.4 does the update from the beginning because of different issue caused by ancient pip version.

omron93:
Thanks for quick fix of this problem. Merging.

PR #259, opened 2018-04-04
Generate Dockerfiles using distgen

### LINKED ISSUE #244: Use distgen to generate Dockerfiles
[Postgresql-container](https://github.com/sclorg/postgresql-container) already uses [distgen](https://github.com/devexp-db/distgen) templating system to generate Dockerfiles. Let's try to implement something similar for s2i-python-container to decrease an amount of maintenance work when updating Dockerfiles.

mcyprian:
Using distgen to generate Dockerfiles will decrease an amount of maintenance work on this project.
This is only the first step, I would like to render more files (directories) from the templates (see #250), more PRs are coming.

To generate Dockerfiles it is necessary to use the latest distgen from GitHub (not released yet), containing [jinja2 behavior adjustment](https://github.com/devexp-db/distgen/commit/940674463f8d76c69e785e2d3c6016b7c9b2426b).

The generate-all makefile rule provided by common-scripts has to be [accustomed](https://github.com/sclorg/container-common-scripts/pull/75) to work for python container correctly. For now custom rule similar to this:
```
DG ?= /usr/bin/dg
DG_EXEC = ${DG} --max-passes 25 --template src/Dockerfile.template --multispec specs/multispec.yml

generate-dockerfiles:
    for version in ${VERSIONS}; do \
        ${DG_EXEC} --distro centos-7-x86_64.yaml --multispec-selector version=$${version} --output $${version}/Dockerfile; \
        ${DG_EXEC} --distro rhel-7-x86_64.yaml --multispec-selector version=$${version} --output $${version}/Dockerfile.rhel7; \
        ${DG_EXEC} --distro fedora-26-x86_64.yaml --multispec-selector version=$${version} --output $${version}/Dockerfile.fedora; \
    done
```
can be added to s2i-python-container's Makefile to check/review how this works. Fixes #244.


mcyprian:
This was rebased after update of Fedora Dockerfile https://github.com/sclorg/s2i-python-container/pull/262.
The makefile rule to generate Dockerfiles will now look like this:

```
DG ?= /usr/bin/dg
DG_EXEC = ${DG} --max-passes 25 --template src/Dockerfile.template --multispec specs/multispec.yml

generate-dockerfiles:
    for version in ${VERSIONS}; do \
        ${DG_EXEC} --distro centos-7-x86_64.yaml --multispec-selector version=$${version} --output $${version}/Dockerfile; \
        ${DG_EXEC} --distro rhel-7-x86_64.yaml --multispec-selector version=$${version} --output $${version}/Dockerfile.rhel7; \
        ${DG_EXEC} --distro fedora-27-x86_64.yaml --multispec-selector version=$${version} --output $${version}/Dockerfile.fedora; \
    done
```

mcyprian:
[test]

torsava:
[test]

2.7/Dockerfile
```diff
@@ -33,16 +33,16 @@ LABEL summary="$SUMMARY" \
...
+        libffi-devel libtool-ltdl enchant" && \
+    yum install -y centos-release-scl && \
+    yum -y --setopt=tsflags=nodocs install --enablerepo=centosplus $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
-    # Remove centos-logos (httpd dependency, ~20MB of graphics) to keep image
-    # size smaller.
+    # Remove centos-logos (httpd dependency) to keep image size smaller.
     rpm -e --nodeps centos-logos && \
-    yum clean all -y
+    yum -y clean all --enablerepo='*'
```
torsava: Why `--enablerepo='*'`?

mcyprian: This is how cleancache command is defined in distgen tool https://github.com/devexp-db/distgen/blob/master/distgen/commands.py#L72. I simply call it's method in the template https://github.com/sclorg/s2i-python-container/pull/259/commits/cabe677976b28eaca14f26d952fce561a7715460#diff-bd0e57c1ca2f20c05a1456d4512e2eafR55.

specs/multispec.yml
```diff
@@ -0,0 +1,79 @@
...
+      prod: "centos7"
+      logos: "centos-logos"
+      enablerepo: "--enablerepo=centosplus"
+      preinstall_cmd: ['yum install -y centos-release-scl']
+      python_pkgs: ['{{ spec.scl }}', '{{ spec.scl }}-python-devel',
+                    '{{ spec.scl }}-python-setuptools', '{{ spec.scl }}-python-pip']
+      base_pkgs: ['nss_wrapper', 'httpd24', 'httpd24-httpd-devel', 'httpd24-mod_ssl',
+                  'httpd24-mod_auth_kerb', 'httpd24-mod_ldap', 'httpd24-mod_session',
+                  'atlas-devel', 'gcc-gfortran', 'libffi-devel', 'libtool-ltdl enchant']
+      extra_pkgs: ['libjpeg-turbo', 'libjpeg-turbo-devel']
```
torsava: Since `extra_pkgs` is only used for python27, would it be possible to create a dictionary here? For example:

    extra_pkgs:
      - 2.7: ['libjpeg-turbo', 'libjpeg-turbo-devel']

That way we could hopefully access the variable as `spec.extra_pkgs[spec.version]`.

If this turns out not to work, please rename the variable to be specific about the python version, for example `python27_extra_pkgs`, so the usage is clearer.

2.7/Dockerfile
omron93: I think difference is mainly in rhel7 Dockerfile.

For first `yum` call default repositories are enabled - so metadata are downloaded. Then `prepare-yum-repositories` disable these repositories... And `yum clean` cleans only enabled repositories, that's why `--enablerepo='*'` is needed.

mcyprian:
All the files in version subdirectories are now generated from templates stored under src/. Also the issue reported in https://github.com/sclorg/s2i-python-container/pull/250 is cover by this change.

mcyprian:
Fedora test suit need to be fixed, see https://github.com/sclorg/s2i-python-container/issues/264, the failure is not related to this PR changes.

torsava:
[test]

specs/multispec.yml
```diff
@@ -0,0 +1,79 @@
...
+      python_pkgs: ['python3', 'python3-devel', 'python3-setuptools', 'python3-pip',
+                    'python3-virtualenv']
+      base_pkgs: ['nss_wrapper', 'httpd', 'httpd-devel', 'atlas-devel', 'gcc-gfortran',
+                  'libffi-devel', 'libtool-ltdl enchant']
+
+  version:
+    "3.6":
+      version: "3.6"
+      short_ver: "36"
+      img_version: "1"
```
torsava: Could you move the "img_version" to the distro instead, ideally into a dictionary as suggested above, or as "python36_img_version" for better readability?

src/Dockerfile.template
```diff
@@ -0,0 +1,73 @@
+{% import "src/common.tpl" as common with context %}
+{% import "src/" + config.os.id + "/macros.tpl" as macros with context %}
+{% if spec.version == "2.7" and config.os.id == 'centos' %}
```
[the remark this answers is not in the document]
torsava: In addition, consider moving this definition of the `extra_pkgs` variable near where it is used in the template, so both the definition and usage is next to each other for better readability.

specs/multispec.yml
mcyprian: Img_version is not distro specific. it is used on more places not only the image tag https://github.com/mcyprian/s2i-python-container/commit/cabe677976b28eaca14f26d952fce561a7715460#diff-c523d7243e565a463b4a2d07a9e2c40fR4. It was introduced in version 3.6 so it is logical to have it defined only for this version. After python 3.7 is added something like python36_img_version won't work.

mcyprian:
[test]

src/Dockerfile.template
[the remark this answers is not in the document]
mcyprian: I like to have initialization of variable at the top of the file for some reason. The rest of the template have clean structure of the Dockerfile that is rendered from it thanks to this approach.

omron93:
@mcyprian You have to set you membership in sclorg/ github org as **public**, to be able to trigger CentOS and Fedora CI.

[test]

src/test/pipenv-test-app/Pipfile.lock
```diff
@@ -0,0 +1,385 @@
+{% if spec.version == "3.6" %}
```
torsava: Wouldn't this be better to separate these into different files? That way if you want to regenerate it, you just overwrite the file instead of having to copy paste contents into this big jumble.

mcyprian: It probably would be better, unfortunately design of manifest.sh https://github.com/mcyprian/s2i-python-container/blob/ultimate-distgen/manifest.sh#L29 file doesn't allow to have different source files for different versions. The whole idea of manifest and generate rule is to generate everything from one collection of source files/templates.

specs/multispec.yml
mcyprian: Dockerfile for centos simply doesn't contain tag in FROM declaration https://github.com/mcyprian/s2i-python-container/blob/ultimate-distgen/3.6/Dockerfile#L3

specs/multispec.yml
```diff
@@ -0,0 +1,79 @@
+version: 1
+
+specs:
+  distroinfo:
+    centos:
```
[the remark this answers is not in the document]
mcyprian: I copied it from multispec.yml postgresql-container useshttps://github.com/sclorg/postgresql-container/blob/master/specs/multispec.yml#L16, but you are right, it doesn't really make sense. I will try to use just rhel here.

src/fedora/macros.tpl
```diff
@@ -0,0 +1,24 @@
+{% macro env_metadata() %}
```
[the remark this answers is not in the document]
mcyprian: src/rhel is a symlink to src/centos, the macros.tpl file is used for both centos and rhel. This import https://github.com/mcyprian/s2i-python-container/commit/cabe677976b28eaca14f26d952fce561a7715460#diff-bd0e57c1ca2f20c05a1456d4512e2eafR2 wouldn't work otherwise.

src/Dockerfile.template
```diff
@@ -1,10 +1,11 @@
 {% import "src/common.tpl" as common with context %}
 {% import "src/" + config.os.id + "/macros.tpl" as macros with context %}
-{% if spec.version == "2.7" and config.os.id == 'centos' %}
-{%   set extra_pkgs = spec.extra_pkgs %}
-{% else %}
-{%   set extra_pkgs = [] %}
-{% endif %}
+{% set ns = namespace(extra_pkgs=[]) %}
```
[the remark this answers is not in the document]
mcyprian: The extra_pkgs variable assigned inside for loop scope cannot be referenced outside the scope without the namespace

src/centos/macros.tpl
```diff
@@ -0,0 +1,25 @@
+{% macro env_metadata() %}{% endmacro %}
+
+{% macro _version_selector(spec) %}
+{% if spec.img_version %}{{ spec.img_version }}{% else %}{{ spec.version }}{% endif %}
```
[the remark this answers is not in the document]
mcyprian: Yes you are right, I haven't realized this difference

mcyprian:
[test]

src/Dockerfile.template
```diff
@@ -6,8 +6,8 @@
 {%     set ns.extra_pkgs = item.value %}
 {%   endif %}
 {% endfor %}
-{% if config.os.id == 'rhel' and spec.img_version %}
-{%   set img_tag = spec.img_version %}
+{% if config.os.id == 'rhel' and spec.base_img_version %}
```
torsava: If the base_img_version is only used for rhel, how about `rhel_base_img_version` to make it clearer? Sorry for nitpicking, I just hate to debug these little things so I like to avoid them if possible.

src/fedora/macros.tpl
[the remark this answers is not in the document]
torsava: Ah, I completely overlooked that, thanks!

src/test/pipenv-test-app/Pipfile.lock
torsava: As we talked, the different files could be included from this one.

mcyprian: Fixed.

src/test/pipenv-test-app/2.7/pipfile.tpl
```diff
@@ -0,0 +1,96 @@
+{% macro content() %}
+{
```
torsava: Thanks! That's much better than having it in one file. However, would it be possible to solve without having to modify the individual `Pipfile.lock` files by adding this macro definition on top? In an ideal situation, one would just copy  the `Pipfile.lock` file into the proper location, overwrite what's there and that'd be it. No modifying of the contents of the file.

Perhaps the `include` directive can be used in the `src/test/pipenv-test-app/Pipfile.lock` file?

    {% include 'src/test/pipenv-test-app/" + spec.version + "/Pipfile.lock' %}

mcyprian: This also works

mcyprian:
Rebased after merge of https://github.com/sclorg/s2i-python-container/pull/267.

mcyprian:
[test-openshift]

torsava:
I went through the new commits and the previous modifications and everything looks in order.

One more thing: We should clarify the minimum jinja2 version requirement in the README alongside the distgen I think.

And after the [container-common-scripts PR](https://github.com/sclorg/container-common-scripts/pull/75) is merged, let's merge this.

mcyprian:
Thanks for the review @torsava, I added note about jinja2 to README and squashed fixup commit.

PR #270, opened 2018-05-07
Fix compatibility with Jinja <= 2.9

praiskup:
The namespace() feature works since 2.10+.  Also update common for
pretty-printing output of the 'make generate' rule.


mcyprian:
LGTM, thanks

PR #271, opened 2018-05-14
Deprecate version 3.4

pkubatrh:


mcyprian:
- Please rebase this on top of master, I realized README files were not up to date after merge of the last PR. 
- Version 3.4 should be probably removed from specs/multispec.yml as well

pkubatrh:
>Version 3.4 should be probably removed from specs/multispec.yml as well

Ah yes, I thought it would suffice to remove it from the Makefile but it seems the Dockerfiles are still being generated, so I have now removed 3.4 from multispec as well.

mcyprian:
Thanks, LGTM

PR #273, opened 2018-06-01
Document npm purpose and no guarantee for version

hhorak:
SSIA

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -20,6 +20,11 @@ Python's elegant syntax and dynamic typing, together with its interpreted nature
 make it an ideal language for scripting and rapid application development in many areas 
 on most platforms.
 
+This container image includes an npm utility, so users can use it to install JavaScript
+modules for their web applications. There is no guarantee for any specific npm or nodejs
+version, that is included in the image; those versions can be changed anytime and
+the nodejs itself is included just to make the npm work.
+
 Usage
 ---------------------
 To build a simple [python-sample-app](https://github.com/sclorg/s2i-python-container/tree/master/2.7/test/setup-test-app) application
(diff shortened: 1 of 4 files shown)


pkubatrh:
[test]

mcyprian:
We use to divide all changes into two separate commits since we use distgen for this repo. First we commit changes of templates/sources under /src directory and than we commit regenerated directories for Python versions. This is not a hard requirement just a note.

mcyprian:
LGTM otherwise.

hhorak:
I'll divide it, that makes sense. (sorry my ignorance, this is my first commit to dist-gen-ed repo :)

hhorak:
Tests fail now due infra, but passed previously, so merging.

omron93:
>We use to divide all changes into two separate commits since we use distgen for this repo. First we commit changes of templates/sources under /src directory and than we commit regenerated directories for Python versions. This is not a hard requirement just a note.

@mcyprian Do you plan to remove generated sources? Like postgresql-container did - it keeps generated sources in automatically updated `generated` branch.

mcyprian:
@omron93 We definitely plan to do so. We've decided to commit sources for a while to make the switch to distgen smooth.

We are not  sure what is the right way to make the generated sources diff visible. The goal is not to make the work of PR reviewer unnecessarily difficult. As soon as there is some good solution to handle these kind of things in postgresql-container, we will be happy to switch to this model as well. **FYI** @torsava

torsava:
@omron93 As Michal said, we'd like the diff of the generated sources be visible during a PR. And the easiest way to do that is to have the generated sources be part of this repo.

However, perhaps it would be possible to have some bot that automatically regenerates the sources for every PR and links to a diff? We could work together on such a mechanism, as it would be useful for all distgen-based repos.

PR #279, opened 2018-06-14
Update dockerfiles to use new BZ components

pkubatrh:


pkubatrh:
CI tests passed locally, and I don't think this change could introduce a CI failure in any case. Merging.

Issue #272, opened 2018-05-22
`make generate` creates symlinks inside examples/*


pkubatrh:
Running `make generate` on a fresh cloned repository creates symlinks inside example app sources, pointing to itself:

```
$ ls -l examples/app-home-test-app/app-home-test-app
examples/app-home-test-app/app-home-test-app -> ../../examples/app-home-test-app
```

This results in a failure in my rebuild tooling due to it following the symlink each time it is encountered (it is rather dumb):

```
OSError: [Errno 40] Too many levels of symbolic links: 'upstreams/python/2.7/test/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app/numpy-test-app'
```

I will make my tooling a bit smarter to work around this but as I do not think this is done on purpose - likely the creation of symlinks inside `$VERSION/test` was expected - I am creating an issue over here as well.

PR #281, opened 2018-06-19
 Generate actual test files instead of symlinks 

pkubatrh:
Resolves: #272


pkubatrh:
@torsava Can you take a look if the change is ok with you?

imo it is no longer necessary to symlink everything now that we are using source generation.

torsava:
@pkubatrh Sorry it took me a while to get to this.

Copying the files instead of symlinking solves the immediate issue, but it heavily pollutes the regenerated diffs, so I'd like to avoid it if possible.

I've tried to track down the source of the issue, and it seems it's a bug in `common/generate.sh`. If you delete all variants (2.7, 3.5, 3.6) it generates fine. If you delete only some variants, it adds the cyclic symlinks, because the symlinks are already there, so it adds them into the symlinked directory, causing the issue.

I don't have too much time to dig into this at the moment to go dig into common scripts. How critical is this, should we merge this as a (maybe) temporary workaround?

pkubatrh:
Not critical at all, just a pita since I need to work around the issue during the sync with github content.

I did not try to dig into the code to find the issue behind this since I felt it would be better to get rid of symlinks entirely now that we have code generation but as you say the secondary use-case of not polluting diffs is still there.

pkubatrh:
We decided off-line some time ago to not go with generating all of the sources with distgen yet.

Additionally #272 has been fixed through https://github.com/sclorg/container-common-scripts/pull/98

Closing for now.

PR #286, opened 2018-08-15
update imagestreams to registry.redhat.io

bparees:


bparees:
@hhorak bump

omron93:
More detailed reasoning from postgresql-container PR:
> registry.redhat.io is the future of the redhat container catalog, registry.access.redhat.com is going to be shut down at some point in the future. The new registry requires authentication and the local reference policy allows the openshift internal registry to perform that authentication on the user's behalf so that every use on a cluster doesn't have to provide their own registry.redhat.io credentials.

Makes sense and already merged for some containers. So merging too.

PR #287, opened 2018-09-18
Update common submodule and fix tests

hhorak:
It is necessary to define `NAME` variable, so we don't need to guess names of the services.


hhorak:
[test-openshift]

hhorak:
[test-openshift]

hhorak:
The Fedora failure is not related to this change, merging.

PR #289, opened 2018-10-08
OKD rebranding and sclorg move for s2i-python-container

liangxia:
docs.openshift.org to docs.okd.io
github.com/openshift/django-ex to github.com/sclorg/django-ex


omron93:
Looks good, thanks.

PR #293, opened 2018-10-09
using Fedora 28 image as a base

goern:
This is changing only Python 3.6


omron93:
Thanks. F28 have python 3.6 too. So merging.

PR #294, opened 2018-10-24
Activate virtualenv in Fedora image

omron93:
- scl_enable isn't sourced automatically in s2i-core because it
would cause 'scl_source: No such file or directory' errors in most
of the images

Remove arch label - 'Optional: if omitted, it will be built for
all supported Fedora Architectures'

@pkubatrh @torsava Please review and merge.


omron93:
[test-openshift]

PR #303, opened 2019-01-27
Add support for incremental builds

rarguello:
Add support for incremental builds


rarguello:
Fixes #304

frenzymadness:
Unfortunately, the situation here is the same as in PR #306 . We don't maintain scripts for each Python version. The content of folders with a Python version in the name is automatically generated and scripts are copied from `src/s2i/bin/`.

Could you please update your PR?

frenzymadness:
Closing this because it's outdated but let's keep the issue open.

PR #312, opened 2019-02-08
Add Integration tests to be run in a remote cluster

hhorak:
The idea is to have an easy way to run tests in an existing cluster, that might be remote via `make test-openshift-remote-cluster` or directly via `test/run-openshift-remote-cluster`.


hhorak:
[test-openshift]

pkubatrh:
[test] [test-openshift]

pkubatrh:
@hhorak Symlink to test-lib-openshift.sh missing from versioned directories:
```
test/run-openshift: line 13: test/test-lib-python.sh: No such file or directory
```

pkubatrh:
For 2.7 that is

pkubatrh:
I expect you will also need to add the symlinks to:
https://github.com/sclorg/s2i-python-container/blob/master/manifest.sh

frenzymadness:
Can I help here somehow?

frenzymadness:
Is this still valid?

hhorak:
The aim to have tests that run with remote cluster is definitely valid, as OCP 4.x does not support `oc cluster up` thing that was used to test the containers on OCP 3.x. Thanks for reminding, this needs more focus now than before.

frenzymadness:
Does the #381 implement the same? Can we close this one?

hhorak:
> Does the #381 implement the same? Can we close this one?

Yes, closing this one.

PR #316, opened 2019-02-15
Fedora 29 & Python 3.7

frenzymadness:
Second try to add Fedora 29 and Python 3.7. Now, content in `3.7` is generated from the source. I found some issues in the process and generated outputs so I open issues for them.

This PR is the draft (new github feature) and I'll change that once all tests will pass.


frenzymadness:
[test]

frenzymadness:
[test-openshift]

frenzymadness:
@torsava Could you please take a look why there are only rhel7 tests?

torsava:
[test]

torsava:
@frenzymadness Apparently, the bot doesn't like you :) Could you try to set your membership in @sclorg public? I believe it checks that.

frenzymadness:
[test-openshift]

frenzymadness:
Tests are passing. PR is ready for review.

2.7/s2i/bin/assemble
```diff
@@ -10,7 +10,8 @@ function should_collectstatic() {
 
 function virtualenv_bin() {
   if head "/etc/redhat-release" | grep -q "^CentOS Linux release 7" || \
-     head "/etc/redhat-release" | grep -q "^Red Hat Enterprise Linux\( Server\)\? release 7"; then
+     head "/etc/redhat-release" | grep -q "^Red Hat Enterprise Linux\( Server\)\? release 7" || \
+     head "/etc/redhat-release" | grep -q "^Fedora release 29 \(Twenty Nine\)\?"; then
```
torsava: `virtualenv-3.7` doesn't work on F29? If we it does, I'd prefer to use it: to be explicit, and to keep uniformity with F28.

torsava: And if it doesn't, could we do it the other way and use unversioned `virtualenv` also on F28? Or is that still python2?

src/fedora/macros.tpl
```diff
@@ -15,7 +15,11 @@ ENV NAME=python3 \
 {% endmacro %}
 
 {% macro permissions_setup(spec) %}
+{% if spec.fedora_version == '28' %}
 RUN virtualenv-$PYTHON_VERSION ${APP_ROOT} && \
+{% else %}
+RUN virtualenv ${APP_ROOT} && \
```
torsava: Same here with virtualenv-X.Y vs virtualenv.

3.7/root/opt/app-root/etc/scl_enable
```diff
@@ -4,6 +4,6 @@
 unset BASH_ENV PROMPT_COMMAND ENV
 if head "/etc/redhat-release" | grep -q "^CentOS Linux release 7" || \
    head "/etc/redhat-release" | grep -q "^Red Hat Enterprise Linux\( Server\)\? release 7"; then
-    source scl_source enable httpd24 $NODEJS_SCL rh-python36
+    source scl_source enable httpd24 $NODEJS_SCL
```
torsava: Even though we don't need it right now, I would probably add the rh-python37 definition into the spec, so this doesn't bite us later.

2.7/s2i/bin/assemble
frenzymadness: Unless I am missing something, this is the only way how to do it because:

* `virtualenv` is provided by python2-virtualenv on F28.
```
# rpm -qf /usr/bin/virtualenv
python2-virtualenv-16.0.0-4.fc28.noarch
```
* and python3-virtualenv provides only `virtualenv` without any suffix on F29.
```
# rpm -ql python3-virtualenv | grep bin
/usr/bin/virtualenv
```

src/fedora/macros.tpl
frenzymadness: Same comment here.

3.7/root/opt/app-root/etc/scl_enable
frenzymadness: Fixed in two fixup commits - one for multispec and one for generated content.

frenzymadness:
[test]

2.7/s2i/bin/assemble
torsava: Damn, that's pretty horrible... but what can we do. Ok, LGTM.

torsava:
Discussions resolved, and the new commits look good. Ready for merge.

torsava:
Tests are passing, code looks good, I'm merging the PR!

Issue #314, opened 2019-02-15
Generated documentation is not correct


frenzymadness:
Documentation generated from `src/README.md` should contain only usage examples for a platform which the Python version is generated for.

For example, there is no Python 3.5 in RHEL8 but you can find it in the readme: https://github.com/sclorg/s2i-python-container/tree/master/3.5#usage

Lorquas:
+1 ; none of the usage examples for 3.7 work:  https://github.com/sclorg/s2i-python-container/blob/master/3.7/README.md

frenzymadness:
Yes because examples are generated for Centos and RHEL but Python 3.7 is available only in Fedora yet.

I am gonna take a look on PR.

PR #318, opened 2019-03-07
README.md: do not generate platform specific content

pkubatrh:
Fixes: #314

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -3,12 +3,13 @@ Python 2.7 container image
 
 This container image includes Python 2.7 as a [S2I](https://github.com/openshift/source-to-image) base image for your Python 2.7 applications.
 Users can choose between RHEL and CentOS based builder images.
-The RHEL7 image is available in the [Red Hat Container Catalog](https://access.redhat.com/containers/#/registry.access.redhat.com/rhscl/python-27-rhel7)
-as registry.access.redhat.com/rhscl/python-27-rhel7
-and the RHEL8 image as [registry.access.redhat.com/rhel8/python-27](https://access.redhat.com/containers/#/registry.access.redhat.com/rhel8/python-27).
-The CentOS7 image is then available on [Docker Hub](https://hub.docker.com/r/centos/python-27-centos7/)
-as centos/python-27-centos7.
-The resulting image can be run using [Docker](http://docker.io).
+The RHEL images are available in the [Red Hat Container Catalog](https://access.redhat.com/containers/),
+the CentOS images are available on [Docker Hub](https://hub.docker.com/r/centos/),
+and the Fedora images are available in [Fedora Registry](https://registry.fedoraproject.org/).
+The resulting image can be run using [podman](https://github.com/containers/libpod) or
+[docker](http://docker.io).
+
+Note: while the examples in this README are calling `podman`, you can replace any such calls by `docker` with the same arguments
 
 Description
 -----------
@@ -28,26 +29,19 @@ the nodejs itself is included just to make the npm work.
 
 Usage
 ---------------------
-To build a simple [python-sample-app](https://github.com/sclorg/s2i-python-container/tree/master/2.7/test/setup-test-app) application
-using standalone [S2I](https://github.com/openshift/source-to-image) and then run the
-resulting image with [Docker](http://docker.io) execute:
 
-*  **For RHEL7 based image**
+For this, we will assume that you are using the `rhscl/python-27-rhel7 image`, available via `python:2.7` imagestream tag in Openshift.
+Building a simple [python-sample-app](https://github.com/sclorg/s2i-python-container/tree/master/2.7/test/setup-test-app) application
+in Openshift can be achieved with the following step:
+
+    ```
+    oc new-app python:3.6~https://github.com/sclorg/s2i-python-container.git --context-dir=2.7/test/setup-test-app/
+    ```
+
+The same application can also be built using the standalone [S2I](https://github.com/openshift/source-to-image) application on systems that have it available:
+
     ```
     $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=2.7/test/setup-test-app/ rhscl/python-27-rhel7 python-sample-app
-    $ docker run -p 8080:8080 python-sample-app
-    ```
-
-*  **For RHEL8 based image**
-    ```
-    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=2.7/test/setup-test-app/ rhel8/python-27 python-sample-app
-    $ docker run -p 8080:8080 python-sample-app
-    ```
-
-*  **For CentOS7 based image**
-    ```
-    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=2.7/test/setup-test-app/ centos/python-27-centos7 python-sample-app
-    $ docker run -p 8080:8080 python-sample-app
     ```
 
 **Accessing the application:**
@@ -232,11 +226,11 @@ configuration file inside your repository with the
 option set to `true`. Make sure to specify your config via the `APP_CONFIG`
 environment variable.
 
-To change your source code in running container, use Docker's
-[exec](https://docs.docker.com/reference/commandline/exec/) command:
+To change your source code in running container, use podman's
+[exec](https://github.com/containers/libpod/blob/master/docs/podman-exec.1.md) command:
 
 ```
-docker exec -it <CONTAINER_ID> /bin/bash
+podman exec -it <CONTAINER_ID> /bin/bash
 ```
 
 After you enter into the running container, your current directory is set
@@ -247,4 +241,5 @@ See also
 --------
 Dockerfile and other sources are available on https://github.com/sclorg/s2i-python-container.
 In that repository you also can find another versions of Python environment Dockerfiles.
-Dockerfile for CentOS is called `Dockerfile`, Dockerfile for RHEL7 is called `Dockerfile.rhel7` and for RHEL8 it's `Dockerfile.rhel8`.
+Dockerfile for CentOS is called `Dockerfile`, Dockerfile for RHEL7 is called `Dockerfile.rhel7`,
+for RHEL8 it's `Dockerfile.rhel8` and the Fedora Dockerfile is called Dockerfile.fedora.
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -226,7 +226,7 @@ configuration file inside your repository with the
 option set to `true`. Make sure to specify your config via the `APP_CONFIG`
 environment variable.
 
-To change your source code in running container, use podman's
+To change your source code in running container, use podman's (or docker's)
 [exec](https://github.com/containers/libpod/blob/master/docs/podman-exec.1.md) command:
 
 ```
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -242,4 +242,4 @@ See also
 Dockerfile and other sources are available on https://github.com/sclorg/s2i-python-container.
 In that repository you also can find another versions of Python environment Dockerfiles.
 Dockerfile for CentOS is called `Dockerfile`, Dockerfile for RHEL7 is called `Dockerfile.rhel7`,
-for RHEL8 it's `Dockerfile.rhel8` and the Fedora Dockerfile is called Dockerfile.fedora.
+for RHEL8 it's `Dockerfile.rhel8` and the Fedora Dockerfile is called `Dockerfile.fedora`.
--- a/README.md
+++ b/README.md
@@ -5,7 +5,8 @@ This repository contains the source for building various versions of
 the Python application as a reproducible container image using
 [source-to-image](https://github.com/openshift/source-to-image).
 Users can choose between RHEL, Fedora and CentOS based builder images.
-The resulting image can be run using [Docker](http://docker.io).
+The resulting image can be run using [podman](https://github.com/containers/libpod) or
+[docker](http://docker.io).
 
 For more information about using these images with OpenShift, please see the
 official [OpenShift Documentation](https://docs.okd.io/latest/using_images/s2i_images/python.html).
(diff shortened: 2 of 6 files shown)


frenzymadness:
I don't think that I am the right person to do a review but I did :slightly_smiling_face: and I have two questions:
1. Would it be possible to add some example usage for Fedora so Python 3.7 redme would not be empty?
2. In `multipes.yml` file, we have `ignore` field for each Python version and then also `matrix` where is info about which distro skip for which Python version. I might be wrong but it feels like storing the same information in two different forms. Would it be possible to merge them somehow? I'd prefer the new `ignore` field instead of `matrix`.

If this is reasonable and doable but out of scope for this PR, let me know and we can discuss it later and then I can take a look at that.

Our plan is also to rename `rhel` to `rhel7` and `fedora` to `fedora28` so there won't be any unversioned names of distros.

@torsava what do you think?

pkubatrh:
>Would it be possible to add some example usage for Fedora so Python 3.7 redme would not be empty?

Sure, will add an example.

>In multipes.yml file, we have ignore field for each Python version and then also matrix where is info about which distro skip for which Python version. I might be wrong but it feels like storing the same information in two different forms. Would it be possible to merge them somehow?

Yeah, its duplicate information. Though iiuc the matrix is used by distgen to generate combinations and is not available inside the template as a value. Though @praiskup might know more about distgen specific behaviour.

praiskup:
I'm not sure I got the question;  (spoiler, I'm since beginning skeptic about "multispec" feature, I dislike that -- It IMO only spread the "matrix" from single place to two places -- `Makefile` and `multispec.yml`).  So, not that I fully understand it -- do you want to access the whole matrix from the template, instead of only the relevant part?

praiskup:
If I got the question right, I think you can not ATM: https://github.com/devexp-db/distgen/blob/bfa425fb68b2a87a00c4f58b4d42ad8ae9e2ab3a/distgen/generator.py#L260

We only "enrich" the `{{ spec }}` dictionary by the relevant part of the matrix.  Feel free to change that.

pkubatrh:
Basically what I would like to be able to do in template (for a given version) is something along the lines of:

```
{% if 'rhel-8-x86_64' not in spec.matrix.exclude.distros %}
# do stuff that is specific to rhel8 for $VERSION
{% endif %}
```

praiskup:
I'm afraid that distgen fix is needed in this case.

pkubatrh:
Added an example for Fedora based images. I had to move the `fedora_version` definition in miultispec from distroinfo to version to make sure the value gets properly picked up.

frenzymadness:
Thanks for the examples.

But the `fedora_version` is the same as `el_version` and therefore should be connected with distribution instead of default Python version in that distribution.

For example, in Fedora 28 we have to use `virtualenv-3.6` command but in Fedora 29 we have just `virtualenv`. https://github.com/sclorg/s2i-python-container/blob/master/src/fedora/macros.tpl#L18

I have to take a look at how to improve the specfile and remove all duplicate info and basically how the distgen works.

pkubatrh:
>But the fedora_version is the same as el_version and therefore should be connected with distribution instead of default Python version in that distribution.

The issue with that is that `fedora_version` will only be available when the file is generated with the fedora config. However, this is only explicitly done for Dockerfiles. For other distgen generated files the first possible version-config combination is used since the file is only generated once, in one copy.

frenzymadness:
I think that the specfile needs to be cleaned and improved but I don't know distgen well enough to do it right now.

But we need to solve the original issue with incorrect readme files and for that purpose, this PR works well.

@torsava Could you please take a look as well?

torsava:
I really dislike the movement of `fedora_version` from the definition of the system to the definition of the Python version. Plus now we have the system data in 3 places: 1. matrix, 2. `ignore` field, 3. `fedora_version` field.

Sadly I agree that we can't get the information from the matrix for this purpose, but we also can't remove the matrix due to distgen.

I think a possible solution would be to change the new `ignore` field from a negative list to a positive list - e.g. `available_on = ['rhel8', 'centos8', 'fedora29']`. It's not perfect, but hopefully a step in the right direction. What do you think?

praiskup:
We are able to get new distgen into epel-7 within 14 days, if anyone wanted to look at implementing what is needed there (so the unfiltered multispec dict is available from template, in separate namespace).

pkubatrh:
@torsava Thanks for the comments.

>I think a possible solution would be to change the new ignore field from a negative list to a positive list - e.g. available_on = ['rhel8', 'centos8', 'fedora29']. It's not perfect, but hopefully a step in the right direction. What do you think?

I can certainly do that but I am not sure how that is any different. All I see is some if clauses in this PR getting negated.

>I really dislike the movement of fedora_version from the definition of the system to the definition of the Python version. 

I do not like the move either. The alternative to that is having the Fedora XY specific content in the README.md file prepared for every version supported and then having a few if clauses testing whether this specific version of Fedora is ignored/active. I decided for not going this way since it seemed like more maintenance burden and even more duplicate content.

torsava:
> > I think a possible solution would be to change the new ignore field from a negative list to a positive list - e.g. available_on = ['rhel8', 'centos8', 'fedora29']. It's not perfect, but hopefully a step in the right direction. What do you think?
> 
> I can certainly do that but I am not sure how that is any different. All I see is some if clauses in this PR getting negated.

Well, the benefits as I see them (correct me if I'm not seeing it clearly) are: 1. Unifying the `ignore` and `fedora_version` fields, and 2. if we add a new system, it won't be automatically added to all Python versions, so we only manually add it where relevant.

> 
> > I really dislike the movement of fedora_version from the definition of the system to the definition of the Python version.
> 
> I do not like the move either. The alternative to that is having the Fedora XY specific content in the README.md file prepared for every version supported and then having a few if clauses testing whether this specific version of Fedora is ignored/active. I decided for not going this way since it seemed like more maintenance burden and even more duplicate content.

I think we will only have 1 or 2 Fedora versions available at any time, so having prepared content for each of them seems ok to me. What do you think, @frenzymadness ?

pkubatrh:
>1. Unifying the ignore and fedora_version fields

The `fedora_version` field was needed just for the release number information so that proper links that used that number could be populated in the `README.md` file. Though I guess you could do some magic to get that number from the `fedoraXY` items in the `available` list... Not that well versed in jinja to be able to say how difficult that would be.

2. if we add a new system, it won't be automatically added to all Python versions, so we only manually add it where relevant.

Ack, this is a legit benefit I missed, thanks!

pkubatrh:
For now I turned the `ignore` list into `available` and modified the README branches accordingly. Will make further changes to the `fedora_version` field once we know the way forward.

torsava:
@pkubatrh Could you try if the `multispec_raw` from https://github.com/devexp-db/distgen/pull/92 could be used? That way we would have no duplication of information.

pkubatrh:
After an offline discussion we decided that instead of generating the `README.md` content by available platforms we will generate the content only for a single platform (rhel7) since the usage examples are very similar, with only the image names being different.

@torsava @frenzymadness ptal

frenzymadness:
That sounds reasonable. I'll take a look at changes soon.

src/README.md
```diff
@@ -8,7 +8,7 @@ as registry.access.redhat.com/rhscl/python-{{ spec.short_ver }}-rhel7
 and the RHEL8 image as [registry.access.redhat.com/rhel8/python-{{ spec.short_ver }}](https://access.redhat.com/containers/#/registry.access.redhat.com/rhel8/python-{{ spec.short_ver }}).
 The CentOS7 image is then available on [Docker Hub](https://hub.docker.com/r/centos/python-{{ spec.short_ver }}-centos7/)
 as centos/python-{{ spec.short_ver }}-centos7.
-The resulting image can be run using [Docker](http://docker.io).
+The resulting image can be run using [podman](https://github.com/containers/libpod).
```
frenzymadness: Could we keep also Docker here? I don't know how many people actually know podman so it would be nice to keep both technologies here and use it to tell that podman is something similar to Docker for those who don't know it yet.

src/README.md
```diff
@@ -8,7 +8,7 @@ as registry.access.redhat.com/rhscl/python-{{ spec.short_ver }}-rhel7
 and the RHEL8 image as [registry.access.redhat.com/rhel8/python-{{ spec.short_ver }}](https://access.redhat.com/containers/#/registry.access.redhat.com/rhel8/python-{{ spec.short_ver }}).
 The CentOS7 image is then available on [Docker Hub](https://hub.docker.com/r/centos/python-{{ spec.short_ver }}-centos7/)
```
frenzymadness: It would be also nice to mention that some Python versions are available on Fedora platform.

src/README.md
```diff
@@ -233,11 +226,11 @@ configuration file inside your repository with the
...
 environment variable.
 
-To change your source code in running container, use Docker's
-[exec](https://docs.docker.com/reference/commandline/exec/) command:
+To change your source code in running container, use podman's
+[exec](https://github.com/containers/libpod/blob/master/docs/podman-exec.1.md) command:
 
 ```
-docker exec -it <CONTAINER_ID> /bin/bash
+podman exec -it <CONTAINER_ID> /bin/bash
```
frenzymadness: Same here. Just a note that Docker command is actually the same.

src/README.md
frenzymadness: We can mention Fedora Dockerfiles also in "See also" section.

src/README.md
torsava: Agreed, we need at least a transition period.

src/README.md
torsava: +1

2.7/README.md
```diff
@@ -28,26 +28,19 @@ the nodejs itself is included just to make the npm work.
...
-To build a simple [python-sample-app](https://github.com/sclorg/s2i-python-container/tree/master/2.7/test/setup-test-app) application
-using standalone [S2I](https://github.com/openshift/source-to-image) and then run the
-resulting image with [Docker](http://docker.io) execute:
 
-*  **For RHEL7 based image**
-    ```
-    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=2.7/test/setup-test-app/ rhscl/python-27-rhel7 python-sample-app
-    $ docker run -p 8080:8080 python-sample-app
-    ```
+For this, we will assume that you are using the rhscl/python-27-rhel7 image, available via python:2.7 imagestream tag in Openshift.
```
torsava: Could you distinguish technical names from the surrounding text please? I.e. onclose `rhscl/python-27-rhel7 image` and `python:2.7` in backticks for easier readability.

torsava:
Overall I like the changes a lot, @pkubatrh, I think we have found a way to go here! We only need to tie up the details.

Also note that we need to also modify the README.md file in the root of the repo (i.e. unversioned), to mention both docker and podman. And ideally to mention Fedora as well (can be done later).

src/README.md
pkubatrh: We only left the mention to the actual docker application because of there not being a podman at all in rhel7. Now that podman is available in rhel 7.6 I personally would like to remove the mention of docker completely. Imo just having the link to the podman reppository, which has pretty good documentation (including [the comparison of docker to podman](https://github.com/containers/libpod/blob/master/transfer.md)) should be enough.

@hhorak WDYT about keeping a note somewhere in the README that podman usage is equivalent to docker?

src/README.md
pkubatrh: Now that I think about it this section should probably be removed completely, else we do not fix #314 ... 
Maybe just pointing to the registries without any image specific links? Something along the lines of:
`RHEL-based images are available in the [Red Hat Registry](https://access.redhat.com/containers])`

2.7/README.md
pkubatrh: Will do, thanks!

src/README.md
pkubatrh: Checked offline with @hhorak and he is in favour of leaving a note in the README as well. So will add one to the beginning of the file.

pkubatrh:
Added mentions of Fedora Dockerfiles and Fedora registry.
Made links to the registries more general so that we do not need to check if any specific version is available for a given platform.
Added a note for the use of docker instead of podman.

@torsava @frenzymadness PTAL

README.md
```diff
@@ -5,7 +5,7 @@ This repository contains the source for building various versions of
 the Python application as a reproducible container image using
 [source-to-image](https://github.com/openshift/source-to-image).
 Users can choose between RHEL, Fedora and CentOS based builder images.
-The resulting image can be run using [Docker](http://docker.io).
+The resulting image can be run using [podman](https://github.com/containers/libpod).
```
torsava: I would add "or docker" here.

src/README.md
torsava: So you please we add a `or podman` on that line? So that if people skim just the initial paragraph, they know they can use both.

src/README.md
torsava: Agreed.

torsava:
@pkubatrh Left a few cosmetic comments, but overall it looks good! Thanks for working on it.

torsava:
[test]

torsava:
[test-openshift]

pkubatrh:
Added a link to docker as well as podman.

[test-openshift] (just in case)

src/README.md
```diff
@@ -248,4 +242,5 @@ See also
 --------
 Dockerfile and other sources are available on https://github.com/sclorg/s2i-python-container.
 In that repository you also can find another versions of Python environment Dockerfiles.
-Dockerfile for CentOS is called `Dockerfile`, Dockerfile for RHEL7 is called `Dockerfile.rhel7` and for RHEL8 it's `Dockerfile.rhel8`.
+Dockerfile for CentOS is called `Dockerfile`, Dockerfile for RHEL7 is called `Dockerfile.rhel7`,
+for RHEL8 it's `Dockerfile.rhel8` and the Fedora Dockerfile is called Dockerfile.fedora.
```
frenzymadness:
```suggestion
for RHEL8 it's `Dockerfile.rhel8` and the Fedora Dockerfile is called `Dockerfile.fedora`.
```

frenzymadness:
One last nitpick (suggested change) and then you can merge it.

Thank you!

pkubatrh:
Thanks, I added the suggested change.

[test-openshift]

src/README.md
torsava: +1

Also: Suggesting changes looks great!

torsava:
Everything is passing and thus a merge is in order!

A big thanks to you, @pkubatrh, for your work here!

Issue #301, opened 2018-12-17
python-36-centos7 - breaking changes without revision tagging 


solvingj:
We've been using this image reference as a base image for a few months and one day we could no longer build from it. We began getting this error: 

```
groupadd: '1001' is not a valid group name
```

Regardless of the cause of this error, we knew the image must have changed, so we went back to this page to get the previous tag number and PIN our build to that: 

https://hub.docker.com/r/centos/python-36-centos7/tags

However, on this page, we found that there are only two tags:  `latest` and `1`, and they had been changed 6 days ago. 

From the perspective of repeatable builds and environments, it's generally an "quasi-standard practice" for docker images owned by trusted organizations like CentOS to create a new tag each time there's a new build published, and simply move the latest tag forward to that new build each time.  Would that be something you would be willing/able to do with the `sclorg` Docker images?

frenzymadness:
This is, of course, a reasonable point. In my opinion, it'd need more than just setting on docker hub because I can imagine that the image is simply rebuilt every time when a new commit lands in the master branch here, which is okay for the latest image, but we should find a way how to keep also the previous versions and how to name/tag them properly.

torsava:
@pkubatrh @praiskup Do we know who is uploading the images to Docker hub?

pkubatrh:
@torsava Done by Jenkins job, this should be the correct one I hope:
https://github.com/sclorg/rhscl-container-ci/blob/master/yaml/builders/image-build-push.yaml

To fix this we would need to add some release versioning of the images since we depend on OSBS auto-increments in both Fedora and RHEL (and I expect in builds for CentOS registry as well)

torsava:
Resolving in PRs https://github.com/sclorg/container-common-scripts/pull/113 and https://github.com/sclorg/rhscl-container-ci/pull/25.

solvingj:
Very impressed by the follow through! Thanks very muc!

PR #321, opened 2019-04-09
New common scripts create additional tags

torsava:
that will allow us to have more versions available on Docker hub

Issue #301


pkubatrh:
Merging, thanks!

pkubatrh:
Note that https://github.com/sclorg/rhscl-container-ci/pull/26 will need to be resolved and deployed first, before the new tags are available on dockerhub.

PR #327, opened 2019-05-27
Regenerated content after the last change of comments in init_script

frenzymadness:
@torsava After you altered my last pull request, we forget to regenerate the content for all versions.

This change contains only comments so it should not have any impact.


pkubatrh:
Comments only, so lgtm and merging

Issue #325, opened 2019-05-26
Fedora/Python 3.7 image should still source scl_enable.


GrahamDumpleton:
In the `Dockerfile`` for Fedora/Python 3.7 image, it does:

```
# For Fedora scl_enable isn't sourced automatically in s2i-core
# so virtualenv needs to be activated this way
ENV BASH_ENV="${APP_ROOT}/bin/activate" \
    ENV="${APP_ROOT}/bin/activate" \
    PROMPT_COMMAND=". ${APP_ROOT}/bin/activate"
```

This should not be using `${APP_ROOT}/bin/activate` directly for these environment variables. It should still be using `/opt/app-root/etc/scl_enable`. This is because derived images which have added additional startup actions to the `scl_enable` script will now fail, since those steps will be skipped.

If changes are needed because actions to enable packages are different, or not required, it should be handled in the `scl_enable` script, and not by changing what these environment variables trigger.

frenzymadness:
Could you please take a look at the PR?

PR #328, opened 2019-05-27
Fedora should use scl_enable script as others

frenzymadness:
Because derived images might add some startup actions into scl_enable script, this script should be used in Fedora as well.
    
Fixes: https://github.com/sclorg/s2i-python-container/issues/325

### DIFF
--- a/3.6/Dockerfile.fedora
+++ b/3.6/Dockerfile.fedora
@@ -12,6 +12,13 @@ ENV PYTHON_VERSION=3.6 \
     LANG=en_US.UTF-8 \
     PIP_NO_CACHE_DIR=off
 
+# RHEL7 base images atomatically set these envvars to run scl_enable. RHEl8
+# images, however, don't as most images don't need SCLs any more. But we want
+# to run it even on RHEL8, because we set the virtualenv environment as part of
+# that script
+ENV BASH_ENV=${APP_ROOT}/etc/scl_enable \
+    ENV=${APP_ROOT}/etc/scl_enable \
+    PROMPT_COMMAND=". ${APP_ROOT}/etc/scl_enable"
 
 ENV NAME=python3 \
     VERSION=0 \
@@ -60,12 +67,6 @@ COPY ./root/ /
 RUN virtualenv-$PYTHON_VERSION ${APP_ROOT} && \
     chown -R 1001:0 ${APP_ROOT} && \
     fix-permissions ${APP_ROOT} -P
-
-# For Fedora scl_enable isn't sourced automatically in s2i-core
-# so virtualenv needs to be activated this way
-ENV BASH_ENV="${APP_ROOT}/bin/activate" \
-    ENV="${APP_ROOT}/bin/activate" \
-    PROMPT_COMMAND=". ${APP_ROOT}/bin/activate"
 
 USER 1001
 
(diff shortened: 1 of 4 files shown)


frenzymadness:
[test-openshift]

frenzymadness:
[test]

frenzymadness:
This change partially reverts commit https://github.com/sclorg/s2i-python-container/commit/00fba32c313721c68021d50d021f68d27e3482cb
@pkubatrh @omron93 Could you please take a look at this change and linked issue and tell me what do you think about it?

pkubatrh:
To be honest it seems weird to me to use `scl_enable` in Fedora-based images, considering there were never any SCLs in Fedora in the first place...

Afaik the script was designed only a hack for us to make the SCL enablement inivisible to the user in RHEL7 images, not as a user customisable pre-run script and I do not think either use is even documented anywhere. If a user knows about `scl_enable` he very likely also knows about how it is used and can change the specific vars accordingly.

GrahamDumpleton:
It was bad choice in calling the script `scl_enable` to begin with since it serves a greater role of providing a hook where anything could be done when a shell environment is created. Enabling of SCL packages would have just been one use of it. Maybe a new script with a different name should be created which reflects usage of being a place to do extra shell initialization. The environment variables could refer to the new script and the new script could for now run `scl_enable` in turn. The docs could then mention the new script and its purpose and that extra things can be added to the end, with a notice that use of `scl_enable` is deprecated and will be removed at some point.

FWIW, in images for our training environment we call what I describe as this new script `profile`.

https://github.com/openshift-labs/workshop-terminal/blob/develop/terminal/Dockerfile#L155

It has been made generic and will search out scripts in `profile.d` directory and source them.

https://github.com/openshift-labs/workshop-terminal/blob/develop/terminal/etc/profile

This way anything created from the base image just needs to drop separate files in `profile.d` directory and doesn't have to edit existing files.

https://github.com/openshift-labs/workshop-terminal/tree/develop/terminal/etc/profile.d

So the setup mirrors somewhat how normal shell initialisation works in global `/etc` profiles.

I know that adding hooking mechanisms for customization in official S2I images is frowned upon, but for us in our system it works well.

frenzymadness:
The question here is whether we want to provide something like this in general and if so, do it in the same way for all containers. I think that it would mean inconsiderable amount of work, proper planning, and deprecation process. Not sure we have enough capacity to do it now but we can discuss it at the next meeting.

frenzymadness:
At today meeting, we agreed that:

* RHEL 8 uses the same script even it has modules instead of collections
* therefore, we have to support `scl_enable` script for the lifetime of RHEL 8
* so, it should be okay to enable the same script for Fedora and unify the process of activating the virtual environment
* the name `scl_enable` is not perfect but changing it would have meant a lot of work

If you agree, I can get back to this PR and prepare a working solution.

phracek:
[test]

frenzymadness:
After careful reconsideration, it seems that this no longer makes sense. The scl_enable script is currently used only on RHEL and Centos 7, all other container images (Fedora, RHEL/Centos 8 and 9) don't have it. Also, a way to adjust container images via multistage build exists and it's simple to use and allows users to do anything they might need.

Issue #315, opened 2019-02-15
Make contributing easier


frenzymadness:
I would like to see a simple list of steps with commands I should do before I open a PR.

For example:
* Do your changes to src/ and multispec.yml
* Clean all generated content
* Generate all files from the source
* Check changes
* Build images (or only a few you changed)
* Tag images
* Run tests

And then open PR.

omron93:
I think PRs are more than welcomed.

frenzymadness:
I got it on my TODO list :slightly_smiling_face:

torsava:
+1

PR #329, opened 2019-05-31
Add a step-by-step guide for contributors

frenzymadness:
Fixes https://github.com/sclorg/s2i-python-container/issues/315

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,20 @@ For more information about concepts used in these container images, see the
 
 Contributing
 ---------------
-In this repository [distgen](https://github.com/devexp-db/distgen/) > 1.0 is used for generating directories for Python versions. Also make sure distgen imports the jinja2 package >= 2.10. If you'd like to update some of the files, please make changes in specs/multispec.yml and/or templates under src/ and run `make generate-all`.
+In this repository [distgen](https://github.com/devexp-db/distgen/) > 1.0 is used for generating directories for Python versions. Also make sure distgen imports the jinja2 package >= 2.10.
+
+Files in directories for a specific Python version are generated from templates in the src directory with values from specs/multispec.yml.
+
+A typical way how to contribute is:
+
+1. Add a feature or fix a bug in templates (src directory) or values (specs/multispec.yml file).
+1. Commit the changes.
+1. Regenerate all files via ```make generate-all```.
+1. Commit generated files.
+1. Build an image you want to test via ```make build TARGET=fedora VERSIONS=3.7```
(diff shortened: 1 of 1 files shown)


omron93:
There is https://github.com/sclorg/welcome/blob/master/contribution.md . That was intended to be a common contributing guideline for all sclorg images. But it's quite out-dated and actually never used... so just pointing out.

Otherwise PR LGTM.

README.md
```diff
@@ -16,7 +16,20 @@ For more information about concepts used in these container images, see the
...
+Files in directories for a specific Python version are generated from templates in the src directory with values from specs/multispec.yml.
+
+A typical way how to contribute is:
+
+1. Add a feature or fix a bug in templates (src directory) or values (specs/multispec.yml file).
+1. Commit the changes.
+1. Regenerate all files via ```make generate-all```.
+1. Commit generated files.
+1. Build an image you want to test via ```make build TARGET=fedora VERSIONS=3.7```
+1. Tag it if you want via ```make tag```
```
pkubatrh: This would build and then tag all images for centos (default target). Either use the same make arguments as above, or, even better, only use `make test TARGET=fedora VERSIONS=3.7` which builds, tags and tests the image.

frenzymadness: Fixed

frenzymadness:
See the latest fixup commit.

README.md
```diff
@@ -16,7 +16,18 @@ For more information about concepts used in these container images, see the
...
+Files in directories for a specific Python version are generated from templates in the src directory with values from specs/multispec.yml.
+
+A typical way how to contribute is:
+
+1. Add a feature or fix a bug in templates (src directory) or values (specs/multispec.yml file).
+1. Commit the changes.
+1. Regenerate all files via `make generate-all`.
+1. Commit generated files.
+1. Test changes via `make test TARGET=fedora VERSIONS=3.7` which will `build`, `tag` and `test` an image in one step.
+1. Open a pull request or clean all local changes via `make clean-versions`.
```
pkubatrh: What `make clean-versions` does is basically `rm -rf $VERSIONS` for `VERSIONS` defined in the Makefile. So it makes more sense to run it at start (before `make generate`), to remove all changes that might not be generated from the templates but could still be located in the sources.

frenzymadness: I'm assuming that the new developer will start with a fresh clone so I removed the note about `make clean-versions`.

frenzymadness:
Is it okay now?

pkubatrh:
lgtm now, thanks!

PR #336, opened 2019-07-18
Clean yum cache for every available repository

pkubatrh:


pkubatrh:
Only old version modified, please ignore.

Issue #341, opened 2019-08-02
numpy-test-app test case failing for python-27 image


pkubatrh:
This is due to the fact that the latest upstream version does not support Python 2.7 anymore:
https://docs.scipy.org/doc/numpy/release.html#numpy-1-17-0-release-notes

Test logs: https://gist.github.com/rhscl-bot/80930281f4eb45ea9dc75b6d8a9114db

pkubatrh:
I guess we could pin the numpy version to 1.16 for python 2.7? WDYT @torsava @frenzymadness

frenzymadness:
I think that this is the best solution. Numpy 1.16 will receive all bugfixes till the end of the year. I'll prepare a PR.

PR #342, opened 2019-08-05
Use the version of numpy compatible with a Python version

frenzymadness:
Fixes: https://github.com/sclorg/s2i-python-container/issues/341

### DIFF
--- a/examples/numpy-test-app/requirements.txt
+++ b/examples/numpy-test-app/requirements.txt
@@ -1,2 +1,3 @@
 gunicorn
-numpy
+numpy<1.17.0; python_version < '3.5'
+numpy>=1.17.0; python_version >= '3.5'


frenzymadness:
[test-openshift]

frenzymadness:
That's a strange error. @pkubatrh do you know where the problem may come from?

pkubatrh:
We have been seeing some issues with rhel8 CI due to the way the scripts are called inside the VM and this seems similar enough (missing permissions due to umask?).

@kubco2 might be able to help here, but I would say that the issue is not related to this change and can be fixed separately.

pkubatrh:
LGTM

I verified rhel8 manually outside of our CI and the tests passed. The rest of the tests are green in CI, so mergings. Thanks!

PR #344, opened 2019-09-25
Fix permissions for all files given in s2i

phracek:
Adds fix-permissions function right after mv command.

### DIFF
--- a/src/s2i/bin/assemble
+++ b/src/s2i/bin/assemble
@@ -41,6 +41,9 @@ shopt -s dotglob
 echo "---> Installing application source ..."
 mv /tmp/src/* "$HOME"
 
+# set permissions for any installed artifacts
+fix-permissions /opt/app-root
+
 if [[ ! -z "$UPGRADE_PIP_TO_LATEST"{% if spec.version in ["2.7", "3.4", "3.5"] %} || ! -z "$ENABLE_PIPENV"{% endif %} ]]; then
   echo "---> Upgrading pip to latest version ..."
   pip install -U pip setuptools wheel
--- a/src/s2i/bin/assemble
+++ b/src/s2i/bin/assemble
@@ -42,7 +42,7 @@ echo "---> Installing application source ..."
 mv /tmp/src/* "$HOME"
 
 # set permissions for any installed artifacts
-fix-permissions /opt/app-root
+fix-permissions /opt/app-root -P
 
 if [[ ! -z "$UPGRADE_PIP_TO_LATEST"{% if spec.version in ["2.7", "3.4", "3.5"] %} || ! -z "$ENABLE_PIPENV"{% endif %} ]]; then
   echo "---> Upgrading pip to latest version ..."
@@ -98,4 +98,4 @@ if should_collectstatic; then
 fi
 
 # set permissions for any installed artifacts
-fix-permissions /opt/app-root
+fix-permissions /opt/app-root -P
--- a/2.7/s2i/bin/assemble
+++ b/2.7/s2i/bin/assemble
@@ -39,6 +39,9 @@ shopt -s dotglob
 echo "---> Installing application source ..."
 mv /tmp/src/* "$HOME"
 
+# set permissions for any installed artifacts
+fix-permissions /opt/app-root -P
+
 if [[ ! -z "$UPGRADE_PIP_TO_LATEST" || ! -z "$ENABLE_PIPENV" ]]; then
   echo "---> Upgrading pip to latest version ..."
   pip install -U pip setuptools wheel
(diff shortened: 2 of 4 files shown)


torsava:
[test]

torsava:
[test-openshift]

frenzymadness:
There is an error in rhel7-openshift test but it does not seem to be related to this change. Let's rerun it.

frenzymadness:
[test-openshift]

pkubatrh:
One more time for rhel7 openshift tests as we did not tag the new image that should fix the issue

[test-openshift]

src/s2i/bin/assemble
```diff
@@ -41,6 +41,9 @@ shopt -s dotglob
 echo "---> Installing application source ..."
 mv /tmp/src/* "$HOME"
 
+# set permissions for any installed artifacts
+fix-permissions /opt/app-root
```
pkubatrh: Why `/opt/app-root` instead of `$HOME`? We want to fix the permissions only on newly added files.

phracek: Good catch. At the end is `/opt/app-root`. I will add `fix-permission $HOME` here and at the end too. There is a function like `install_pipenv` which does some operation in $HOME. I will test it locally too, whether all works fine.

pkubatrh: Hmm, the `/opt/app-root` was added a long time ago in https://github.com/sclorg/s2i-python-container/commit/1aae37118cb1ed8183960f9078bcbd354a3cf7bf

I am not sure what "installed artifacts" mean here. Anything installed through pip? Would it even need permission changing?
@torsava @frenzymadness Any idea?

torsava: `$HOME` is set to `/opt/app-root/src`, so `fix-permissions /opt/app-root` should cover `$HOME` too since it's recursive.

I think "installed artifacts" meant mainly files copied into the container. I think pip-installed files should have the correct permissions already.

phracek:
Referenced issue: https://github.com/sclorg/container-common-scripts/issues/119

torsava:
We talked with @phracek offline, and it seems that this is a general issue with s2i containers. It might not be manifesting itself right now in the python container, but it's good to fix it preemptively.

src/s2i/bin/assemble
pkubatrh: @torsava The thing is, if you take a look at the commit I linked there are actually two calls to fix-permissions being added:
1. `fix-permissions ./` which is equivalent to `fix-permissions $HOME` now.
2. `fix-permissions /opt/app-root/` which is called because "set permissions for any installed artifacts"

So when this was added it looks like it was specifically added as a separate case of fixing permissions for files outside of copied in files (since those reside in `$HOME`). Hence my question if it is ok to change.

torsava: @pkubatrh Hmm, in that case, it's indeed likely it was aimed at pip-installed or pipenv-installed files and virtual environments. Maybe preemptively, maybe not.

> Hence my question if it is ok to change.

Wouldn't leaving it as is in this PR (`fix-permissions /opt/app-root`) be safest? It should fix `$HOME` and anything else like pip/env-installed files, etc.

frenzymadness:
All green so let's merge it?

src/s2i/bin/assemble
pkubatrh: >Wouldn't leaving it as is in this PR (fix-permissions /opt/app-root) be safest?

Probably. Though one issue I see with the way it is currently done is that the `fix-permissions` script is following symlinks by default. And since a lot of the files inside `/opt/app-root/lib/python3.6/` is linked to files inside `/opt/rh/rh-python36` we are modifying the permissions of files installed by rpms...

The `fix-permissions` [script](https://github.com/sclorg/s2i-base-container/blob/master/core/root/usr/bin/fix-permissions#L6) has a `SYMLINK_OPT` env variable, that can be set to be empty if following symlinks is not expected so we could use that. WDYT?

pkubatrh: Answering myself, yes we should probably do that:
https://github.com/sclorg/s2i-python-container/pull/245

pkubatrh:
>All green so let's merge it?

The versioned sources need to be generated via `make generate`.

src/s2i/bin/assemble
torsava: Agreed, let's use `-P` not to follow symlinks.

phracek: It looks good. I think both `fix-permissions` scripts in `assemble` script have to use `-P`. 
Thanks to both.

pkubatrh:
[test][test-openshift]

pkubatrh:
LGTM now, thanks!

PR #346, opened 2019-09-27
Add missing packages for RHEL-8 builds

hhorak:
During rebuild of 8.1.0 containers, there have been two issues:

1) There was an issue with setting locales, the test log was full of those warnings:
```
/bin/bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
```
I admin I don't understand the whole story here, but saw installing `glibc-langpack-en` in `ubi8/python-27` image, and the same works for `ubi8/pyton-36` image. So, this is a suggested fix for this issue as well.

2) During `mod-wsgi-test-app` test, we saw this issue:
```
gcc: error: /usr/lib/rpm/redhat/redhat-hardened-cc1: No such file or directory
```
Since `/usr/lib/rpm/redhat/redhat-hardened-cc1` is shipped by `redhat-rpm-config` on RHEL-8.1, adding this package helped. Again, it is not clear to me what changed since RHEL-8.0 or why it works fine on other distros.

### DIFF
--- a/2.7/Dockerfile.rhel8
+++ b/2.7/Dockerfile.rhel8
@@ -41,10 +41,10 @@ LABEL summary="$SUMMARY" \
       usage="s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=2.7/test/setup-test-app/ ubi8/python-27 python-sample-app" \
       maintainer="SoftwareCollections.org <sclorg@redhat.com>"
 
-RUN INSTALL_PKGS="python2 python2-devel glibc-langpack-en python2-setuptools python2-pip \
-        python2-virtualenv nss_wrapper httpd httpd-devel mod_ssl \
-        mod_auth_gssapi mod_ldap mod_session atlas-devel gcc-gfortran \
-        libffi-devel libtool-ltdl enchant" && \
+RUN INSTALL_PKGS="python2 python2-devel glibc-langpack-en redhat-rpm-config python2-setuptools \
+        python2-pip python2-virtualenv nss_wrapper httpd httpd-devel \
+        mod_ssl mod_auth_gssapi mod_ldap mod_session atlas-devel \
+        gcc-gfortran libffi-devel libtool-ltdl enchant" && \
     yum -y module enable python27:2.7 httpd:2.4 && \
     yum -y --setopt=tsflags=nodocs install $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
--- a/3.6/Dockerfile.rhel8
+++ b/3.6/Dockerfile.rhel8
@@ -41,10 +41,10 @@ LABEL summary="$SUMMARY" \
       usage="s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=3.6/test/setup-test-app/ ubi8/python-36 python-sample-app" \
       maintainer="SoftwareCollections.org <sclorg@redhat.com>"
 
-RUN INSTALL_PKGS="python36 python36-devel glibc-langpack-en python3-setuptools python3-pip \
-        python3-virtualenv nss_wrapper httpd httpd-devel mod_ssl \
-        mod_auth_gssapi mod_ldap mod_session atlas-devel gcc-gfortran \
-        libffi-devel libtool-ltdl enchant" && \
+RUN INSTALL_PKGS="python36 python36-devel glibc-langpack-en redhat-rpm-config python3-setuptools \
+        python3-pip python3-virtualenv nss_wrapper httpd httpd-devel \
+        mod_ssl mod_auth_gssapi mod_ldap mod_session atlas-devel \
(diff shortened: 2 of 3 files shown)


frenzymadness:
Looks good to me and the build failure on rhel8 does not seem to be related to these changes.

frenzymadness:
@torsava do you want to take a look before I merge this?

torsava:
Interesting, also no idea why these packages aren't pulled in anymore. Since they are needed now in both Python versions, I would move them to `base_pkgs`, but we can do that later when adding a new Python version.

LGTM!

pkubatrh:
I wonder... Since we are installing `glibc-all-langpacks` in [s2i-core](https://github.com/sclorg/s2i-base-container/blob/master/core/Dockerfile.rhel8#L45) I would expect any further installs of specific glibc langpacks would be redundant. But looking inside the image with `glibc-all-langpacks` installed there seem to be no locales available? Do I perhaps misunderstand the use of the rpm?

pkubatrh:
Answering myself - my understanding is correct but it does not really reflect the reality.

`glibc-all-langpacks` only installs locales defined by the `%_install_langs` rpm macro during its posttrans scriptlet. ubi 8.0 image has the macro defined as `en_US` (which is why it worked in the past), ubi 8.1 image changed it to `C.utf8` (which is why it does not work anymore), and it seems to be `all` outside of container images.

imo it would make more sense to move the change to s2i-core instead, since installing `glibc-all-langpacks` is basically useless now (`C.utf8` is already present in the base image).

hhorak:
Agreement on the meeting: `glibc-langpack-en` into s2i-core (instead of `glibc-all-langpacks`) and `redhat-rpm-config` should be installed in s2i-base.

phracek:
`s2i-base` image contains `redhat-rpm-config` package and `s2i-core` image containes `glibc-langpack-en instead of `glibc-all-langpacks`.

Removing `[DO NOT MERGE]` from title

phracek:
@torsava Please remove packages `redhat-rpm-config` and `glibc-langpack-en` from PR. They are already part of s2i-{base,core} images.

2.7/Dockerfile.rhel8
```diff
@@ -41,10 +41,10 @@ LABEL summary="$SUMMARY" \
       usage="s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=2.7/test/setup-test-app/ ubi8/python-27 python-sample-app" \
       maintainer="SoftwareCollections.org <sclorg@redhat.com>"
 
-RUN INSTALL_PKGS="python2 python2-devel glibc-langpack-en python2-setuptools python2-pip \
-        python2-virtualenv nss_wrapper httpd httpd-devel mod_ssl \
-        mod_auth_gssapi mod_ldap mod_session atlas-devel gcc-gfortran \
-        libffi-devel libtool-ltdl enchant" && \
+RUN INSTALL_PKGS="python2 python2-devel glibc-langpack-en redhat-rpm-config python2-setuptools \
```
phracek: Remove redhat-rpm-config

pkubatrh:
Closing as this got fixed by https://github.com/sclorg/s2i-base-container/pull/195

3.6/Dockerfile.rhel8
```diff
@@ -41,10 +41,10 @@ LABEL summary="$SUMMARY" \
       usage="s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=3.6/test/setup-test-app/ ubi8/python-36 python-sample-app" \
       maintainer="SoftwareCollections.org <sclorg@redhat.com>"
 
-RUN INSTALL_PKGS="python36 python36-devel python3-setuptools python3-pip python3-virtualenv \
-        nss_wrapper httpd httpd-devel mod_ssl mod_auth_gssapi \
-        mod_ldap mod_session atlas-devel gcc-gfortran libffi-devel \
-        libtool-ltdl enchant" && \
+RUN INSTALL_PKGS="python36 python36-devel glibc-langpack-en redhat-rpm-config python3-setuptools \
```
phracek: The same here. Remove `redhat-rpm-config` and `glibc-langpack-en`

specs/multispec.yml
```diff
@@ -54,8 +54,8 @@ specs:
                   'mod_auth_gssapi', 'mod_ldap', 'mod_session',
                   'atlas-devel', 'gcc-gfortran', 'libffi-devel', 'libtool-ltdl enchant']
       extra_pkgs:
-        "2.7": ['{{ spec.pkg_prefix }}', '{{ spec.pkg_prefix }}-devel', 'glibc-langpack-en']
-        "3.6": ['python36', 'python36-devel']
+        "2.7": ['{{ spec.pkg_prefix }}', '{{ spec.pkg_prefix }}-devel', 'glibc-langpack-en', 'redhat-rpm-config']
```
phracek: remove it as above.

PR #351, opened 2019-10-31
Update requests in pipenv-test-app to 2.20.0

frenzymadness:


2.7/test/pipenv-test-app/Pipfile.lock
```diff
@@ -49,47 +36,171 @@
...
+            ],
+            "version": "==0.1.7"
+        },
+        "zipp": {
+            "hashes": [
+                "sha256:3718b1cbcd963c7d4c5511a8240812904164b7f381b647143a89d3b98f9bcd8e",
+                "sha256:f06903e9f1f43b12d371004b4ac7b06ab39a44adc747266928ae6debfa7b3335"
             ],
-            "version": "==3.2.3"
+            "version": "==0.6.0"
```
phracek: Vie the version is so dramatically changed?

frenzymadness: I am not sure what happened but zip version 0.6.0 is the latest version available on [PyPI](https://pypi.org/project/zipp/#history) and [GitHub](https://github.com/jaraco/zipp/releases). I'll try to investigate it.

frenzymadness: I got it. It's not a different version of Zipp because Zipp was not in this file previously so it's only the diff what is confusing. Pytest was the latest package in this file so 3.2.3 is the previous pytest version.

PR #352, opened 2019-11-04
Clean & update

frenzymadness:
Clean unsupported versions from config files and update Python 3.7 image to Fedora 31.


frenzymadness:
Fedora 31 base/core images are not ready yet. https://github.com/sclorg/s2i-base-container/issues/196

PR #353, opened 2019-11-11
gunicorn does not support python 2.7 since version 20.0.0

phracek:
gunicorn does not support python 2.7 since version 20.0.0.
See http://docs.gunicorn.org/en/stable/install.html
Update requirements.txt files for examples userd in s2i-python-container.
Requirements.txt were updated so that for Python2.7 is taken version less then 20.0.0
Otherwise, the latest version is used.

Add requirements.txt which are missing in `setup-test-app`.


phracek:
@torsava Can you please look at this pull request which fixes support python2.7 for gunicorn case. Thanks

frenzymadness:
The change itself looks good to me.

The test failure in rehl8 is caused by this assert: `assert locale.getpreferredencoding() == 'UTF-8'`.

```
sys.stdout.encoding == "UTF-8"
locale.getlocale() == (None, None)
locale.getdefaultlocale() == ('en_US', 'UTF-8')

but

locale.getpreferredencoding() == "ANSI_X3.4-1968"
```
The log is full of messages like this:
```
bin/sh: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
```
AFAIK this should be already fixed by installation of proper langpack package. Can we just ignore this error and wait for an update of the base image?

pkubatrh:
Should be fixed by https://github.com/sclorg/s2i-base-container/pull/195

It is only a matter of time before it gets to rhel8/s2i-base so I am ok with waiving the error.

PR #354, opened 2019-11-17
Update README.md (3.6) – note regarding npm

isbur:
reference to base image where npm is actually included


frenzymadness:
Looks good to me but why is this only in 3.6/readme instead of src/readme and regenerated for all?

phracek:
@isbur Can you please update this PR as we suggest?

frenzymadness:
Let me know if you need any help. If you don't want to finish this, I can do it.

frenzymadness:
Replaced by #364

PR #355, opened 2019-12-03
Clean & update

frenzymadness:
After I prepared s2i-[core,base] for Fedora 3[01], here is an update of s2i-python container.

The one big change here is that we have to drop Python 3.6 Fedora image because no currently supported Fedora version has Python 3.6 as a default. Is there anything else I have to do to stop supporting this image?

With this change, we have everything on Fedora 31 which will be supported for a long time and my plan is to prepare Fedora 32 with the Python 3.8.


frenzymadness:
[test]

frenzymadness:
[test-openshift]

frenzymadness:
RHEL8 error does not seem to be related to these changes. Does anybody know what happened there?

pkubatrh:
>RHEL8 error does not seem to be related to these changes. Does anybody know what happened there?

Try updating the common submodule (`git submodule update --remote`), the version used in `master` is quite old so the cause might be some patch for podman that did not get picked up.

frenzymadness:
[test]

frenzymadness:
[test-openshift]

frenzymadness:
There seems to be some problem with CI: `ssh: Could not resolve hostname .ci.centos.org: Name or service not known`

frenzymadness:
[test]

frenzymadness:
[test-openshift]

frenzymadness:
Now, it looks good and the rhel8 failure is caused by the well-known problem which is already fixed and we are just waiting for update of base images.

frenzymadness:
Can anybody do a review please?

2.7/s2i/bin/assemble
```diff
@@ -11,7 +11,7 @@ function should_collectstatic() {
 function virtualenv_bin() {
   if head "/etc/redhat-release" | grep -q "^CentOS Linux release 7" || \
      head "/etc/redhat-release" | grep -q "^Red Hat Enterprise Linux\( Server\)\? release 7" || \
-     head "/etc/redhat-release" | grep -q "^Fedora release 29 \(Twenty Nine\)\?"; then
+     head "/etc/redhat-release" | grep -q "^Fedora release 31 \(Thirty One\)\?"; then
```
pkubatrh: Can we leave just `grep -q "^Fedora release";` or are there still some non-EOL Fedora releases that use `virtualenv-${PYTHON_VERSION}`?

pkubatrh:
Otherwise lgtm

2.7/s2i/bin/assemble
frenzymadness: You are right. We have only one version of Fedora now.

frenzymadness:
[test]

frenzymadness:
[test-openshift]

PR #356, opened 2019-12-11
Fedora 32 and Python 3.8

frenzymadness:
SSIA

Let's see what CI tells me.


frenzymadness:
It's ready for a review.

phracek:
I would also add `.exclude-rhel7`, `.exclude-centos7` and `.exclude-rhel8` files like it is mentioned here https://github.com/sclorg/nginx-container/pull/97/files

This means, that `3.8` version will not be built on RHEL and CentOS.

frenzymadness:
Ok, one by one:

* I have no idea why distgen fails on rhel8. I cannot reproduce the issue and I also cannot find any mention of a configuration file named like `fedora-32-x86_64.yaml` or similar for other distros.
* Python 3.8 is mentioned in the main readme. Just mentioned because it's available only for Fedora and I think that the examples in the main readme should contain Python 3.6 (LTS version)
* Yes, `3.8/Readme` contains links to RHEL and Centos repositories and we do not provide these combinations yet. There is no way how to fix it in distgen because the readme is generated only once without any info from specfile matrix. I have an idea how to do it but it would be better to discuss it at a meeting and fix it later. We provide the same inaccurate info also for other Python versions (3.7).
* `.exclude-*` files don't make sense here because Dockerfiles for RHEL and Centos are not generated.

Could you please help me with the RHEL issue?

frenzymadness:
I cannot find where the distgen comes from. It isn't available as a RPM package on RHEL 8. Any idea?

pkubatrh:
Installed on internal virtual machines from fc28 packages. We should move to epel8 once https://github.com/devexp-db/distgen/pull/101 is done.

frenzymadness:
@pkubatrh distgen seems to be ready in epel8:
* build https://koji.fedoraproject.org/koji/buildinfo?buildID=1428234
* update https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2020-15ebcb589e

Could you please try to update our virtual machines?

phracek:
[test]

hhorak:
[test]

hhorak:
@pkubatrh @frenzymadness I've built a new rhel8 image as promised, it includes the latest `distgen`, so should hopefully make these tests succeed.

hhorak:
I see it fails but during a test, so hopefully not because of the distgen anymore.

pkubatrh:
>ERROR: Failed to upgrade 'setuptools' to the latest version.

So I guess a rebase should fix it

frenzymadness:
[test]

frenzymadness:
GPG keys after branching problem yet again. We have to wait for a while.

pkubatrh:
>GPG keys after branching problem yet again. We have to wait for a while.

I expect this will actually need a rebuild for s2i-core/s2i-base right? Created tickets to ask for f32 branches for now.

frenzymadness:
F32 is branched so we can create branches for all images. Next time, I'll definitely start with this after branching. Have you requested a new branch for python container or just for base and core containers?

frenzymadness:
[test]

pkubatrh:
>Have you requested a new branch for python container or just for base and core containers?

Only for core and base.

frenzymadness:
[test]

frenzymadness:
[test]

frenzymadness:
It seems that after a rebuild of s2i-core and base we are finally ready with Fedora 32 and Python 3.8.

@phracek Could you plase do a review again or take a look on my comment from December ([this one](https://github.com/sclorg/s2i-python-container/pull/356#issuecomment-567384520))?

frenzymadness:
[test]

torsava:
I did a review and it looks good!

The only problem I see is with the README. Maybe we can create some new variables in the multispec to turn those examples off? Or maybe we just merge it, since CentOS is coming soon anyway?

What do you think, @frenzymadness ?

torsava:
Oh, and there's a forgotten number `3.6` in `README.md` and in `root/help.1`.

frenzymadness:
RHEL 7 CI failure is not related to this PR. I'm gonna fix the readme issue in a new pull request.

PR #357, opened 2019-12-16
add ubi8 labels to rhel8 versions of python images

pkubatrh:


pkubatrh:
Thanks for review!. Tests are green except for rhel8 where the failure is known, merging.

PR #359, opened 2020-01-15
Add imagestream file for s390x and remove v3.5 that is EOL

hhorak:
Similar PR to sclorg/mariadb-container#115

### DIFF
--- a/imagestreams/python-centos7.json
+++ b/imagestreams/python-centos7.json
@@ -49,26 +49,6 @@
         }
       },
       {
-        "name": "3.5",
-        "annotations": {
-          "openshift.io/display-name": "Python 3.5",
-          "openshift.io/provider-display-name": "Red Hat, Inc.",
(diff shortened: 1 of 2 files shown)


frenzymadness:
Your patch looks good but the CI fails again on the completely unrelated issue.

The problem is that the latest version of setuptools does not support Python 2.7 and therefore the latest version cannot be installed during the test with Python 2.7. https://github.com/pypa/setuptools/blob/master/CHANGES.rst#v4500  https://github.com/sclorg/s2i-python-container/blob/master/examples/npm-virtualenv-uwsgi-test-app/app.sh#L31

pkubatrh:
LGTM as well. No need to hold this because of a CI failure since the imagestreams are not tested in any way.

@frenzymadness Can you follow up with the fix for the failure (pin some older release of setuptools?) in a different PR?

frenzymadness:
Follow up issue for discussion: https://github.com/sclorg/s2i-python-container/issues/360

Issue #360, opened 2020-01-16
Testing of a library version with Python 2.7


frenzymadness:
Test [npm-virtualenv-uwsgi-test-app](https://github.com/sclorg/s2i-python-container/blob/master/examples/npm-virtualenv-uwsgi-test-app/app.sh#L31) tests that we have the latest version of libraries installed in a virtual environment.

This won't work anymore because Python 2.7 doesn't get the latest versions due to its EOL. For example, the latest setuptools is 45.0.0 but the Python 2.7 support ends on 44.0.0. https://github.com/pypa/setuptools/blob/master/CHANGES.rst#v4500

It is possible to get the version compatibility info from the PyPI API but it won't be easy to implement:

setuptools 45
```
"45.0.0": [
…
        "requires_python": ">=3.5",
```

setuptools 44
```
    "44.0.0": [
…
        "requires_python": "!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,>=2.7",
```
I'd say that better would be to remove this part of the test.

andyfeller:
Will second having an issue with `UPGRADE_PIP_TO_LATEST` upgrading to setuptools>=45.0.0 breaks Python 2.7 usage and something to fix it would be appreciated.

frenzymadness:
I might know where the problem comes from and I try to reproduce it next week. It seems that the version of pip/setuptools installed in the container is too old to support the "supported pythons" feature which means that the setuptools/pip stack is updated to the latest version of setuptools which does not support Python 2 but this fact is not known before the update.

frenzymadness:
The solution here might be to do an update in two steps:
1. Update to the latest version supported by the Python 2.
2. Update to the latest version. This step would be performed only for Python 3 because the newer setuptools/pip stack would know that there is no newer version with Python 2 support available.

andyfeller:
@frenzymadness TIL!  Thanks for the theory there; definitely need to bone up on that feature as I saw situations where this was and wasn’t an issue but didn’t know where to start looking

PR #361, opened 2020-01-28
Too old pip for Python 2.7 support in packages and for manylinux2014 support

frenzymadness:
My attempt to deal with setuptools without support for Python 2.7. Fixes: https://github.com/sclorg/s2i-python-container/issues/360

### DIFF
--- a/examples/npm-virtualenv-uwsgi-test-app/app.sh
+++ b/examples/npm-virtualenv-uwsgi-test-app/app.sh
@@ -2,23 +2,6 @@
 
 # Test virtualenv environment and pip upgrade
 
-# Get the latest stable version of the package released on PyPI
-function get_latest_stable_version() {
-  echo $(curl -sL 'https://pypi.org/pypi/'$1'/json' | python -c \
-"""
-import sys
-import json
-from pip._vendor.packaging.version import parse
-versions = [parse(v) for v in json.load(sys.stdin)['releases'].keys()]
-print(str((sorted([v for v in versions if not v.is_prerelease])[-1])))
-""")
-}
-
-# Get version of the package installed on the system
-function get_installed_version() {
-  echo $(pip freeze --all | grep $1 | cut -d"=" -f3)
-}
-
 echo "Testing that the virtual environment's Python is being used ..."
 if [ "$(which python)" != "/opt/app-root/bin/python" ]; then
     echo "ERROR: Initialization of the virtual environment failed."
@@ -28,7 +11,10 @@ fi
 echo "Testing UPGRADE_PIP_TO_LATEST=1 (set in .s2i/environment) ..."
 packages=("pip" "setuptools" "wheel")
 for pkg in ${packages[@]}; do
-  if [ $(get_latest_stable_version $pkg) != $(get_installed_version $pkg) ]; then
+  # grep returns exit code 1 if the output contains only one line starting
+  # with "Requirement already …" which means that the package is updated
+  python -m pip install -U --no-python-version-warning $pkg 2>&1 | grep -v "^Requirement already up-to-date: "
+  if [ $? -eq 0 ]; then
     echo "ERROR: Failed to upgrade '$pkg' to the latest version."
     exit 1
   fi
--- a/2.7/s2i/bin/assemble
+++ b/2.7/s2i/bin/assemble
@@ -42,6 +42,13 @@ mv /tmp/src/* "$HOME"
 # set permissions for any installed artifacts
 fix-permissions /opt/app-root -P
 
+# We have to first upgrade pip to at least 19.3 because:
+# * pip < 9 does not support different packages' versions for Python 2/3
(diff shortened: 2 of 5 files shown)


frenzymadness:
We need a more general solution for this because a too old version of pip is also troublemaker for wheels. In short, architectures like ppc64le or aarch64 are supported only in the lastest standard manylinux2014. For example, numpy produces these new wheels but pip<19.3 cannot work with them. And python36 container contains Python 3.6.8 with pip 9.0.3 which causes pip to build and install numpy from sources. Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1791833

So we need to update pip everywhere to at least 19.3. The question here is whether it would be better to make `UPGRADE_PIP_TO_LATEST=1` the default or pin pip to one specific version we need and keep `UPGRADE_PIP_TO_LATEST` as it is.

frenzymadness:
@torsava this is the solution we were talking about.
@phracek do you want to take a look?

frenzymadness:
[test-openshift]

examples/npm-virtualenv-uwsgi-test-app/app.sh
```diff
@@ -28,7 +11,10 @@ fi
 echo "Testing UPGRADE_PIP_TO_LATEST=1 (set in .s2i/environment) ..."
 packages=("pip" "setuptools" "wheel")
 for pkg in ${packages[@]}; do
-  if [ $(get_latest_stable_version $pkg) != $(get_installed_version $pkg) ]; then
+  # grep returns exit code 1 if the output contains only one line starting
+  # with "Requirement already …" which means that the package is updated
+  python -m pip install -U --no-python-version-warning $pkg | grep -v "^Requirement already up-to-date: "
+  if [ $? -eq 0 ]; then
```
torsava: Have you tested what happens if the update fails for any other reason?

frenzymadness: This step should not update anything. If the command will have anything else in the output and the grep won't fail, it means that:
1. there is a problem checking the versions from PyPI
2. the installed packages are not the latest versions

torsava: I don't understand why it won't update anything?

frenzymadness: This command is here only to check that all packages in the virtual environment are up to date. The packages are actually updated in the `$UPGRADE_PIP_TO_LATEST` section in the assemble script. This is only a test. If all the packages are correctly updated by the pip command in the assemble script, this command does nothing and the test pass. Otherwise, this command does an update which is wrong and the test fails.

torsava: Aha! So this test is run only after pip has already been upgraded to latest? Ok that makes sense.

So we have 3 cases for the lines 16 and 17:
1. command works and grep fails => the if doesn't succeed - GOOD
2. command works and grep succeeds => if does succeed and reports error - GOOD
3. command fails => if does not succeed and does not report error - BAD?

Or is there something I'm missing with case 3?

frenzymadness: > Aha! So this test is run only after pip has already been upgraded to latest? Ok that makes sense.

pip, setuptools, wheel, yes

> 
> So we have 3 cases for the lines 16 and 17:
> 
>     1. command works and grep fails => the if doesn't succeed - GOOD
>
>     2. command works and grep succeeds => if does succeed and reports error - GOOD
> 
>     3. command fails => if does not succeed and does not report error - BAD?
> 
> 
> Or is there something I'm missing with case 3?

If pip fails and produces anything unexpected on stdout, grep does not fail (because the output is not empty) and the failure is reported.

```
$ pip install six --index-url https://example.com  | grep -v "^Requirement already up-to-date: "
Looking in indexes: https://example.com
ERROR: Could not install packages due to an EnvironmentError: ('Received response with content-encoding: gzip, but failed to decode it.', error('Error -3 while decompressing data: incorrect header check',))
$ echo $?
0
```

We can make it better for cases when pip produces something only to stderr by redirecting stderr to stdout so the test will fail in more cases. Is this what you meant?

torsava: Ah, I thought a failed command does not continue with a pipe. Maybe that's a setting in bash?

In any case, thanks for the explanation. In that case it looks good to me!

frenzymadness: I've added the mentioned redirection of stderr to stdout to make this more bulletproof.

frenzymadness:
[test-openshift]

frenzymadness:
[test]

frenzymadness:
[test-openshift]

frenzymadness:
I cannot manually reproduce the issue we see in Fedora even with older container images. I'll try it again if it will fail in the future.

PR #362, opened 2020-02-11
Fix example app name in 3.7 README

Glutexo:
The README for Python 3.7 still uses _python3.6_ as an example OpenShift app name. Renamed to 3.7.


frenzymadness:
I am sorry but I have to revert this change. The content of the directories (2.7, 3.6…) is generated via `distgen` from the `src` directory. I'll prepare a proper PR for the fix so you don't have to learn the process for the one small typo.

PR #363, opened 2020-02-11
Revert "Fix example app name in 3.7 README"

frenzymadness:
Reverts sclorg/s2i-python-container#362

The content is generated via distgen so this fix has to happen in `src` directory.


frenzymadness:
Cc @Glutexo @phracek

PR #370, opened 2020-03-13
Allow to run only one test separately by specifying as an argument

hhorak:
This is just a small enhancement to make the debugging faster.

### DIFF
--- a/test/run
+++ b/test/run
@@ -198,6 +198,9 @@ check_result $?
 test_docker_run_usage
 check_result $?
 
+# For debugging purposes, this script can be run with one or more arguments
+# those arguments list is a sub-set of values in the WEB_APPS array defined above
+# Example: ./run app-home-test-app pipenv-test-app
 for app in ${@:-${WEB_APPS[@]}}; do
   prepare ${app}
   run_s2i_build ${app}
--- a/test/run
+++ b/test/run
@@ -198,7 +198,7 @@ check_result $?
 test_docker_run_usage
 check_result $?
 
-for app in ${WEB_APPS[@]}; do
+for app in ${@:-${WEB_APPS[@]}}; do
   prepare ${app}
   run_s2i_build ${app}
   RESULT=$?


frenzymadness:
Well, it looks simple but I had to take a proper look at this :) It looks good to me and makes sense. Have you tried it with optional arguments? I guess that the standard case without an optional argument is tested by our CI.

hhorak:
> Have you tried it with optional arguments?

I'm not sure I follow -- which arguments do you mean?

frenzymadness:
> > Have you tried it with optional arguments?
> 
> I'm not sure I follow -- which arguments do you mean?

When you try it with only one test so only the test you give to the script (as an argument) is executed..

frenzymadness:
Could you please note how can I use this new possibility?

phracek:
[test][test-openshift]

phracek:
I have tried to find out explanation about `${@:-${WEB_APPS[@]}}` what `@:-` should do, but no chance.

The output looks the same as before. I would like to understand this improvement.

My test was:
```bash
$ declare -a WEB_APPS=({test1,test2-foo,test3-bar}-webn)
$ for u in ${@:-${WEB_APPS[@]}}; do echo ${u}; done
test1-webn
test2-foo-webn
test3-bar-webn
$
```

pkubatrh:
>I would like to understand this improvement.

`$@` is a list of quoted arguments to the script, `${@:-$value}` checks if any arguments are provided and uses `$value` as the default if there are none. In this case `${WEB_APPS[@]}` is used which is a list of test cases

phracek:
I see, then if `test/run` is executed with the arguments, then they are used for testing the app
otherwise, the default set is used.

phracek:
I would be glad for any kind of note, how can it be used in real time?

hhorak:
I've rebased this for newest master and added a comment how to use this. Let me know if it's not clear.

phracek:
@frenzymadness What do you think now?

PR #372, opened 2020-03-16
Very first try to add micropipenv to the images

frenzymadness:
It should be optional and because it can handle all types of requirements specifications we currently have, we can run all our tests with it.

Let's see what CI thinks about this.


frenzymadness:
[test]

frenzymadness:
[test]

frenzymadness:
This is a mess. Don't even look at this. But I am on a good way.

frenzymadness:
[test]

frenzymadness:
[test]

frenzymadness:
Tests are passing. Now, it's time for squashing :)

frenzymadness:
[test-openshift]

2.7/s2i/bin/assemble
```diff
@@ -18,23 +18,31 @@ function virtualenv_bin() {
...
 }
 
 set -e
 
+# First of all, check that we don't have disallowed combination of ENVs
+if [[ ! -z "$ENABLE_PIPENV" && ! -z "$ENABLE_MICROPIPENV" ]]; then
+  echo "ERROR: Pipenv and micropipenv cannot be enabled at the same time!"
+  # podman/buildah does not relay this exit code but it will be fixed hopefuly
+  # https://github.com/containers/buildah/issues/2305
+  exit 3
```
phracek, line 43: Do we really want exit 3? I guess, enough is exit 1.

frenzymadness: The plan here is that when [this change in buildah](https://github.com/containers/buildah/pull/2306) will be released, we will be able to do a better check for the specific exit code which is not possible now.
In other words, I am not able to check that the unsupported combination of micropipenv and pipenv is the issue why the build failed but with a specific exit code and a newer buildah, I'll be able to do it in the future.
WDYT?

frenzymadness:
> Any reason why do you removed symlinks to `../../test/run` for all images.
> 
> I dont' follow this change.

Yes, the reason is that `test/run` files are no longer a symlink but they are generated via distgen now.

And thank you for the review!

test/run
```diff
@@ -192,7 +201,14 @@ check_result $?
 for app in ${WEB_APPS[@]}; do
   prepare ${app}
   run_s2i_build ${app}
-  check_result $?
+  RESULT=$?
+  if [[ "$app" == "pipenv-and-micropipenv-fails-test-app" ]]; then
```
pkubatrh, line 205: I wonder, how about somehow adding the expected result to the test name instead (eg `pipenv-and-micropipenv-should-fail-app`)? Would make it easier for future expected failure test cases to be added.

src/README.md
```diff
@@ -127,6 +127,14 @@ file inside your source code repository.
     and Pipfile.lock.{% if spec.version in ["2.7"] %} (Implies `UPGRADE_PIP_TO_LATEST` to satisfy dependencies of
     Pipenv.){% endif %}
 
+{% if spec.version.startswith("3.") %}
+* **ENABLE_MICROPIPENV**
+
+    Set this variable to use [micropipenv](https://github.com/thoth-station/micropipenv),
+    a lightweight wrapper for pip to support requirements.txt, Pipenv and Poetry lock
+    files or converting them to pip-tools compatible output. Designed for containerized Python applications.
+    Available only for Python 3 images.
```
pkubatrh, line 136: Do we need to expliicitly mention this if the section is only generated for python 3 images?

2.7/s2i/bin/assemble
torsava: Sounds good! Maybe I would choose a higher code so there's less chance of something else throwing code 3, but that's probably too paranoid :)

src/README.md
torsava: Some people might read e.g. python36 README and then expect it to work on python27, I would say it's worth keeping for clarity.

manifest.sh
```diff
@@ -109,4 +122,16 @@ COPY_RULES="
...
+
+    src=examples/micropipenv-test-app/setup.py
+    dest=test/micropipenv-test-app/setup.py;
+
+    src=examples/micropipenv-test-app/.s2i/environment
+    dest=test/micropipenv-test-app/.s2i/environment;
+
+    src=examples/micropipenv-test-app/.gitignore
+    dest=test/micropipenv-test-app/.gitignore;
 "
```
torsava, line 137: Does `manifest.sh` contain all the new files? I might be looking at it wrong, but for example I don't see the new file `examples/micropipenv-test-app/.s2i/environment`, and there's probably others.

torsava:
I left some comments, besides that it looks good!

src/README.md
frenzymadness: I agree with Tomáš. I'd keep it there.

test/run
frenzymadness: Good idea, I'm gonna implement it.

manifest.sh
frenzymadness: The file you mentioned `examples/micropipenv-test-app/.s2i/environment` is in the code snippet you commented on line 133 :confused: 

Nonetheless, missing files or some other errors in manifest would cause problems with distgen or non-working tests. Don't forget that some files are copied, some are symlinks and some are generated via distgen, so they appear in different parts of manifest.

frenzymadness:
@pkubatrh and @phracek could you please check it again? Hope I addresses your feedbacks and clarified some changes.

phracek:
[test-openshift]

phracek:
[test]

PR #373, opened 2020-03-30
Readme cleanups

frenzymadness:
The purpose of this change is to make the readme files more general and lower the possibility of a mistake when we support many versions of Python but some of them only on specific platforms. Also, it's not necessary to show multiple examples when the only difference is in one word. Another thing is that we copy those files to downstream repositories and it does not make really sense to have RHEL examples in Fedora repository.

Cc @torsava

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -32,7 +32,7 @@ the nodejs itself is included just to make the npm work.
 Usage
 ---------------------
 
-For this, we will assume that you are using the `rhscl/python-27-rhel7 image`, available via `python:2.7` imagestream tag in Openshift.
+For this, we will assume that you are using the supported image, available via `python:2.7` imagestream tag in Openshift.
 Building a simple [python-sample-app](https://github.com/sclorg/s2i-python-container/tree/master/2.7/test/setup-test-app) application
 in Openshift can be achieved with the following step:
 
@@ -43,7 +43,7 @@ in Openshift can be achieved with the following step:
 The same application can also be built using the standalone [S2I](https://github.com/openshift/source-to-image) application on systems that have it available:
 
     ```
-    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=2.7/test/setup-test-app/ rhscl/python-27-rhel7 python-sample-app
+    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=2.7/test/setup-test-app/ <image_name> python-sample-app
     ```
 
 **Accessing the application:**
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -46,6 +46,8 @@ The same application can also be built using the standalone [S2I](https://github
     $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=2.7/test/setup-test-app/ <image_name> python-sample-app
     ```
 
+Where `<image_name>` is the s2i-python image you [downloaded from RHEL, Centos or Fedora registry](../README.md#Download) or [built](../README.md#Build) from these sources. For example ubi8/python-36, centos/python-36-centos7 or f31/python3.
+
 **Accessing the application:**
 ```
 $ curl 127.0.0.1:8080
--- a/3.6/README.md
+++ b/3.6/README.md
@@ -46,6 +46,8 @@ The same application can also be built using the standalone [S2I](https://github
     $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=3.6/test/setup-test-app/ <image_name> python-sample-app
     ```
 
+Where `<image_name>` is the s2i-python image you [downloaded from RHEL, Centos or Fedora registry](../README.md#Download) or [built](../README.md#Build) from these sources. For example ubi8/python-36, centos/python-36-centos7 or f31/python3.
+
 **Accessing the application:**
 ```
 $ curl 127.0.0.1:8080
--- a/3.6/README.md
+++ b/3.6/README.md
@@ -32,7 +32,7 @@ the nodejs itself is included just to make the npm work.
 Usage
 ---------------------
 
-For this, we will assume that you are using the `rhscl/python-36-rhel7 image`, available via `python:3.6` imagestream tag in Openshift.
+For this, we will assume that you are using the supported image, available via `python:3.6` imagestream tag in Openshift.
 Building a simple [python-sample-app](https://github.com/sclorg/s2i-python-container/tree/master/3.6/test/setup-test-app) application
 in Openshift can be achieved with the following step:
 
@@ -43,7 +43,7 @@ in Openshift can be achieved with the following step:
 The same application can also be built using the standalone [S2I](https://github.com/openshift/source-to-image) application on systems that have it available:
 
     ```
-    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=3.6/test/setup-test-app/ rhscl/python-36-rhel7 python-sample-app
+    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=3.6/test/setup-test-app/ <image_name> python-sample-app
     ```
 
 **Accessing the application:**
--- a/3.7/README.md
+++ b/3.7/README.md
@@ -32,7 +32,7 @@ the nodejs itself is included just to make the npm work.
 Usage
 ---------------------
 
-For this, we will assume that you are using the `rhscl/python-37-rhel7 image`, available via `python:3.7` imagestream tag in Openshift.
+For this, we will assume that you are using the supported image, available via `python:3.7` imagestream tag in Openshift.
 Building a simple [python-sample-app](https://github.com/sclorg/s2i-python-container/tree/master/3.7/test/setup-test-app) application
 in Openshift can be achieved with the following step:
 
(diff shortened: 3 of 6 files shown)


src/README.md
```diff
@@ -43,7 +43,7 @@ in Openshift can be achieved with the following step:
 The same application can also be built using the standalone [S2I](https://github.com/openshift/source-to-image) application on systems that have it available:
 
     ```
-    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir={{ spec.version }}/test/setup-test-app/ rhscl/python-{{ spec.short_ver }}-rhel7 python-sample-app
+    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir={{ spec.version }}/test/setup-test-app/ <image_name> python-sample-app
```
torsava: The reader doesn't know what image_name format to use. "rhscl/python-38-rhel7" is not a very intuitive name. Can we list somewhere the possible image names for example?

README.md
```diff
@@ -41,102 +41,76 @@ Python versions currently provided are:
 * [python-3.8](3.8)
 
 RHEL versions currently supported are:
-* RHEL7
+* RHEL 7 ([catalog.redhat.com](https://catalog.redhat.com/software/containers))
+* RHEL 8 ([catalog.redhat.com](https://catalog.redhat.com/software/containers))
```
torsava: This link leads to a very general page. How about using `https://catalog.redhat.com/software/containers/search`?

torsava:
Besides those nitpicks, I like the changes, @frenzymadness !

2.7/README.md
```diff
@@ -43,7 +43,7 @@ in Openshift can be achieved with the following step:
 The same application can also be built using the standalone [S2I](https://github.com/openshift/source-to-image) application on systems that have it available:
 
     ```
-    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=2.7/test/setup-test-app/ rhscl/python-27-rhel7 python-sample-app
+    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=2.7/test/setup-test-app/ <image_name> python-sample-app
```
phracek: I don't see any advantage of this cleanup. Copy & paste of this command was easy. in new format you have to overwrite <image_name>

3.6/README.md
```diff
@@ -43,7 +43,7 @@ in Openshift can be achieved with the following step:
 The same application can also be built using the standalone [S2I](https://github.com/openshift/source-to-image) application on systems that have it available:
 
     ```
-    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=3.6/test/setup-test-app/ rhscl/python-36-rhel7 python-sample-app
+    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=3.6/test/setup-test-app/ <image_name> python-sample-app
```
phracek: The same as above

3.7/README.md
```diff
@@ -43,7 +43,7 @@ in Openshift can be achieved with the following step:
 The same application can also be built using the standalone [S2I](https://github.com/openshift/source-to-image) application on systems that have it available:
 
     ```
-    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=3.7/test/setup-test-app/ rhscl/python-37-rhel7 python-sample-app
+    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=3.7/test/setup-test-app/ <image_name> python-sample-app
```
phracek: The same here.

3.8/README.md
```diff
@@ -43,7 +43,7 @@ in Openshift can be achieved with the following step:
 The same application can also be built using the standalone [S2I](https://github.com/openshift/source-to-image) application on systems that have it available:
 
     ```
-    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=3.8/test/setup-test-app/ rhscl/python-38-rhel7 python-sample-app
+    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=3.8/test/setup-test-app/ <image_name> python-sample-app
```
phracek: The same here.

README.md
```diff
@@ -41,102 +41,76 @@ Python versions currently provided are:
...
-    $ cd s2i-python-container
-    $ make build TARGET=centos7 VERSIONS=3.6
-    ```
+```
+$ git clone https://github.com/sclorg/s2i-python-container.git
+$ cd s2i-python-container
+$ make build TARGET=centos7 VERSIONS=3.6
+```
 
-Note: while the installation steps are calling `podman`, you can replace any such calls by `docker` with the same arguments.
```
phracek: This is important NOTE for users who do not know the differences between podman / docker in detail. I would prefer to leave the Note here.

src/README.md
phracek: The README.md is a jinja2 template. README.md alone is generated by distgen. But agree with @torsava <image_name> is not known. Examples would be fine.

README.md
frenzymadness: You are right, thank you.

src/README.md
frenzymadness: There is no specific list of names because it depends on when you download an image from. Or you can also build it on your own and then the name might be completely different.
What about something like this: `Where <image_name> is the s2i-python image you downloaded from RHEL, Centos or Fedora registry or built from these sources.` And the "downloaded" and "built" might be links to the main readme where links to registries are provided. What do you think about it?

2.7/README.md
frenzymadness: Yes, but in many cases, the example is not valid. For example, we have Python 3.8 still supported only on Fedora but the example is the same containing `rhscl/python-38-rhel7`. The same applies on Python 3.7 which won't be available in RHEL at all. I provided a set of links where to download images for Centos, Fedora or RHEL and I can add a note about `<image_name>`, see the discussion above.

3.6/README.md
frenzymadness: This file is generated via distgen so we don't have to discuss this change for every single file.

3.7/README.md
frenzymadness: This file is generated via distgen so we don't have to discuss this change for every single file.

frenzymadness: BTW, the example is not valid and does not work. We don't have Python 3.7 on RHEL 7/8 and we never will. That's the main reason I am changing this.

3.8/README.md
frenzymadness: This file is generated via distgen so we don't have to discuss this change for every single file.

README.md
frenzymadness: The support for docker and podman is already mentioned in the very first paragraph but I have nothing against mentioning it here again.

pkubatrh: iirc the note was specifically asked for by @torsava in a previous PR

frenzymadness: I've added it to the top so it's one of the first things you see. And it's also the same for generated Readmes for Python versions.

src/README.md
pkubatrh: I myself lean a bit towards the older format (only listing rhel7 images). Sure, it might not be usable for every version (maybe add a note somewhere?) but at least the user knows what image name to use.
If you want to go with your proposal I would agree with Tomas - you need to list the image names somewhere so the user does not have to look up what to use. Which means you need to generate them with distgen. At that point, why not just generate a working example in the first place (and list alternatives)?

frenzymadness: First of all, there is no single example working with all currently supported Python versions. That's the problem I am trying to solve.
The problem with names and examples is that it might vary. For example, when I download `python-36-centos7` from docker hub, podman shows me this image with the full name `docker.io/centos/python-36-centos7`. I can use it by the full name, or just `centos/python-36-centos7` or the shortest `python-36-centos7`. But I cannot do the same with Fedora python3 image because the shortest name does not work.

I can update the proposed note this way (links do not work yet):
Where <image_name> is the s2i-python image you [downloaded from RHEL, Centos or Fedora registry]() or [built]() from these sources. For example `ubi8/python-36`, `centos/python-36-centos7` or `f31/python3`.

pkubatrh: >First of all, there is no single example working with all currently supported Python versions. That's the problem I am trying to solve.

There are working examples for some versions. For the others a substitute needs to happen first (for a Fedora or Centos image name). After this change, there are no working examples for any versions as a substitute needs to happen for all of them. To me, this seems more like a downgrade.
But again, I do not want to block you from doing it, just providing some feedback.

>But I cannot do the same with Fedora python3 image because the shortest name does not work.
This is interesting and I never noticed. If this is important maybe we should look into why this does not happen for Fedora images? Though in general I would opt for fully qualified names everywhere, as that is the only way to point to a correct image.

frenzymadness: I think that users of S2I images are experienced enough to do substitution and nobody should really use copy-pasted code examples.
From my point of view: as a user, I'd prefer a command that fails which reminds me to do a substitution instead of examples that work but might do something I don't want. Even accidentally when I copy a newline with the command and it is executed.

We can discuss it more tomorrow at the meeting.

torsava: > I think that users of S2I images are experienced enough to do substitution and nobody should really use copy-pasted code examples.

Absolutely. But let's at least point them to the list of where they can download the images from that's on the min README?

frenzymadness: > Absolutely. But let's at least point them to the list of where they can download the images from that's on the min README?

Yes, in the latest proposal, there is a plan to add a link to the Download section in the main Readme and also a link to the Build section.

frenzymadness: I am going to commit that, so we can discuss it directly and add suggestions.

frenzymadness: Please, take a look at the new note below the example code.

frenzymadness:
I forgot to mention this PR at the meeting. Could you please take a look at the latest changes?

PR #374, opened 2020-03-30
Python venv

frenzymadness:
For Python >= 3.8 use `venv` module from stdlib instead of virtualenv package. `venv` has less features but will be easier to maintain since it's part of Python itself.

### DIFF
--- a/src/centos/macros.tpl
+++ b/src/centos/macros.tpl
@@ -27,11 +27,19 @@ python
 {% endmacro %}
 
 {% macro permissions_setup(spec) %}
+{# Enable SCL only for RHEL 7 #}
 {% if spec.el_version == '7' %}
 RUN source scl_source enable {{ spec.scl }} && \
+{% else %}
+RUN \
+{% endif %}
+{# Use different virtualenv command of venv module based on Python and platform version #}
+{% if spec.version not in ["2.7", "3.6"] %}
+    python{{ spec.version }} -m venv ${APP_ROOT} && \
+{% elif spec.el_version == '7' %}
     virtualenv ${APP_ROOT} && \
 {% else %}
-RUN virtualenv-$PYTHON_VERSION ${APP_ROOT} && \
+    virtualenv-$PYTHON_VERSION ${APP_ROOT} && \
 {% endif %}
     chown -R 1001:0 ${APP_ROOT} && \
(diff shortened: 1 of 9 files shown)


src/s2i/bin/assemble
```diff
@@ -9,13 +9,19 @@ function should_collectstatic() {
 }
 
 function virtualenv_bin() {
+  {% if spec.version in ["3.8"] %}
+    # New versions of Python (>=3.8) should use venv module
+    # from stdlib instead of virtualenv package
+    python{{ spec.version }} -m venv
```
torsava:
```suggestion
    python{{ spec.version }} -m venv $1
```

torsava:
And it also needs to be removed/changed in the Dockerfile.

frenzymadness:
[test]

frenzymadness:
There are multiple issues with annobin in Fedora beta/rawhide https://bugzilla.redhat.com/show_bug.cgi?id=1817452

src/centos/macros.tpl
```diff
@@ -27,7 +27,9 @@ python
 {% endmacro %}
 
 {% macro permissions_setup(spec) %}
-{% if spec.el_version == '7' %}
+{% if spec.version == '3.8' %}
```
torsava: Can we make this `> 3.8` for future proofing? I'm not sure if it'll work though for e.g. `3.10`, might be a string comparison :/

src/s2i/bin/assemble
```diff
@@ -9,13 +9,19 @@ function should_collectstatic() {
 }
 
 function virtualenv_bin() {
+  {% if spec.version in ["3.8"] %}
```
torsava: Here also would be great to do `> 3.8`

src/centos/macros.tpl
frenzymadness: After the advice Petr gave me at the meeting today, I've reversed the condition so it now contains older versions of Python and we don't have to update it with every new Python version nor make it compare versions in the right way.

frenzymadness:
Everything seems to be okay. WDYT Tomáši?

frenzymadness:
[test]

torsava:
Looks good to me, merge when the gating finishes!

PR #375, opened 2020-03-30
New method of testing if a package is updated to the latest from PyPI

torsava:
Upstreaming change from RHEL


torsava:
@frenzymadness Review appreciated.

@pkubatrh I found this downstream in your commit. I assume the original method did not work properly?

torsava:
My bad, the code I've proposed in the PR is actually the *old* version of the functionality that was removed: https://github.com/sclorg/s2i-python-container/commit/7fdbdcf53cac39ac7ceb56215c0d090a0126e699

Apologies.

PR #377, opened 2020-04-27
Replace references to `test` subdir

worldofgeese:
Test apps now reside at `examples`. The redirect presently at `test` does not recurse correctly when invoking `oc new-app`.

### DIFF
--- a/3.6/README.md
+++ b/3.6/README.md
@@ -37,13 +37,13 @@ Building a simple [python-sample-app](https://github.com/sclorg/s2i-python-conta
 in Openshift can be achieved with the following step:
 
     ```
-    oc new-app python:3.6~https://github.com/sclorg/s2i-python-container.git --context-dir=3.6/test/setup-test-app/
+    oc new-app python:3.6~https://github.com/sclorg/s2i-python-container.git --context-dir=examples/setup-test-app/
     ```
 
 The same application can also be built using the standalone [S2I](https://github.com/openshift/source-to-image) application on systems that have it available:
 
     ```
-    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=3.6/test/setup-test-app/ rhscl/python-36-rhel7 python-sample-app
+    $ s2i build https://github.com/sclorg/s2i-python-container.git --context-dir=examples/setup-test-app/ rhscl/python-36-rhel7 python-sample-app
     ```
 
 **Accessing the application:**


frenzymadness:
Thank you for the PR. Could you please explain what happens to `oc new app` if a context directory is a symlink?

frenzymadness:
BTW, the code in this repository is managed via distgen so I cannot merge this but we can definitely discuss a fix together and I can then take care of it.

worldofgeese:
> Thank you for the PR. Could you please explain what happens to `oc new app` if a context directory is a symlink?

Thank you for considering it! Here are the complete logs:
`Cloning "https://github.com/sclorg/s2i-python-container.git" ...
	Commit:	5aefd307e6340f4c82ea4b1904aaaaa790bafb67 (Merge pull request #374 from sclorg/python_venv)
	Author:	Lumír 'Frenzy' Balhar <lbalhar@redhat.com>
	Date:	Fri Apr 3 13:15:26 2020 +0200
error: provided context directory does not exist: test/setup-test-app
`
Unrelatedly, I'm also having issues making use of the Python 3.8 imagestream, something about a missing assemble for s2i. I'm investigating that one but for now have dropped to using 3.6.

pkubatrh:
From the log it looks like you might be missing the `3.6` part of the context dir path. I tried the same command from the docs on my local Openshift 3 cluster and it managed to build the image without any issues.

pkubatrh:
As for the image stream issue - what image are you building on top? Afaik no rhel-based s2i-enabled python 3.8 image is released yet, and even if it was it would be very weird for it to not have the assemble script.

worldofgeese:
> From the log it looks like you might be missing the `3.6` part of the context dir path. I tried the same command from the docs on my local Openshift 3 cluster and it managed to build the image without any issues.

You're right about this, I feel like a real tool now. :facepalm: 



> As for the image stream issue - what image are you building on top? Afaik no rhel-based s2i-enabled python 3.8 image is released yet, and even if it was it would be very weird for it to not have the assemble script.

Try to run `oc new-app python:3.8~https://github.com/sclorg/s2i-python-container.git --context-dir=3.8/test/setup-test-app --strategy=source` and you will get the same errors I did.

pkubatrh:
Glad to hear adding the full path did the trick for you.

Running the command you provided I see: `Found Docker image a6be143 (4 days old) from Docker Hub for "python:3.8"`
Can you confirm your command picked up the same? It is unlikely a python image from Dockerhub (other than sclorg centos7 based ones) will be s2i-enabled.

worldofgeese:
> Glad to hear adding the full path did the trick for you.
> 
> Running the command you provided I see: `Found Docker image a6be143 (4 days old) from Docker Hub for "python:3.8"`
> Can you confirm your command picked up the same? It is unlikely a python image from Dockerhub (other than sclorg centos7 based ones) will be s2i-enabled.

Confirmed. So given no s2i-enabled RHEL image `oc` is going out and picking the next best thing it can find, which is not s2i-enabled, is that right thinking?

pkubatrh:
As far as I can tell that is what is happening here

PR #378, opened 2020-05-20
Add Python 3.8 for RHEL8 and preparation for RHEL7 and CentOS

torsava:
Note that rh-python38 container for EL7 still uses virtualenv instead of venv in distgit. But I think it's time to update it to venv.

### DIFF
--- a/specs/multispec.yml
+++ b/specs/multispec.yml
@@ -81,20 +81,20 @@ specs:
                   'libffi-devel', 'libtool-ltdl enchant', 'redhat-rpm-config']
 
   version:
-    "3.7":
-      version: "3.7"
-      short_ver: "37"
-      base_img_version: "1"
-      python_img_version: "1"
-      scl: "rh-python37"
-      pkg_prefix: "python3"
-
     "3.8":
       version: "3.8"
       short_ver: "38"
       base_img_version: "1"
       python_img_version: "1"
       scl: "rh-python38"
+      pkg_prefix: "python3"
+
+    "3.7":
+      version: "3.7"
+      short_ver: "37"
+      base_img_version: "1"
+      python_img_version: "1"
+      scl: "rh-python37"
       pkg_prefix: "python3"
 
     "3.6":
--- a/specs/multispec.yml
+++ b/specs/multispec.yml
@@ -64,8 +64,7 @@ specs:
       fedora_version: "31"
       s2i_base: registry.fedoraproject.org/f31/s2i-base
       img_tag: "latest"
-      python_pkgs: ['python3', 'python3-devel', 'python3-setuptools', 'python3-pip',
-                    'python3-virtualenv']
+      python_pkgs: ['python3', 'python3-devel', 'python3-setuptools', 'python3-pip']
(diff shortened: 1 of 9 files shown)


torsava:
There are some problems with the repos, so I'm trying something by manually editing the generated Dockerfile. Do not merge.

torsava:
RHSCL 3.5 is not out yet, thus RHEL7 and CentOS7 are not building yet. I've added them to the excludes, and we can enable them later.

I've tested that the RHEL7 version works using [devel repos](https://github.com/sclorg/postgresql-container/blob/master/src/Dockerfile#L65), but after talking to @pkubatrh I think it's better to wait for the GA so as not to leave the devel repos forgotten there.

torsava:
After this is merged, I'll open a PR for CentOS and RHEL7 to be merged when their repos are available.

torsava:
All tests passed!

specs/multispec.yml
```diff
@@ -48,15 +48,16 @@ specs:
...
       base_pkgs: ['nss_wrapper', 'httpd', 'httpd-devel', 'mod_ssl',
                   'mod_auth_gssapi', 'mod_ldap', 'mod_session',
                   'atlas-devel', 'gcc-gfortran', 'libffi-devel', 'libtool-ltdl enchant']
       extra_pkgs:
-        "2.7": ['{{ spec.pkg_prefix }}', '{{ spec.pkg_prefix }}-devel', 'glibc-langpack-en']
-        "3.6": ['python36', 'python36-devel']
-      ubi_versions: ['2.7', '3.6']
+        "2.7": ['{{ spec.pkg_prefix }}', '{{ spec.pkg_prefix }}-devel', 'glibc-langpack-en',
+                    '{{ spec.pkg_prefix }}-virtualenv']
+        "3.6": ['python36', 'python36-devel', '{{ spec.pkg_prefix }}-virtualenv']
```
frenzymadness, line 58: I'd use `{{ spec.pkg_prefix }}` everywhere, which means:
* remove `pkg_prefix` from Python 3.7 — it should not be used there
* redefine `pkg_prefix: "python3"` → `pkg_prefix: "python36"` for Python 3.6
* use `'{{ spec.pkg_prefix }}` for all `extra_pkgs` in RHEL 8.

WDYT?

torsava: You're right this is a bit inconsistent, the problem is that we can't do this:

> * redefine `pkg_prefix: "python3"` → `pkg_prefix: "python36"` for Python 3.6

Because in RHEL8 the python3 packages are called "python3-foo", but the python3 interpreter is called "python36" - the prefix is different for each.

However, I could use `{{ spec.pkg_prefix }}` for `python38{,-devel}` so that it's unified with python2. Does that sound ok?

frenzymadness: Yes, please do.

torsava: Amended the commit, please take a look!

torsava:
Issues resolved, waiting for CI before merge.

frenzymadness:
[test]

frenzymadness:
[test-openshift]

torsava:
[test]

frenzymadness:
Thank you!

PR #379, opened 2020-05-25
Reorder imagestreams

phracek:
This commit only reorder imagestream.
First is the latest tag and second is the latest version

### DIFF
--- a/imagestreams/python-centos7.json
+++ b/imagestreams/python-centos7.json
@@ -29,26 +29,6 @@
         }
       },
       {
-        "name": "2.7",
-        "annotations": {
-          "openshift.io/display-name": "Python 2.7",
-          "openshift.io/provider-display-name": "Red Hat, Inc.",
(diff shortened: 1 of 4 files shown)


phracek:
In reference to PR, why this change is needed https://github.com/sclorg/ansible-tests/pull/4

phracek:
@torsava @frenzymadness Can you please look at it? The PR itself only moves the latest version right after the latest tag. Nothing to more complicated.

torsava:
@phracek I did take a look already. The changes indeed reorder the imagestreams. +1

torsava:
@phracek Should Python 3.8 be added as the first entry, when it's available for all platforms?

phracek:
@torsava If it is already available for RHEL and CentOS, then yes. But images itself are not available in registry yet.

@pkubatrh What do you think?

torsava:
> @torsava If it is already available for RHEL and CentOS, then yes. But images itself are not available in registry yet.

Oh absolutely, I'm not saying you should delay the PR to wait. Just asking whether this needs to be done later. +1

frenzymadness:
Looks good to me, thank you!

Issue #290, opened 2018-10-08
Do not use the upstream Pipenv


fridex:
The recent issues on upstream Pipenv caused s2i builds to always fail:

```

New python executable in /opt/app-root/src/.local/venvs/pipenv/bin/python3
--
  | Also creating executable in /opt/app-root/src/.local/venvs/pipenv/bin/python
  | Installing setuptools, pip, wheel...done.
  | Collecting pipenv
  | Downloading https://files.pythonhosted.org/packages/eb/64/9b2747d54f2008ac3dfe86c0b1c8ec126042726fd8a540d5208d26732701/pipenv-2018.7.1-py3-none-any.whl  (5.0MB)
  | Collecting virtualenv (from pipenv)
  | Downloading https://files.pythonhosted.org/packages/b6/30/96a02b2287098b23b875bc8c2f58071c35d2efe84f747b64d523721dc2b5/virtualenv-16.0.0-py2.py3-none-any.whl  (1.9MB)
  | Collecting certifi (from pipenv)
  | Downloading https://files.pythonhosted.org/packages/df/f7/04fee6ac349e915b82171f8e23cee63644d83663b34c539f7a09aed18f9e/certifi-2018.8.24-py2.py3-none-any.whl  (147kB)
  | Collecting pip>=9.0.1 (from pipenv)
  | Downloading https://files.pythonhosted.org/packages/c2/d7/90f34cb0d83a6c5631cf71dfe64cc1054598c843a92b400e55675cc2ac37/pip-18.1-py2.py3-none-any.whl  (1.3MB)
  | Collecting virtualenv-clone>=0.2.5 (from pipenv)
  | Downloading https://files.pythonhosted.org/packages/6d/c2/dccb5ccf599e0c5d1eea6acbd058af7a71384f9740179db67a9182a24798/virtualenv_clone-0.3.0-py2.py3-none-any.whl
  | Collecting setuptools>=36.2.1 (from pipenv)
  | Downloading https://files.pythonhosted.org/packages/96/06/c8ee69628191285ddddffb277bd5abdf769166e7a14b867c2a172f0175b1/setuptools-40.4.3-py2.py3-none-any.whl  (569kB)
  | Installing collected packages: virtualenv, certifi, pip, virtualenv-clone, setuptools, pipenv
  | Found existing installation: pip 9.0.1
  | Uninstalling pip-9.0.1:
  | Successfully uninstalled pip-9.0.1
[...]

Issue #278, opened 2018-06-08
include pipenv at build of s2i container image rather than downloading it during use of s2i


goern:
As a S2I user, I want use python 3.6 with Pipenv, so that pipenv has a specific version installed during build of the s2i builder image and is not pulled down from the internet during usage of the s2i builder image.

touchardv:
Related to this issue, it is not possible to use pipenv in a closed environment (no direct access to the Internet) as it can not be installed properly (it seems the `--isolated` argument makes pip ignore `PIP_INDEX_URL` that could be used to go via an internal mirror).

Therefore, yes, it would be great to have a default pipenv installed in the image.

goern:
@hhorak ping

omron93:
@torsava ???

torsava:
The problem is, pipenv is not packaged in any version of CentOS right now, only Fedora. We could theoretically `pip install` it during the build of the image, but then we're distributing an untested version right in the image. I'm leaning towards not proceeding with this for the time being. Let's fix the `PIP_INDEX_URL` for the closed environment so that can be used as an alternative.

omron93:
> theoretically pip install it during the build of the image, but then we're distributing an untested version right in the image (plus it's unpackaged dependencies)

I thing this wouldn't be even possible in internal build system (used by Fedora/RHEL images). Only rpms can be installed during the build. (or some dirty hack is needed to be able to access Internet)

torsava:
> > theoretically pip install it during the build of the image, but then we're distributing an untested version right in the image (plus it's unpackaged dependencies)
> 
> I thing this wouldn't be even possible in internal build system (used by Fedora/RHEL images). Only rpms can be installed during the build. (or some dirty hack is needed to be able to access Internet)

Ah, indeed. Thus I don't consider this feasible.

torsava:
> > pipenv is not packaged in any version of CentOS right now, only Fedora
> 
> @torsava so I understand that using CentOS is a hard requirement?

We want to keep the Fedora/CentOS versions functionally as similar as possible.

> 
> > but then we're distributing an untested version right in the image (plus it's unpackaged dependencies).
> 
> which is even better than downloading a (potentially completely broken) version from the internet during build of an application via s2i

In that case one is already downloading other packages using pipenv, so adding pipenv itself is not a large departure from that.

frenzymadness:
[...]

PR #383, opened 2020-05-30
hotfix: Pipenv version lock

### LINKED ISSUE #382: Latest Pipenv changed behaviour on virtualenv isolation = dependencies are skipped to install
The latest release of Pipenv https://github.com/pypa/pipenv/releases/tag/v2020.5.28 has changed the behaviour of `pipenv install` https://github.com/pypa/pipenv/issues/3057. As a result, Pipenv no longer installs packages that can be located in its path. This breaks s2i builds, since Pipenv is run in a different context than the actual applications. Let me illustrate:

Trying to install any package, that is already present in the global env, results in Pipenv skipping this package installation into the pipenv's virtual env. That includes all the dependencies of Pipenv itself like `six` or `importlib_metadata` or `certifi`. And these packages are used as a dependency in many bigger libraries, which transitively makes them misbehave and break...

Let's assume we have an app using `certifi` and `chardet`. The first named is a Pipenv dependency and the second one is not:
```
$ echo "import chardet, certifi" > app.py

$ pipenv install certifi chardet

$ grep -A 2 "\[packages\]" Pipfile
[packages]
certifi = "*"
chardet = "*"

$ s2i build -c . registry.access.redhat.com/ubi8/python-38:latest test -e ENABLE_PIPENV=true

$ docker run test
---> Running application from Python script (app.py) ...
  File "app.py", line 1, in <module>
    import chardet, certifi
ModuleNotFoundError: No module named 'certifi'
```
Now let's see where are the packages installed and what our python path actually is: 
```
$ docker run test find / -type d -name certifi 2>/dev/null | grep -v vendor
/opt/app-root/src/.local/venvs/pipenv/lib/python3.8/site-packages/certifi

$ docker run test find / -type d -name chardet 2>/dev/null | grep -v vendor
/opt/app-root/lib/python3.8/site-packages/chardet

$ docker run test python -c "import sys; print(sys.path)"
['', '/usr/lib64/python38.zip', '/usr/lib64/python3.8', '/usr/lib64/python3.8/lib-dynload', '/opt/app-root/lib64/python3.8/site-packages', '/opt/app-root/lib/python3.8/site-packages']
```

As you can see, each of the packages is located in different location. One is reachable for the python runtime and one is not.

This issue means that any new build (based on s2i Python with Pipenv and started after the Pipenv update) of an application that shares a single (or more) dependency with Pipenv will fail at build time or at runtime for a really inexplicable reason - because a dependency which is specified in the Pipfile doesn't get installed.

It took me really long to trace this issue all the way back to Pipenv, I'm not sure how to solve this without locking `pipenv` version though.

tumido:
Fixes #382 by locking pipenv version to [2018.11.26](https://pypi.org/project/pipenv/2018.11.26/) which is a [previous release to the current latest](https://pypi.org/project/pipenv/#history)

If there's a better, more robust fix, please close this PR in favor of that one.

This change is abusing the `install_tool`'s `$VENV_DIR/bin/pip --isolated install -U $1$2` notation.

### DIFF
--- a/2.7/s2i/bin/assemble
+++ b/2.7/s2i/bin/assemble
@@ -63,7 +63,7 @@ if [[ ! -z "$UPGRADE_PIP_TO_LATEST" || ! -z "$ENABLE_PIPENV" ]]; then
 fi
 
 if [[ ! -z "$ENABLE_PIPENV" ]]; then
-  install_tool "pipenv"
+  install_tool "pipenv" "==2018.11.26"
   echo "---> Installing dependencies via pipenv ..."
   if [[ -f Pipfile ]]; then
     pipenv install --deploy
--- a/3.6/s2i/bin/assemble
+++ b/3.6/s2i/bin/assemble
@@ -62,7 +62,7 @@ if [[ ! -z "$UPGRADE_PIP_TO_LATEST" ]]; then
 fi
 
 if [[ ! -z "$ENABLE_PIPENV" ]]; then
-  install_tool "pipenv"
+  install_tool "pipenv" "==2018.11.26"
   echo "---> Installing dependencies via pipenv ..."
   if [[ -f Pipfile ]]; then
     pipenv install --deploy
--- a/3.7/s2i/bin/assemble
+++ b/3.7/s2i/bin/assemble
@@ -58,7 +58,7 @@ if [[ ! -z "$UPGRADE_PIP_TO_LATEST" ]]; then
 fi
 
 if [[ ! -z "$ENABLE_PIPENV" ]]; then
-  install_tool "pipenv"
+  install_tool "pipenv" "==2018.11.26"
   echo "---> Installing dependencies via pipenv ..."
   if [[ -f Pipfile ]]; then
     pipenv install --deploy
--- a/3.8/s2i/bin/assemble
+++ b/3.8/s2i/bin/assemble
@@ -58,7 +58,7 @@ if [[ ! -z "$UPGRADE_PIP_TO_LATEST" ]]; then
 fi
 
 if [[ ! -z "$ENABLE_PIPENV" ]]; then
-  install_tool "pipenv"
+  install_tool "pipenv" "==2018.11.26"
   echo "---> Installing dependencies via pipenv ..."
   if [[ -f Pipfile ]]; then
     pipenv install --deploy
--- a/src/s2i/bin/assemble
+++ b/src/s2i/bin/assemble
@@ -71,7 +71,7 @@ if [[ ! -z "$UPGRADE_PIP_TO_LATEST"{% if spec.version in ["2.7"] %} || ! -z "$EN
 fi
 
 if [[ ! -z "$ENABLE_PIPENV" ]]; then
-  install_tool "pipenv"
+  install_tool "pipenv" "==2018.11.26"
   echo "---> Installing dependencies via pipenv ..."
   if [[ -f Pipfile ]]; then
     pipenv install --deploy


goern:
This seems to bring back a discussion we have had earlier: we should not install parts of the toolchain during build of the application. 

Could we think about moving the required tools into the container's build process/time rather than application build?

tumido:
> This seems to bring back a discussion we have had earlier: we should not install parts of the toolchain during build of the application. 

Since I wasn't part of the discussion earlier I may not see the whole context here. However, my 2 cents: if we want and expect to deliver reproducible and deterministic builds we shouldn't count on dynamic installation of the tools marketed as provided by the builder image.

If I rerun an older build - based on the exact same s2i Python revision as before, same sha digest, the same app source commit hash, the same dependencies locked - the assemble script will still may go and install a different pipenv version. That makes the app build process nondeterministic.

In my eyes it should handle the provided tools in the same way as the builder image "locks" and bakes in the Python micro version releases.

The other thing is, there was no new pipenv version for 2 years. I totally can see why it was counted on as a stable, solid thing...

fridex:
> > This seems to bring back a discussion we have had earlier: we should not install parts of the toolchain during build of the application.
> 
> Since I wasn't part of the discussion earlier I may not see the whole context here. However, my 2 cents: if we want and expect to deliver reproducible and deterministic builds we shouldn't count on dynamic installation of the tools marketed as provided by the builder image.


We had this discussion almost exactly 2 years back in https://github.com/sclorg/s2i-python-container/issues/278 and https://github.com/sclorg/s2i-python-container/issues/290, both still open.

+1 on my side - the builder container image should come with all the tools necessary to build the application without relying on PyPI. An example could be builds happening in an isolated environment without any access to Internat and installing packages solely from the intranet (see also linked issues stating broken releases of Pipenv itself that blocked our development for quite some time at that point).

> If I rerun an older build - based on the exact same s2i Python revision as before, same sha digest, the same app source commit hash, the same dependencies locked - the assemble script will still may go and install a different pipenv version. That makes the app build process nondeterministic.
>
> In my eyes it should handle the provided tools in the same way as the builder image "locks" and bakes in the Python micro version releases.

+1 on both paragraphs
 
> The other thing is, there was no new pipenv version for 2 years. I totally can see why it was counted on as a stable, solid thing...

The situation around Pipenv [was different](https://github.com/pypa/pipenv/issues/4058).

---

Anyway, what are the arguments not to provide the ability to install dependencies using Pipenv files by default in Python s2i container images? As Pipenv is in PyPA, it looks alive again now and Pipenv files seem to be the way PyPA is expecting lock files should look like?

I can [imagine size could be an issue](https://github.com/fridex/s2i-example-micropipenv#micropipenv-vs-pipenv-size), but [we have a solution for that](https://github.com/thoth-station/micropipenv) (that can help also when installing requirements from requirements.txt as micropipenv can resolve issues caused by relative ordering of dependencies installed).

frenzymadness:
This is a real problem we have to solve. Unfortunately, neither pipenv nor micropipenv is available in RHEL/Centos so we cannot install them from RPMs in all our container images. Moreover, pipenv has a lot of dependencies and even more bundled dependencies so maintain a package like that in RHEL is painful and nobody would want to include it there. Fixing CVEs in bundled libraries is a thing I don't want to spend my time on more than necessary. 

Unfortunately, we cannot install these tools from PyPI during the build of the container images because our build platforms have no internet access during the build.

But yes, we have to find a way how to avoid problems like this one.

The best of what we can do now is to fix pipenv upstream which would also fix all of our containers. I'm gonna report the issue upstream. Then we can lock pipenv to specific version but we also have to find a way how to test the images with the latest versions. We'll try to find a better solution.

fridex:
> This is a real problem we have to solve. Unfortunately, neither pipenv nor micropipenv is available in RHEL/Centos so we cannot install them from RPMs in all our container images.

I understand that maintaining different solutions for RHEL/Centos and Fedora adds additional overhead, but long-term accomodating micropipenv once it eventually goes to RHEL IMHO could be an option to consider.

>Moreover, pipenv has a lot of dependencies and even more bundled dependencies so maintain a package like that in RHEL is painful and nobody would want to include it there. Fixing CVEs in bundled libraries is a thing I don't want to spend my time on more than necessary.

This was one of the main reasons [why micropipenv was introduced](https://github.com/thoth-station/micropipenv#whats-the-difference-in-comparision-to-pipenv).

> Unfortunately, we cannot install these tools from PyPI during the build of the container images because our build platforms have no internet access during the build.

To my understanding that's how even the current build process works. Pipenv won't be installed if there is no internet access as it is installed during application assembling (not during the actual build of the builder container image). Hence, users without Internet access won't have any support for projects that use Pipenv files for managing dependencies.

> The best of what we can do now is to fix pipenv upstream which would also fix all of our containers. I'm gonna report the issue upstream. Then we can lock pipenv to specific version but we also have to find a way how to test the images with the latest versions. We'll try to find a better solution.

Even if they do fix this issue, the issue described above and in https://github.com/sclorg/s2i-python-container/issues/278 will remain.

frenzymadness:
> > This is a real problem we have to solve. Unfortunately, neither pipenv nor micropipenv is available in RHEL/Centos so we cannot install them from RPMs in all our container images.
> 
> I understand that maintaining different solutions for RHEL/Centos and Fedora adds additional overhead, but long-term accomodating micropipenv once it eventually goes to RHEL IMHO could be an option to consider.

That definitely is. I'd like to see micropipenv in RHEL and no installations from PyPI in assemble script.

> > Unfortunately, we cannot install these tools from PyPI during the build of the container images because our build platforms have no internet access during the build.
> 
> To my understanding that's how even the current build process works. Pipenv won't be installed if there is no internet access as it is installed during application assembling (not during the actual build of the builder container image). Hence, users without Internet access won't have any support for projects that use Pipenv files for managing dependencies.

If users of s2i don't have internet access, they usualy provides their own packages index and this index might include also pipenv and its dependencies. This approach is not flawless now but we have a plan how to fix it.

> > The best of what we can do now is to fix pipenv upstream which would also fix all of our containers. I'm gonna report the issue upstream. Then we can lock pipenv to specific version but we also have to find a way how to test the images with the latest versions. We'll try to find a better solution.
> 
> Even if they do fix this issue, the issue described above and in #278 will remain.

The issue is already fixed upstream but we'll  still lock the version and try to find a better solution.

frenzymadness:
[test]

frenzymadness:
[test-openshift]

fridex:
> > > This is a real problem we have to solve. Unfortunately, neither pipenv nor micropipenv is available in RHEL/Centos so we cannot install them from RPMs in all our container images.
> > 
> > 
> > I understand that maintaining different solutions for RHEL/Centos and Fedora adds additional overhead, but long-term accomodating micropipenv once it eventually goes to RHEL IMHO could be an option to consider.
> 
> That definitely is. I'd like to see micropipenv in RHEL and no installations from PyPI in assemble script.

Is there anything needed from our side to support this idea?


> > > Unfortunately, we cannot install these tools from PyPI during the build of the container images because our build platforms have no internet access during the build.
> > 
> > 
> > To my understanding that's how even the current build process works. Pipenv won't be installed if there is no internet access as it is installed during application assembling (not during the actual build of the builder container image). Hence, users without Internet access won't have any support for projects that use Pipenv files for managing dependencies.
> 
> If users of s2i don't have internet access, they usualy provides their own packages index and this index might include also pipenv and its dependencies. This approach is not flawless now but we have a plan how to fix it.

What is the current plan?

frenzymadness:
> > That definitely is. I'd like to see micropipenv in RHEL and no installations from PyPI in assemble script.
> 
> Is there anything needed from our side to support this idea?

We'll have to discuss it and find answers to some questions like:
* Is micropipenv reliable and stable enough that we can use just it and nothing else?
* Even it makes perfect sense to have a lighter tool in containers, will users accept micropipenv as a replacement or will they still want to use the same tool everywhere (pipenv).

> > If users of s2i don't have internet access, they usualy provides their own packages index and this index might include also pipenv and its dependencies. This approach is not flawless now but we have a plan how to fix it.
> 
> What is the current plan?

The assemble script will try to install [micro]pipenv without `--isolated` first which means that if a user provides some specific settings for pip (like their own index), the tools might be installed from their index without internet access. If this try fails (something is missing at the private index) the script will try it again with `--isolated` which will need internet access.

fridex:
> > > That definitely is. I'd like to see micropipenv in RHEL and no installations from PyPI in assemble script.
> > 
> > 
> > Is there anything needed from our side to support this idea?
> 
> We'll have to discuss it and find answers to some questions like:
> 
> * Is micropipenv reliable and stable enough that we can use just it and nothing else?
> * Even it makes perfect sense to have a lighter tool in containers, will users accept micropipenv as a replacement or will they still want to use the same tool everywhere (pipenv).

Makes sense, please keep us in any related discussion. :+1:

frenzymadness:
[test]

frenzymadness:
CI looks green also for RHEL images so I'm gonna merge this PR and update Fedora images. Thank you!

frenzymadness:
Bodhi updates:
* F31: https://bodhi.fedoraproject.org/updates/FEDORA-CONTAINER-2020-f2436e3673
* F32: https://bodhi.fedoraproject.org/updates/FEDORA-CONTAINER-2020-b7d205861b

It would be awesome if somebody can test them and give them a karma points so they can reach stable registry soon.

PR #388, opened 2020-06-23
Isolated fix for pip

frenzymadness:


frenzymadness:
Failure on Fedora is expected because of the data center move.

PR #380, opened 2020-05-27
Add Python 3.8 for RHEL7

torsava:
This is a follow up to PR https://github.com/sclorg/s2i-python-container/pull/378 which prepared the repo for Python 3.8 for RHEL7 and CentOS, but the RHSCL 3.5 repos were not yet available. This PR is to be merged when the repos become available.

PR #389, opened 2020-06-23
Add Python 3.8 for CentOS7

torsava:
Follow up on https://github.com/sclorg/s2i-python-container/pull/380.


frenzymadness:
Manually rebased and the CI seems to be finally green.

torsava:
Woohoo!!

PR #391, opened 2020-07-08
Fix docs for PIP_INDEX_URL

torsava:
@frenzymadness I think we left this in by accident.

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -146,9 +146,12 @@ file inside your source code repository.
 
 * **PIP_INDEX_URL**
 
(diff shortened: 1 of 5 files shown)


frenzymadness:
Good catch. Do you think that the second sentence is okay? I'd add there a note that pipenv and micropipenv are also installed from a custom index if possible.

torsava:
> Good catch. Do you think that the second sentence is okay? I'd add there a note that pipenv and micropipenv are also installed from a custom index if possible.

You're right, that needed modification as well. What do you think of the current wording? @frenzymadness

frenzymadness:
Perfect!

PR #392, opened 2020-07-13
Update imagestreams

yselkowitz:


frenzymadness:
The diff looks good to me. Is there any way how to test it?

yselkowitz:
AFAIK only within the context of OpenShift, which should happen after this is merged here, automatically synced into openshift/library, then imported from there into the cluster-samples-operator.

frenzymadness:
And are you able to do it after I merge this PR?

yselkowitz:
No, I'm not set up for that.  How do you normally deal with imagestreams updates?

yselkowitz:
To clarify, I did make sure that these changes will import and verify in the openshift library.

frenzymadness:
Okay, let's test this. Thank you!

PR #398, opened 2020-09-15
Remove streams that will be soon EOL.

hhorak:


hhorak:
[test-openshift]

hhorak:
@yselkowitz @phracek @pkubatrh WDYT?

yselkowitz:
Besides the fact that it seems a bit odd to EOL rh-python36 when python3 is supported in RHEL7 for the rest of its life, LGTM, but any explanation for the test failures?

yselkowitz:
Mentions of 3.6-ubi7 as an option in django-ex should also be removed -- even better if it can be moved to 3.8 instead.

hhorak:
> Besides the fact that it seems a bit odd to EOL rh-python36 when python3 is supported in RHEL7 for the rest of its life

Yeah, it might seem odd, but those are two different stacks, with different use cases and reasons for inclusion; they only have one common thing, the same python version.

pkubatrh:
[test-openshift]

hhorak:
We still didn't convince Jenkins to work properly, so I verified manually that the imagestream is valid. Merging.

PR #406, opened 2020-10-26
Document usage with a Dockerfile

fila43:
This adds the following:
*  Dockerfile examples for a variant that uses s2i scripts and a variant without s2i scripts
 * Documentation enhancement that describes how the package can be used using a Dockerfile
 * a test for the example Dockerfiles

This is a similar pull-request as sclorg/httpd-container#99., sclorg/nodejs-container#247 and sclorg/perl-container#192

To see the formatted README.md, see https://github.com/fila43/s2i-python-container/blob/master/2.7/README.md and it is also visible on the catalog, which will be the most important view: https://catalog.redhat.com/software/containers/rhel8/python-27/5ba244d35a134643ef2f048c

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -29,33 +29,112 @@ modules for their web applications. There is no guarantee for any specific npm o
 version, that is included in the image; those versions can be changed anytime and
 the nodejs itself is included just to make the npm work.
 
-Usage
----------------------
+Usage in Openshift
+------------------
 
 For this, we will assume that you are using the supported image, available via `python:2.7` imagestream tag in Openshift.
-Building a simple [python-sample-app](https://github.com/sclorg/s2i-python-container/tree/master/2.7/test/setup-test-app) application
+Building a simple [python-sample-app](https://github.com/sclorg/django-ex.git) application
 in Openshift can be achieved with the following step:
 
     ```
-    oc new-app python:2.7~https://github.com/sclorg/s2i-python-container.git --context-dir=2.7/test/setup-test-app/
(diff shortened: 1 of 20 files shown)


fila43:
[test]

frenzymadness:
First of all, thank you very much for this work. It's on my todo list for a while :slightly_smiling_face: Unfortunately, this repository uses distgen which means that the content of the folders for all Python versions is generated from source files. No worries. I'll take a look how to adapt your work to this process and I'll also take a look how to make the info about distgen more visible for new contributors.

frenzymadness:
My notes and questions:
* Why was necessary to switch to Django example from the internal setup-based example? I'm just curious. It's actually good and we can use the similar way 
* Updated: `/usr/libexec/s2i/run` does much more than just execution of the Django web server.
* Updated: `s/Node.js/Python/g` in the readme
* Python 3.7 is not available for RHEL/Centos but the readme will be generated anyway so I'll try to find a way how to put Fedora there.
* The completely manual way should be mentioned but IMHO without an example as the example would look like very differently for modules/software collections/Fedora and it would be hard to maintain. I'd also say that it's suitable only for experts and it's not future-proof.
* The folder with examples is something we'd like to show as examples of possible setups so IMHO the generated dockerfiles are better for tests than examples. Also, the thing with the manual installations applies here as well, so I'd test it only with s2i scripts.

I'm gonna prepare a new pull request with your work and my updates on top of it so we can discuss it further.

frenzymadness:
The new version of this is in #408

PR #407, opened 2020-10-30
set SCL env variables on build

fila43:

PR #408, opened 2020-11-03
Dockerfile usage

frenzymadness:
Based on #406 from @fila43. Updated to use distgen and some small updates. Will see what CI thinks about the generated Dockerfile tests.

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -29,28 +29,104 @@ modules for their web applications. There is no guarantee for any specific npm o
 version, that is included in the image; those versions can be changed anytime and
 the nodejs itself is included just to make the npm work.
 
-Usage
+Usage in Openshift
 ---------------------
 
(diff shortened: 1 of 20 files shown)


frenzymadness:
CI is green. @fila43 @hhorak @torsava do you want to take a look? According to a discussion with @fila43, we should be able to add also fully-manual dockerfiles (without s2i scripts) after #407 is done.

hhorak:
@frenzymadness, @fila43 is correct, this is important use case for all non-OpenShift use cases, but also in OpenShift. Especially on RHEL-8 where s2i tool is not available as a standalone package, but not only, the Dockerfile is de-facto standard, and people already use the images that way. We just need to make it easier for them by documenting how to do it. So, if you don't see any other issue, I'm ok with merging..

frenzymadness:
My plan is:
- merge this PR
- finish #407 
- add some general dockefile usage without s2i scripts as @fila43 proposed in his PR

PR #411, opened 2020-11-19
Fix pip downgrade

frenzymadness:
The current implementation needs at least pip 19.3.1 to support new wheels etc. which is fine for RHEL and Centos but in Fedora, we already have a newer pip version so this change should prevent a downgrade if a newer version is already available.

@torsava could you please take a look?


frenzymadness:
[test][test-openshift]

torsava:
Ah, good catch, downgrading's not good :)

It I played with the code on the command line, and it seems to always install the latest version of pip (since it's >=19.3.1).
I think we don't want that: 1. that would mean that the `UPGRADE_TO_LATEST` stopped being useful, and 2. we would always be using a new version of pip, which could very well break the container with any new release of pip.

What do you think?

frenzymadness:
I'm gonna try to implement it with an internal wheel from Fedora, which should bring stable, tested and patched pip to the containers as a default choice.

frenzymadness:
Replaced by #412

PR #412, opened 2020-11-24
Get pip from wheel and install it during build

frenzymadness:
This change will basically make pip in app root to be installed from wheel we have in Fedora. Wheel from Fedora has many advantages (including downstream patches) and it's more stable than anything we can install from PyPI. We'll also have one version for all platforms.

There is minimal impact on image size as the wheel itself is removed after the installation. The only negative thing is one more layer (causing 10 MB growth) which can be removed by squash. Otherwise, /opt folder is only 1 MB bigger than before which is caused by the newer pip.

Now, pip is baked into the images and we donẗ have to spend time/resources on its update in assemble script and do it possibly multiple times depending on the customer settings.

Pip maintainers do releases four times a year so this won't cause any maintenance burden and it makes builds more reproducible with one less dependency on external sources. However, we have to make sure that the future version of pip will still support all Python versions we have to support. Python 2 support will be EOL in pip in version 21.1 in January 2021.


frenzymadness:
[test][test-openshift]

frenzymadness:
One more improvement implemented during fixing a bug. The bug was caused by wrong permission so the pip wasn't able to update itself. I moved the installation step to macros between the create venv step and the fix permission step so it no longer adds a new layer to the resulting image.

frenzymadness:
[test][test-openshift]

phracek (commented):
Please update also README.md. This change is IMPORTANT and should be documented.

frenzymadness:
[test][test-openshift]

frenzymadness:
I've managed to switch the docker build context to the root folder so now we can use symlinks to the different folders and maintain the new wheel file only as one copy.

The disadvantage is one more layer in the build but it does not add any significant space. The resulting image is 2 MB bigger than the one currently produced from master branch (Centos 7, Python 3.8) – One MB is caused by the newer and therefore bigger pip and the second one is probably caused by the new layer. If I use `--squash` the size of the resulting image is identical.

frenzymadness:
I think I'll need multiple reviews here because the change is kinda complex, so Ccing @torsava @hhorak @phracek @pkubatrh 

The CI failure for OS does not seem to be related to this change.

frenzymadness:
Rebased to the latest master [test][test-openshift]

2.7/Dockerfile
```diff
@@ -55,10 +55,16 @@ RUN INSTALL_PKGS="libjpeg-turbo libjpeg-turbo-devel python27 python27-python-dev
...
 
 # Copy extra files to the image.
-COPY ./root/ /
+COPY 2.7/root/ /
+
+# Yes, the directory below is already copied by the previous command.
+# The problem here is that the wheels directory is copied as a symlink.
+# Only if you specify symlink directly as a source, COPY copies all the
+# files from the symlink destination.
+COPY 2.7/root/opt/wheels /opt/wheels
```
torsava, line 67: `/opt/` is kind of a free for all, so I'm thinking whether we shouldn't choose some longer directory name, to decrease likelihood of any possible conflicts in the future? Or maybe I'm being too paranoid.

src/README.md
```diff
@@ -242,8 +242,8 @@ file inside your source code repository.
 
     Set this variable to a non-empty value to have the 'pip' program and related
     python packages (setuptools and wheel) be upgraded to the most recent version
-    before any Python packages are installed. If not set it will use whatever
-    the default version is included by the platform for the Python version being used.
+    before any Python packages are installed. If not set it will use the default
+    version from the latest stable Fedora we have in the source repository in a form of wheel.
```
torsava, line 246:
```suggestion
    before any Python packages are installed. If not set, the container will use
    the stable pip version this container was built with, taken from a recent Fedora release.
```

If we plan to update once every year, it will not always be from the latest Fedora. What do you think of this wording?

src/centos/macros.tpl
```diff
@@ -41,6 +41,15 @@ RUN \
 {% else %}
     virtualenv-$PYTHON_VERSION ${APP_ROOT} && \
 {% endif %}
+    # We have to upgrade pip to a newer verison because: \
+    # * pip < 9 does not support different packages' versions for Python 2/3 \
+    # * pip < 19.3 does not support manylinux2014 wheels. Only manylinux2014 wheels \
+    #   support platforms like ppc64le, aarch64 or armv7 \
+    # We are newly using wheel from the latest stable Fedora release (RPM python-pip-wheel) \
```
torsava, line 48:
```suggestion
    # We are newly using wheel from one of the latest stable Fedora releases (from RPM python-pip-wheel) \
```

src/centos/macros.tpl
```diff
@@ -41,6 +41,15 @@ RUN \
...
+    # because it's tested better that whatever version from PyPI and contains useful patches. \
+    # We have to do it here so the permissions are correctly fixed and pip is able  to reinstall \
+    # itself in the next build phases and in the assemble script if user wants the latest version \
+    ${APP_ROOT}/bin/pip install /opt/wheels/pip-* && rm -r /opt/wheels && \
```
torsava, line 52: > rm -r /opt/wheels 

Don't we need `rm -rf` for a directory?

src/fedora/macros.tpl
```diff
@@ -16,6 +16,15 @@ ENV NAME=python3 \
 
 {% macro permissions_setup(spec) %}
 RUN python{{ spec.version }} -m venv ${APP_ROOT} && \
+# We have to upgrade pip to a newer verison because: \
+# * pip < 9 does not support different packages' versions for Python 2/3 \
+# * pip < 19.3 does not support manylinux2014 wheels. Only manylinux2014 wheels \
+#   support platforms like ppc64le, aarch64 or armv7 \
+# We are newly using wheel from the latest stable Fedora release (RPM python-pip-wheel) \
```
torsava, line 23:
```suggestion
# We are newly using wheel from one of the latest stable Fedora releases (from RPM python-pip-wheel) \
```

torsava:
Besides the minor comments, it looks good to me, @frenzymadness !

2.7/Dockerfile
frenzymadness: Something like `__wheels`?

src/centos/macros.tpl
frenzymadness: we need only `-r` eg `--recursive` to remove directories and their contents recursively

frenzymadness:
Suggestions from Tomáš applied, squashed, regenerated …

[test][test-openshift]

2.7/Dockerfile
torsava: How about `__wheels__`? :D

pkubatrh: Is there a reason for not using `/tmp` considering we are removing the directory shortly after?

torsava: @pkubatrh Good question. However, that makes me question if we should be even removing it. I think we're shipping the images not squashed, is that right? If so, we could just as well keep the file in the container, for debugging and non-standard use cases.

pkubatrh: rhel-based images are definitely getting squashed. Not sure about centos images now that we are moving to quay.io. @phracek might know

torsava: Ah, ok, if that's so, Ii agree that /tmp is not a bad choice.

frenzymadness: Switched to `/tmp/wheels/`.

frenzymadness:
Rebased, improved [test][test-openshift]

I'm gonna discuss the wheel usage here with product security team.

frenzymadness:
I still don't have any response from the security team :disappointed: I'm gonna ping them again.

frenzymadness:
So, the response is that the security team basically doesn't track the content of the containers. I still think this is a good idea to implement but we have to find a way how to automate it or at least check and compare versions we have here and in the latest stable Fedora. Let's discuss it at the next meeting.

fridex:
> So, the response is that the security team basically doesn't track the content of the containers. I still think this is a good idea to implement but we have to find a way how to automate it or at least check and compare versions we have here and in the latest stable Fedora. Let's discuss it at the next meeting.

I don't have context here but this sounds like we could help here. We extract information about packages present in a container image, discover RPMs or Python packages installed and shipped. We already have a service up and running which can give you a JSON report with packages present. Feel free to let me know if you want to use something like that.

frenzymadness:
[test] [test-openshift]

frenzymadness:
The latest implementation uses bundled wheel for Python 3 and the latest supported pip from PyPI for Python 2.

frenzymadness:
[test] [test-openshift]

frenzymadness:
There is a problem with permissions when I use "/tmp" as a destination for wheels because it for some reason changes permissions for that folder. Hence I've tried to use "/opt/wheels" instead.

frenzymadness:
[test] [test-openshift]

frenzymadness:
[test] [test-openshift]

frenzymadness:
[test] [test-openshift]

frenzymadness:
After some trials and errors, this implementation is ready for a review

src/README.md
```diff
@@ -266,10 +266,15 @@ file inside your source code repository.
 
 * **UPGRADE_PIP_TO_LATEST**
 
+    {% if spec.version == "2.7" %}
+    Python 2 images are using the latest compatible pip release (<21.0) so this
+    option controls update of setuptools and wheel packages only.
+    {% else %}
     Set this variable to a non-empty value to have the 'pip' program and related
```
torsava, line 273: The Python 2 description is too terse, it's missing "Set this variable to a non-empty value to have the ..." for example.

src/centos/macros.tpl
```diff
@@ -41,6 +41,19 @@ RUN \
 {% else %}
     virtualenv-$PYTHON_VERSION ${APP_ROOT} && \
 {% endif %}
+{% if spec.version != "2.7" %}
+    # Python 3 only code, Python 2 installs pip from PyPI in the assemble script. \
+    # We have to upgrade pip to a newer verison because: \
+    # * pip < 9 does not support different packages' versions for Python 2/3 \
+    # * pip < 19.3 does not support manylinux2014 wheels. Only manylinux2014 wheels \
```
torsava, line 48: Future proofing: `Only manylinux2014 (and later) wheels` ... there's probably gonna be a new manylinux standard.

src/centos/macros.tpl
```diff
@@ -41,6 +41,19 @@ RUN \
...
+    #   support platforms like ppc64le, aarch64 or armv7 \
+    # We are newly using wheel from one of the latest stable Fedora releases (from RPM python-pip-wheel) \
+    # because it's tested better that whatever version from PyPI and contains useful patches. \
```
torsava, line 51:
```suggestion
    # because it's tested better that newer versions from PyPI and contains useful patches. \
```

src/centos/macros.tpl
```diff
@@ -41,6 +41,19 @@ RUN \
...
+    # We have to do it here so the permissions are correctly fixed and pip is able  to reinstall \
+    # itself in the next build phases and in the assemble script if user wants the latest version \
```
torsava, line 53:
```suggestion
    # itself in the next build phases in the assemble script if user wants the latest version \
```

Like this? Or is it both in the next build phase and in the assemble script?

src/centos/macros.tpl
frenzymadness: I'd say that the Fedora wheel is also tested better than older versions from PyPI. Isn't it? It's only comment, nothing really visible to users.

src/centos/macros.tpl
frenzymadness: Fixed everywhere.

src/README.md
frenzymadness: Fixed.

src/centos/macros.tpl
torsava: Oh, I get what you mean. In that case there's a typo: `that` instead of `then`.

src/s2i/bin/assemble
```diff
@@ -66,17 +66,20 @@ mv /tmp/src/* "$HOME"
...
+echo "---> Upgrading pip to version <21.0 (The last Python 2 compatible version) ..."
+if ! pip install -U "pip<21.0"; then
+  echo "WARNING: Installation of 'pip<21.0' failed, trying again from official PyPI with pip --isolated install"
+  pip install --isolated -U "pip<21.0"
 fi
+{% endif %}
 
-if [[ ! -z "$UPGRADE_PIP_TO_LATEST"{% if spec.version in ["2.7"] %} || ! -z "$ENABLE_PIPENV"{% endif %} ]]; then
+if [[ ! -z "$UPGRADE_PIP_TO_LATEST" ]]; then
   echo "---> Upgrading pip to latest version ..."
```
torsava, line 83: Let's fix the echo to mention setuptools and wheel as well?

src/centos/macros.tpl
```diff
@@ -41,6 +41,19 @@ RUN \
...
+{% if spec.version != "2.7" %}
+    # Python 3 only code, Python 2 installs pip from PyPI in the assemble script. \
+    # We have to upgrade pip to a newer verison because: \
+    # * pip < 9 does not support different packages' versions for Python 2/3 \
+    # * pip < 19.3 does not support manylinux2014 wheels. Only manylinux2014 (and later) wheels \
+    #   support platforms like ppc64le, aarch64 or armv7 \
+    # We are newly using wheel from one of the latest stable Fedora releases (from RPM python-pip-wheel) \
+    # because it's tested better that whatever version from PyPI and contains useful patches. \
+    # We have to do it here so the permissions are correctly fixed and pip is able  to reinstall \
+    # itself in the next build phases in the assemble script if user wants the latest version \
```
torsava, line 53: >   We have to do it here so the permissions are correctly fixed and pip is able  to reinstall
>   itself in the next build phases in the assemble script if user wants the latest version

I don't understand the reasoning here. Would there really be some problem with permissions if we installed the wheel in the assemble script? I don't know of any (but I might be wrong).

I think we're installing it here during the build because it's better to have it already installed in the image — assemble script will run faster, and we can remove the original wheels directory from the image before shipping it, thus it's lighter. How about:

```
We install the wheel here during the build of the image so that the assembly script runs faster
and so that we can remove the wheel from the image after installing it.
```

torsava:
I left some more comments about the code comments, besides that the PR looks good!

src/s2i/bin/assemble
frenzymadness: Fixed.

src/centos/macros.tpl
frenzymadness: Fixed.

src/centos/macros.tpl
frenzymadness: It's not about Dockerfile vs assemble script, it's a comment describing why the installation is a part of the macro. The fact that the installation has to happen during podman build is not questionable. But the installation has to happen between the step when venv is created and the step when we set the proper permissions for the reasons described in the comment. And because the two steps are currently in this macro, the installation step has to be there as well or we have to split the macro into multiple macros.

frenzymadness:
[test][test-openshift]

frenzymadness:
@phracek could you please tell us what is wrong with the Travis CI? We are implementing kinda big changes here so I'd like to be sure that everything is okay.

frenzymadness:
[test][test-openshift]

frenzymadness:
[test] [test-openshift]

frenzymadness:
[test] [test-openshift]

frenzymadness:
Now it seems that the CI is completely broken here. @phracek do you know what is happening? Is this anyhow related to the OpenStack outage?

hhorak:
@frenzymadness -- CI depends on OpenStack, which is broken these days, so that's most likely the reason. We need to wait for PSI OpenStack getting back to better shape and re-run the CI.

frenzymadness:
[test][test-openshift]

frenzymadness:
[test][test-openshift]

frenzymadness:
[test][test-openshift]

frenzymadness:
[test][test-openshift]

frenzymadness:
[test][test-openshift]

phracek:
[test][test-openshift-4][test-openshift]

2.7/Dockerfile
phracek: I guess CentSO7 images are squashed. https://quay.io/repository/centos7/python-27-centos7?tab=builds

phracek (approved):
LGTM.

phracek:
It seems like CI is blocked. Let's try one more

[test][test-openshift]

phracek:
CentOS7 test is failing here:

```bash
17:17:50 [1m[INFO] Testing variable presence during `docker exec`[0m
17:17:50 
17:17:50  Value /opt/rh/rh-nodejs14/root/usr/share/man is missing from variable MANPATH
17:17:51 MANPATH=/opt/rh/rh-python27/root/usr/share/man:/opt/rh/python27/root/usr/share/man:/opt/rh/httpd24/root/usr/share/man:/opt/rh/rh-nodejs10/root/usr/share/man
```

I think it is related to https://github.com/sclorg/s2i-base-container/pull/206

@frenzymadness Please fix it and I think, we can merge it. As soon as it is fixed, I will execute a new set of tests and merge it.

phracek:
[test][test-openshift]

frenzymadness:
> CentOS7 test is failing here:
> 
> ```shell
> 17:17:50 �[1m[INFO] Testing variable presence during `docker exec`�[0m
> 17:17:50 
> 17:17:50  Value /opt/rh/rh-nodejs14/root/usr/share/man is missing from variable MANPATH
> 17:17:51 MANPATH=/opt/rh/rh-python27/root/usr/share/man:/opt/rh/python27/root/usr/share/man:/opt/rh/httpd24/root/usr/share/man:/opt/rh/rh-nodejs10/root/usr/share/man
> ```
> 
> I think it is related to [sclorg/s2i-base-container#206](https://github.com/sclorg/s2i-base-container/pull/206)
> 
> @frenzymadness Please fix it and I think, we can merge it. As soon as it is fixed, I will execute a new set of tests and merge it.

Good catch, thank you, should be fixed now.

[test][test-openshift]

pkubatrh:
Hmm weird that the tests are still failing. They should have the new image with node 14 from quay now...

pkubatrh:
centos7 failure should be fixed by https://github.com/sclorg/s2i-python-container/pull/431

Not sure what the centos7 openshift failure is caused by. But might be related to the CI-scripts move @phracek

frenzymadness:
The nodejs scl version should no longer be hardcoded in the Dockerfiles but the test still fails with the same error. I'll have to investigate it more.

pkubatrh:
I find it weird that the logs still have the version hardcoded:

```
11:08:53 Trying to pull repository quay.io/centos7/s2i-base-centos7 ... 
11:08:54 latest: Pulling from quay.io/centos7/s2i-base-centos7
11:08:54 Digest: sha256:fd9d486038eb0b9fc4ad2375b7c50fc10d4db639230417df499971079ad0b628
11:08:54 Status: Image is up to date for quay.io/centos7/s2i-base-centos7:latest
11:08:54  ---> cb0d45261f9b
11:08:54 Step 2/12 : EXPOSE 8080
11:09:01  ---> Running in 291c008e8a4e
11:09:01  ---> d45b685308b4
11:09:01 Removing intermediate container 291c008e8a4e
11:09:01 Step 3/12 : ENV PYTHON_VERSION 2.7 PYTHON_SCL_VERSION 27 PATH $PATH:/opt/rh/rh-python27/root/usr/bin:$HOME/.local/bin/:/opt/rh/$NODEJS_SCL/root/usr/bin:/opt/rh/httpd24/root/usr/bin:/opt/rh/python27/root/usr/bin:/opt/rh/httpd24/root/usr/sbin:/opt/rh/rh-python27/root/usr/local/bin LD_LIBRARY_PATH /opt/rh/rh-python27/root/usr/lib64:/opt/rh/$NODEJS_SCL/root/usr/lib64:/opt/rh/httpd24/root/usr/lib64:/opt/rh/python27/root/usr/lib64 LIBRARY_PATH /opt/rh/httpd24/root/usr/lib64 X_SCLS python27 MANPATH /opt/rh/rh-python27/root/usr/share/man:/opt/rh/python27/root/usr/share/man:/opt/rh/httpd24/root/usr/share/man:/opt/rh/rh-nodejs10/root/usr/share/man VIRTUAL_ENV /opt/app-root PYTHONPATH /opt/rh/$NODEJS_SCL/root/usr/lib/python2.7/site-packages XDG_DATA_DIRS /opt/rh/python27/root/usr/share:/opt/rh/rh-python27/root/usr/share:/usr/local/share:/usr/share PKG_CONFIG_PATH /opt/rh/python27/root/usr/lib64/pkgconfig:/opt/rh/httpd24/root/usr/lib64/pkgconfig:/opt/rh/rh-python27/root/usr/lib64/pkgconfig PYTHONUNBUFFERED 1 PYTHONIOENCODING UTF-8 LC_ALL en_US.UTF-8 LANG en_US.UTF-8 PIP_NO_CACHE_DIR off
```

frenzymadness:
I'm gonna run the tests locally.

pkubatrh:
Actually, there is no mention of `rm -r /opt/wheels` which got added in this PR either.... At this point I am fairly certain the CI is not testing the code in the PR.

frenzymadness:
When I inspect the image locally, the manpath is correct and the path is there (twice):
```
declare -x MANPATH="/opt/rh/rh-nodejs14/root/usr/share/man:/opt/rh/httpd24/root/usr/share/man:/opt/rh/rh-python27/root/usr/share/man:/opt/rh/python27/root/usr/share/man:/opt/rh/httpd24/root/usr/share/man:/opt/rh/rh-nodejs14/root/usr/share/man"
```

> Actually, there is no mention of `rm -r /opt/wheels` which got added in this PR either.... At this point I am fairly certain the CI is not testing the code in the PR.

Which is awesome, isn't it?

phracek:
OpenShift tests are failing because OpenShift 3 does not start.

CentOS7 failed with this reason.

```bash
11:18:06 [1m[INFO] Running test test_scl_variables_in_dockerfile ... [0m
11:18:06 
11:18:06 
11:18:06 [1m[INFO] Testing variable presence during `docker exec`[0m
11:18:06 
11:18:06  Value /opt/rh/rh-nodejs14/root/usr/share/man is missing from variable MANPATH
11:18:07 MANPATH=/opt/rh/rh-python27/root/usr/share/man:/opt/rh/python27/root/usr/share/man:/opt/rh/httpd24/root/usr/share/man:/opt/rh/rh-nodejs10/root/usr/share/man
11:18:07 
```

phracek:
I will try it after the weekend. Maybe the test is wrong. I will debug it more.

frenzymadness:
[test][test-openshift]

frenzymadness:
> I will try it after the weekend. Maybe the test is wrong. I will debug it more.

Any progress?

phracek:
[test]

phracek:
[test]

phracek:
Let's test OpenShift.

[test-openshift]

phracek:
CentOS CI is fixed by https://github.com/sclorg/rhscl-container-ci/pull/76

frenzymadness:
Ok, this is getting bigger and more complicated than anticipated and needed. I'm gonna split it up into pieces. Here is my plan:
1. I'm gonna close this PR.
2. I'm gonna open a new one just for the fix of the hardcoded nodejs scl version. This PR should have all CI jobs green so we can then rebase the rest on top of it.
3. I'm gonna open a new PR for this change with the same code but one small difference - because the Fedora 34 is out, I can update the pip wheel to 21.0.1 which is the latest released version in the latest stable fedora.

PR #414, opened 2020-11-26
Experiment with different docker context and symlinks

frenzymadness:


frenzymadness:
[test][test-openshift]

PR #415, opened 2020-12-01
Update usage script

fila43:


fila43:
[test]

frenzymadness (changes requested):
I'd rather mention the documentation first and then the specific case for OpenShift users. Something like:
```
There are multiple ways how to use this image documented in …

To run this image in OpenShift, use the oc command:
…
```

Does it make sense? I think that not everybody knows `oc` command.

I'm gonna investigate the tests failures, it seems to be something with upstream wheels/pip.

fila43:
> I'd rather mention the documentation first and then the specific case for OpenShift users. Something like:
> 
> ```
> There are multiple ways how to use this image documented in …
> 
> To run this image in OpenShift, use the oc command:
> …
> ```
> 
> Does it make sense? I think that not everybody knows `oc` command.
> 
> I'm gonna investigate the tests failures, it seems to be something with upstream wheels/pip.

Yea, good idea

frenzymadness:
Tests are fixed in #417 , please rebase this PR along the way.

fila43:
[test]

frenzymadness (approved):
Good job! Thank you!

PR #416, opened 2020-12-02
Deprecate Python 3.7

frenzymadness:
Python 3.7 was available only for Fedora 31 which is now EOL.


frenzymadness:
[test][test-openshift]

torsava:
Everything looks good, thanks!

frenzymadness:
Rebase should fix tests. [test][test-openshift]

PR #417, opened 2020-12-02
Fix compatibility with latest pip (20.3)

frenzymadness:
Old resolver reports `up-to-date` while the new one (default) reports `satisfied`.


frenzymadness:
[test][test-openshift]

torsava:
LGTM, local testing looks goo too!

PR #418, opened 2020-12-11
Document and test direct Dockerfile usage without s2i scripts

frenzymadness:
SSIA

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -106,17 +106,43 @@ To use the Source-to-Image scripts and build an image using a Dockerfile, create
 
 ```
 FROM registry.access.redhat.com/ubi8/python-27
+
 # Add application sources to a directory that the assemble script expects them
 # and set permissions so that the container runs without root access
 USER 0
 ADD app-src /tmp/src
-RUN chown -R 1001:0 /tmp/src
+RUN /usr/bin/fix-permissions /tmp/src
 USER 1001
+
 # Install the dependencies
 RUN /usr/libexec/s2i/assemble
+
 # Set the default command for the resulting image
 CMD /usr/libexec/s2i/run
 ```
+
+If you decide not to use the Source-to-Image scripts, you will need to manually tailor the Dockerfile to your application and its needs.
+Example Dockerfile for a simple Django application:
+
+```
+FROM registry.access.redhat.com/ubi8/python-27
+
+# Add application sources with correct permissions for OpenShift
+USER 0
+ADD app-src .
+RUN chown -R 1001:0 ./
+USER 1001
+
+# Install the dependencies
+RUN pip install -U "pip>=19.3.1" && \
+    pip install -r requirements.txt && \
+    python manage.py collectstatic --noinput && \
+    python manage.py migrate
+
+# Run the application
+CMD python manage.py runserver 0.0.0.0:8080
+```
+
 #### 4. Build a new image from a Dockerfile prepared in the previous step
 
 ```
--- a/src/README.md
+++ b/src/README.md
@@ -106,14 +106,17 @@ To use the Source-to-Image scripts and build an image using a Dockerfile, create
 
 ```
 FROM {{ spec.main_image }}
+
 # Add application sources to a directory that the assemble script expects them
 # and set permissions so that the container runs without root access
 USER 0
 ADD app-src /tmp/src
 RUN chown -R 1001:0 /tmp/src
 USER 1001
+
 # Install the dependencies
 RUN /usr/libexec/s2i/assemble
+
 # Set the default command for the resulting image
 CMD /usr/libexec/s2i/run
 ```
--- a/src/README.md
+++ b/src/README.md
@@ -117,6 +117,29 @@ RUN /usr/libexec/s2i/assemble
 # Set the default command for the resulting image
 CMD /usr/libexec/s2i/run
 ```
+
+If you decide not to use the Source-to-Image scripts, you will need to manually tailor the Dockerfile to your application and its needs.
+Example Dockerfile for a simple Django application:
+
+```
+FROM {{ spec.main_image }}
+
+# Add application sources with correct permissions for OpenShift
+USER 0
+ADD app-src .
+RUN chown -R 1001:0 ./
+USER 1001
+
+# Install the dependencies
+RUN pip install -U "pip>=19.3.1" && \
+    pip install -r requirements.txt && \
+    python manage.py collectstatic --noinput && \
+    python manage.py migrate
+
+# Run the application
+CMD python manage.py runserver 0.0.0.0:8080
+```
+
 #### 4. Build a new image from a Dockerfile prepared in the previous step
 
 ```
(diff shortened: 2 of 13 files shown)


frenzymadness:
[test][test-openshift]

src/README.md
```diff
@@ -117,6 +117,27 @@ RUN /usr/libexec/s2i/assemble
 # Set the default command for the resulting image
 CMD /usr/libexec/s2i/run
 ```
+
+A content of a Dockerfile without Source-to_image script would highly depends on your application and its needs.
```
frenzymadness, line 121:
```suggestion
Content of a Dockerfile without Source-to-image script would highly depend on your application and its needs.
```

frenzymadness:
The logs look good to me so I think this is ready for a review. Cc @hhorak @torsava @phracek

src/README.md
```diff
@@ -117,6 +117,27 @@ RUN /usr/libexec/s2i/assemble
...
+An example for simple Django application might look like this:
```
torsava, line 122:
```suggestion
Example Dockerfile for a simple Django application:
```

src/README.md
torsava:
```suggestion
If you decide not to use the Source-to-Image scripts, you will need to manually tailor the Dockerfile to your application and its needs.
```

src/README.md
```diff
@@ -117,6 +117,27 @@ RUN /usr/libexec/s2i/assemble
...
+
+```
+FROM {{ spec.main_image }}
+
+# Add application sources
+USER 0
```
torsava, line 128: Shouldn't we switch to `USER 1001` after copying and `chown`ing the sources, like in the previous example? AFAIK it's good practice not to run containers with `UID 0` inside, and we don't really need root, we can install the deps as `--user`. What do you think, @frenzymadness ? On the other hand, maybe it would be too much hassle to set up the priviledges?

torsava:
Left some questions / comments, the rest looks good!

src/README.md
frenzymadness: That makes perfect sense, thank you! I'm gonna implement it.

frenzymadness:
[test][test-openshift]

2.7/README.md
```diff
@@ -117,6 +117,29 @@ RUN /usr/libexec/s2i/assemble
...
 ```
+
+If you decide not to use the Source-to-Image scripts, you will need to manually tailor the Dockerfile to your application and its needs.
+Example Dockerfile for a simple Django application:
+
+```
+FROM registry.access.redhat.com/ubi8/python-27
+
+# Add application sources
+USER 0
```
hhorak, line 128: Although we don't have it documented in other images either, we might document here why we change USER to 0 and group for the app files -- that it allows to run the resulting image as any user in OpenShift.

2.7/README.md
```diff
@@ -117,6 +117,29 @@ RUN /usr/libexec/s2i/assemble
 # Set the default command for the resulting image
 CMD /usr/libexec/s2i/run
 ```
+
+If you decide not to use the Source-to-Image scripts, you will need to manually tailor the Dockerfile to your application and its needs.
```
hhorak, line 121: What about to stay more direct here, like:
"To use your own Dockerfile for building a new image without the Source-to-Image scripts, use `pip install --user` for installing dependencies. Example Dockerfile for a simple Django application:"

src/README.md
frenzymadness: I've realized that our images already have an active virtual environment so we don't have to install with `--user` but the switch to 1001 still makes sense.

frenzymadness:
[test][test-openshift]

2.7/README.md
frenzymadness: I'd say that there are many more possibilities like installing system packages, compiling your own tools etc. So I'd stay general. If you need to do this manually, you have to have a good reason and knowledge to do it. What do you think?

2.7/README.md
frenzymadness: During testing, I've just realized that we don't need to switch users to 0 and back at all because all we do has an impact in user-writable folders in the virtual environment.

What are you probably talking about is documented [here](https://github.com/sclorg/s2i-python-container/blob/master/src/Dockerfile.template#L85) and it's achieved by our scripts so we either skip this step here as not needed or add and document it.

frenzymadness: I'm gonna try it without it and we'll what openshift tests will say,

frenzymadness:
[test][test-openshift]

src/README.md
frenzymadness: Or maybe not. From what I see in my local testing, switching user to 0 and back might not be necessary at all.

2.7/README.md
pkubatrh: I am not sure I follow here. The `USER 0` was needed so that the files copied into the new layer have group set to `0` (since Openshift is run with a random uid user that is member of the 0 group).

>What are you probably talking about is documented here and it's achieved by our scripts so we either skip this step here as not needed or add and document it.

Which scripts make this not an issue for the Dockerfile example?

pkubatrh: That said, would this (non-s2i) example even be aimed at Openshift use?

frenzymadness:
[test][test-openshift]

frenzymadness:
@torsava @hhorak so, the Dockerfile with s2i scripts is almost identical to what `s2i build --as-dockerfile` produces so I consider it as correct. I've documented the fix-permission usage but it's not included in tests because there is no reason to use it there.

2.7/README.md
frenzymadness: Do I understand it correctly that we have to switch to user 0, copy files, change ownership and switch batch to user 1001 in all our examples? Is the fix-permissions needed here after the commands?

frenzymadness:
Tests in OS are failing because `toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading`.

frenzymadness:
I've kinda lost my context here but we need to finish this soon because the example is needed for the documentation team. @pkubatrh @hhorak could you please take one more look? I think that the only unsolved problem is the user switching in the dockerfile.

hhorak:
I played with this a bit myself yesterday and it looks all good now, I'm +1 for merging as it is. Thank!

frenzymadness:
> I played with this a bit myself yesterday and it looks all good now, I'm +1 for merging as it is. Thank!

The only issue I'm afraid of is something with access rights. Have you tried only read-only applications or also something more complex?

frenzymadness:
[test][test-openshift]

frenzymadness:
So, thanks to @pkubatrh, I understand why we need to change user to 0 and back to 1001 in all our examples. The change to user 0 guarantees that also group 0 has access to all the files and therefore in OpenShift, where we have random user UIDs, the sources are still readable.

One more change here is that in the dockerfile with s2i scripts we can use fix-permissions script to show its power and on the other hand, in the manual dockerfile, we want to show how to change permissions manually.

torsava:
@frenzymadness Both the commits and the final version looks good to me!

PR #422, opened 2021-01-08
Recommend ADD --chown 1001:0 syntax if possible

hhorak:
Unfortunately, --chmod option was added into docker in v17, so it does not work
with docker we have in RHEL-7.

For supporting docker with our example Dockerfiles, we can use the older way,
but in README.md, where we recommend using podman (that supports the --chown option),
we can already use the newer shorter syntax that also does not need changing
USER to 0 that might be prohibitted.

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -108,10 +108,13 @@ To use the Source-to-Image scripts and build an image using a Dockerfile, create
 FROM registry.access.redhat.com/ubi8/python-27
 # Add application sources to a directory that the assemble script expects them
 # and set permissions so that the container runs without root access
-USER 0
-ADD app-src /tmp/src
-RUN chown -R 1001:0 /tmp/src
-USER 1001
+# With older docker that does not support --chown option for ADD statement,
+# use these statements instead:
+#  USER 0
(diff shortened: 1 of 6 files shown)


frenzymadness:
Honestly, I'm not sure about this. I'd rather keep the Dockerfiles we have in readme and the ones we test in sync so we can be sure that we are providing examples that actually work on our platforms.

It might be also possible to do both and let distgen to generate old files for old platforms and new files for new platforms. It would make things more complex so we should discuss whether it's worth it.

torsava:
> Honestly, I'm not sure about this. I'd rather keep the Dockerfiles we have in readme and the ones we test in sync so we can be sure that we are providing examples that actually work on our platforms.
> 
> It might be also possible to do both and let distgen to generate old files for old platforms and new files for new platforms. It would make things more complex so we should discuss whether it's worth it.

I agree I'd keep the Dockerfile as is for RHEL platforms. And I think it's not worth generating both in distgen, not worth the complexity. It's hard to read as it is :smile:

frenzymadness:
Feel free to reopen this if you want.

PR #425, opened 2021-02-01
Fix x scl name

phracek:
This pull request fixes a typo in Dockerfile.rhel7.

Collection `rh-python27` does not exists. But python27 yes.


phracek:
[test]

frenzymadness (changes requested):
You should run distgen when you change source files. However, the failure in the tests is caused by something else and I know what is the cause, I'll investigate it tomorrow.

phracek:
[test]

> You should run distgen when you change source files. However, the failure in the tests is caused by something else and I know what is the cause, I'll investigate it tomorrow.

The second commit is after dist-gen generated sources ;) The only `2.7/Dockerfile.rhel7` was changed. Which is true.

frenzymadness:
> The second commit is after dist-gen generated sources ;) The only `2.7/Dockerfile.rhel7` was changed. Which is true.

As far as I see in my local copy of this PR, distgen also regenerates `2.7/Dockerfile` for Centos 7. Or do I misunderstand something?

frenzymadness:
Could you please rebase this PR to fix the CI and then run make clean-versions && make generate-all?

frenzymadness:
BTW I see that this is already fixed downstream. At least for python 2 container,

phracek:
The downstream already contains a fix. I have rebased and regenerated content.
But for some reason these Dockerfiles are missing:

```bash
2.7/Dockerfile
3.8/Dockerfile
3.8/Dockerfile.fedora
3.9/Dockerfile.fedora
```

phracek:
[test]

frenzymadness (changes requested):
There must be something wrong in your setup or somewhere else. The change you have made to `src/Dockerfile.template` should be reflected in `2.7/Dockerfile.rhel7` and `2.7/Dockerfile` but your commit with generated files contains only the first mentioned file.
When I check out this PR and run `make clean-versions && make generate-all` I see the uncommited change in the Dockerfile for Centos in 2.7 folder.

phracek:
[test]

frenzymadness (approved):
Thank you!

PR #426, opened 2021-02-02
Fix pip version for python 2 in tests

frenzymadness:
pip 21 does not support python 2 anymore

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -134,7 +134,7 @@ RUN chown -R 1001:0 ./
 USER 1001
 
 # Install the dependencies
-RUN pip install -U "pip>=19.3.1" && \
+RUN pip install -U "pip>=19.3.1,<21.0" && \
     pip install -r requirements.txt && \
     python manage.py collectstatic --noinput && \
     python manage.py migrate
--- a/2.7/test/from-dockerfile
+++ b/2.7/test/from-dockerfile
@@ -1,1 +1,0 @@
-../../src/test/from-dockerfile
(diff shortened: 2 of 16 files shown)


frenzymadness:
[test] [test-openshift]

frenzymadness:
There seems to be an unrelated problem in OpenShift. @phracek could you please take a look?

@torsava review of this change appreciated although it's only about tests.

torsava:
> @torsava review of this change appreciated although it's only about tests.

@frenzymadness It's a good change for python27, absolutely. But I'm worried that adding it also to python38/python39 README files is very confusing. When I would be reading the README I would be asking myself why do I have to use an old version of pip.

frenzymadness:
> @frenzymadness It's a good change for python27, absolutely. But I'm worried that adding it also to python38/python39 README files is very confusing. When I would be reading the README I would be asking myself why do I have to use an old version of pip.

You are right. I'm gonna try to fix it. It'd require to switch from a symlink to the dockerfile to a full copy of the file generated by the distgen to have a different version for Python 2 and Python 3 but it'll mean that we actually test the exact same version we have in readmes so it's worth it.

frenzymadness:
[test] [test-openshift]

frenzymadness:
Fixed!

* test/from-dockerfile (folder) is no longer a symlink
* test/from-dockerfile/Dockerfile.tpl is now a symlink (simply because it's the same for all versions)
* test/from-dockerfile/Dockerfile_no_s2i.tpl is now generated via distgen to have exactly the same content as the readme

torsava (approved):
Very nice solution. LGTM!

frenzymadness:
Thanks!

PR #431, opened 2021-04-16
rhel7: do not hardcode nodejs SCL version

pkubatrh:


pkubatrh:
[test][test-openshift]

frenzymadness:
See the last two commits in #412 . I did the fix before you commented for the first time there today.

pkubatrh:
Ah right, sorry for the duplication. I only checked the centos7 logs and saw the nodejs10 definition was still present there.

PR #451, opened 2021-06-01
add buildpack metadata to 3.9 RHEL8 image

lance:
Similar to the work that was done in the Node.js images, this commit adds
buildpack metadata to the RHEL8 Dockerfile so that it can be used as a
"stack" when creating builder images. The OpenShift Serverless Functions
team is looking into using these images for our stacks, and these metadata
bits are a requirement to enable that.

See: https://github.com/sclorg/s2i-nodejs-container/pull/273
See: https://github.com/boson-project/buildpacks/tree/main/stacks/python

Signed-off-by: Lance Ball <lball@redhat.com>

### DIFF
--- a/3.9/Dockerfile.rhel8
+++ b/3.9/Dockerfile.rhel8
@@ -10,7 +10,11 @@ ENV PYTHON_VERSION=3.9 \
     PYTHONIOENCODING=UTF-8 \
     LC_ALL=en_US.UTF-8 \
     LANG=en_US.UTF-8 \
-    PIP_NO_CACHE_DIR=off
+    PIP_NO_CACHE_DIR=off \
+    CNB_STACK_ID=com.redhat.stacks.ubi8-python-39 \
+    CNB_USER_ID=1001 \
+    CNB_GROUP_ID=0
+
 
 # RHEL7 base images automatically set these envvars to run scl_enable. RHEl8
 # images, however, don't as most images don't need SCLs any more. But we want
@@ -35,6 +39,7 @@ LABEL summary="$SUMMARY" \
       io.k8s.display-name="Python 3.9" \
       io.openshift.expose-services="8080:http" \
       io.openshift.tags="builder,python,python39,python-39,rh-python39" \
+      io.buildpacks.stack.id="com.redhat.stacks.ubi8-python-39" \
       com.redhat.component="python-39-container" \
       name="ubi8/python-39" \
       version="1" \


frenzymadness:
Hello, thanks for the PR. I'm gonna take a look as soon as possible but I have to reimplement it because the repository and the folders for each Python version are generated by distgen from src dir and multispec.yaml.

Do I understand correctly that we need this metadata only for Python 3.9 on UBI 8?

Could you please point me to some documentation about stacks? I've been working with buildpacks, builders, pack etc. but I don't understand how multiple stacks can be interconnected to a single universal builder.

lance:
> Could you please point me to some documentation about stacks? I've been working with buildpacks, builders, pack etc. but I don't understand how multiple stacks can be interconnected to a single universal builder.

Sure. :smiley: https://buildpacks.io/docs/operator-guide/create-a-stack/

The relationship between stacks and builders is 1:1 at the moment. This is, in many ways, the root of our problems with how Builders work, and what "stack packs" are meant to address. See: https://github.com/buildpacks/rfcs/blob/main/text/0069-stack-buildpacks.md. In the "correct" scenario, there would be a single UBI-based builder/stack combination, and each runtime buildpack would be responsible for adding the appropriate language at build time. Since we install our languages with RPMs, we can't do this (it is not possible to modify system files during the build phase). When RFC-0069 has been implemented, we should be able to do something similar - effectively allowing multiple stacks with a single builder.

> Do I understand correctly that we need this metadata only for Python 3.9 on UBI 8?

AFAIK Functions are the only team at RHT currently needing a builder/stack supporting Python. Our current needs are only Python 3.9. But since these changes are just labels and environment variables, it might be simple enough to add them to all versions - especially if these Dockerfiles are all generated. :shrug:

frenzymadness:
Thanks for the info. After carefully reading the RFC for the third time :smiley: I can imagine that it might work like this:
1. We will have one builder with one stack, both universal providing only a minimal set of common packages.
2. Our universal stack will contain a set of stackpacks - for example, one universal capable to install any RPM package (like the universal apt stackpack in the examples of the RFC), one specific handling Python packages etc, and each stackpack will provide its content and capabilities.
3. Buildpacks then can require a set of stackpacks to use before they start with the build. So buildpack handling Python application would require Python stackpack to make sure the packages are there and it can install the app.

Sorry for asking so many unrelated questions but you seem to have a lot of information and knowledge I need to make the ecosystem clear to me.

lance:
> Sorry for asking so many unrelated questions but you seem to have a lot of information and knowledge I need to make the ecosystem clear to me.

No worries! I think your assessment of items 1-3 is spot on. I'm somewhat new to buildpacks as well, but have been mucking around with them for a year or so. Whatever limited knowledge I have, I'm happy to share.

frenzymadness:
I'm gonna try to implement it in #452

PR #455, opened 2021-06-14
Fix common scripts source

frenzymadness:
In the last PR, I've accidentally merged a change of a source branch for the common scripts. The changes are proposed in https://github.com/sclorg/container-common-scripts/pull/199 but we need to switch back to the original source (master branch).


frenzymadness:
[test]

PR #457, opened 2021-06-25
Add runs-on in ubuntu-20.04

phracek:


phracek:
This does not modify s2i-python-container. Merging

PR #462, opened 2021-06-29
This update was tested on phracek/s2i-python-container repo

phracek:


phracek:
This two files were tested directly on phracek/s2i-python-container.

PR #467, opened 2021-07-28
Standalone minimal container

frenzymadness:
This is the first implementation of the minimal Python container image which keeps the assemble script and therefore can be used as a build and run image as we do for the full container image.

The size is around ~ **680 MB** smaller than the full image.

See the individual commits - they contain changes I have to make to the initial copy of the full 3.9 image.

The biggest open question is: How to maintain it? Given the fact that all other images are generated via distgen, we should use it here as well, but there are some differences that might make it hard. Another possibility is to maintain some kind of transform script that would transform the full image to the minimal one but this might be even harder than the first idea and would probably require some patching and scripting.

ToDo:
- [x] Update readme and document all the limitations
- [x] Remove tests that are currently failing. The failing tests use mod_wsgi or uwsgi which are not installable from the source in the minimal image (missing build dependencies) but they are testing more than just installation so the other parts have to be transferred somewhere else.
- [x] Remove tests of django-example. Django-ex depends on psycopg-binary which is not installable from sources.
- [x] Implement, document, and add some tests for the use case, where we build an app on top of the full image and then move it to the minimal image. These tests will cover the mod_wsgi and uwsgi packages because they are installable from the source on top of the full image.
- [x] Implement, document, and add some tests for the use case, where instead of using the full image as a build image user decides to adjust their dockerfile and install some dependencies manually.


frenzymadness:
Tests news:

- npm-virtualenv-uwsgi is now just uwsgi and tests the one-stage build from dockerfile
- micropipenv-requirements tests also the one-stage build from a dockerfile but with different dependencies for mod_wsgi
- mod_wsgi has been removed

Readme now covers both ways how to work around the limitations of the minimal container image. The first way is also tested but I don't see a way how to test the second one without pulling one extra image (the full one) from the registry.

I also still don't know how to maintain this folder in this repository.

phracek:
In order to test 3.9-minimal image add `3.9-minimal` here: https://github.com/sclorg/s2i-python-container/blob/master/Makefile#L3

frenzymadness:
I'm proposing some changes to common scripts and I'm testing them here and I'm also trying to find a way how to make the CI run while skipping distgen for 3.9-minimal.

frenzymadness:
Well, at least it no longer breaks the standard tests. Let's run the rest [test]

3.9-minimal/Dockerfile.rhel8
```diff
@@ -1,19 +1,25 @@
...
     CNB_STACK_ID=com.redhat.stacks.ubi8-python-39 \
     CNB_USER_ID=1001 \
     CNB_GROUP_ID=0 \
-    PIP_NO_CACHE_DIR=off
+    PIP_NO_CACHE_DIR=off \
+    # The following variables are usually available from parent s2i images \
+    STI_SCRIPTS_PATH=/usr/libexec/s2i \
+    APP_ROOT=/opt/app-root \
+    HOME=/opt/app-root/src \
+    PLATFORM="el8"
```
torsava: Just a nitpick thought: It might be good to define the commn s2i variables on top and use them in the PATH definition.

torsava:
@frenzymadness  Thank you for the copy commits, it's so much easier to review when I only need to check the diff!

frenzymadness:
The failure of RHEL8 was caused by the fact that we are newly testing Python 3.9 with Django example branch 2.2.x where Django depends on psycopg2-binary version 2.8.5 but this version has not wheels for Python 3.9 so I have to update it to at least version 2.8.6. I'm gonna test this update here and then propose it in the django-ex project.

frenzymadness:
[test]

3.9-minimal/Dockerfile.rhel8
```diff
@@ -41,33 +47,38 @@ LABEL summary="$SUMMARY" \
...
 # Python 3 only
 # Yes, the directory below is already copied by the previous command.
 # The problem here is that the wheels directory is copied as a symlink.
 # Only if you specify symlink directly as a source, COPY copies all the
 # files from the symlink destination.
 COPY 3.9/root/opt/wheels /opt/wheels
+
+# Also creates the directory - it has to be done here so the latter
+# fix-permissions fixes this directory as well.
+WORKDIR ${HOME}
```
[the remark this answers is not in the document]
frenzymadness: Because it not just set the directory as your workdir but _also_ creates it if it does not exists yet. For me, it was new and I have to move the line before fix-permissions so I've also documented the reason. Should I somehow rephrase it?

3.9-minimal/s2i/bin/run
```diff
@@ -17,7 +17,9 @@ function should_migrate() {
 
 # Guess the number of workers according to the number of cores
 function get_default_web_concurrency() {
-  limit_vars=$(cgroup-limits)
+  # Using python and just python here because the script has
+  # platform-python in its shebang which we don't have in the minimal image
+  limit_vars=$(python /usr/bin/cgroup-limits)
```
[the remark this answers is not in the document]
frenzymadness: python3.9 is the same as python here because the main virtual environment is already activated. This way, it's more universal and less error-prone but it will be generated in the future hopefully so the second argument does not matter that much. Should I change it?

3.9-minimal/README.md
```diff
@@ -1,117 +1,61 @@
...
+means that the resulting containers it produces consume a lot of disk space. This is caused mainly by the fact that the image contains
+npm, compilers, header files and some other packages one might need to install and deploy their applications.
 
-```
-podman pull registry.access.redhat.com/ubi8/python-39
-```
+Because size does matter for us and our customers, we have prepared this minimal container image with very limited subset
+of installed packages. There are no compilers, no header files, no npm etc and dnf tool is replaced with minimalistic
+reimplementation called microdnf making the resulting container images much smaller. It also causes some limitations
+but we also provide ways how to workaround them.
```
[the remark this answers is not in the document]
frenzymadness: Really yum? I mean, we are creating something new and we are building it on top of RHEL8 where we already have dnf so mentioning yum here seems to me like skipping one part of the yum - dnf - microdnf chain. I also think that the minimal container and containers, in general, are for modern people not for the ones who still use RHEL 5/6/7.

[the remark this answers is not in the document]
torsava: There's been a marketing decision that the package manager in RHEL is supposed to be called `yum` even in RHEL 8, so that customers don't feel like they have to "switch" in some way. It's just a symlink to `dnf` and works the same, but `yum` is the official name.

See e.g. our official documentation: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_basic_system_settings/assembly_installing-and-using-python_configuring-basic-system-settings

3.9-minimal/s2i/bin/run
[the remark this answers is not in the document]
torsava: [Facepalm] Of course, I forgot we're in the virtual environment here. Makes sense, don't change anything.

frenzymadness:
The current problem in the CI is that the new minimal container image has the same name in the Dockerfile which causes it to overwrite the full one during tagging. Fix is on the way.

frenzymadness:
[test]

3.9-minimal/Dockerfile.rhel8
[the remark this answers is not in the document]
torsava: Oooh, now I understand. I thought the "also" referred to some previous commands, or something like that.

How about "This command sets (and also creates if necessary) the home directory", or something like that?

3.9-minimal/README.md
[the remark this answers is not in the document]
frenzymadness: Oukej, oukej, old man :smile:

3.9-minimal/Dockerfile.rhel8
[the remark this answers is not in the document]
frenzymadness: Fixed, thank you.

frenzymadness:
We are closer and closer to happy CI. [test]

frenzymadness:
[test]

frenzymadness:
[test]

frenzymadness:
[test]

frenzymadness:
CI status:
* Centos 7 and Fedora - CWT plan failed but all tests passed
* Centos 7 OS 3 - the failure there does not seem to be related to this PR (Cc @phracek)

What has to be done before we merge this:

- [ ] django-ex needs newer psycopg2-binary which provides Python 3.9 wheels: https://github.com/sclorg/django-ex/pull/185
- [ ] common scripts needs to support skipping a version in distgen: https://github.com/sclorg/container-common-scripts/pull/208
- [ ] test from dockerfile needs to support local application: https://github.com/sclorg/container-common-scripts/pull/207
- [ ] after all above, we can rebase this PR, switch common submodule to the master branch and remove the one commit editing the django-ex git repo URL.

frenzymadness:
But, all the above tasks are needed only for CI, so @pkubatrh if you want to or need to, you can take the image and build it in downstream.

3.9-minimal/README.md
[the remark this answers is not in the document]
torsava: Things should stay just as I like them! :rofl:

3.9-minimal/Dockerfile.rhel8
torsava: @frenzymadness What do you think?

frenzymadness: Implemented in the latest commit.

frenzymadness:
[test]

frenzymadness:
[test]

frenzymadness:
[test]

frenzymadness:
[test]

frenzymadness:
[test]

frenzymadness:
[test]

frenzymadness:
@torsava take a look on the latest reverted commit. I had to revert the change because it seems that ENV cannot substitute env variables defined in the same command. There are two possible ways to do it:
1. Leave it as is. Not so nice but similar to what we have in other containers. Or
2. Set only the PATH in a separated ENV call where it should also work with variables from the previous call. This is more readable but it also creates one more (empty) layer in the resulting image.

torsava:
> @torsava take a look on the latest reverted commit. I had to revert the change because it seems that ENV cannot substitute env variables defined in the same command. There are two possible ways to do it:
> 
> 1. Leave it as is. Not so nice but similar to what we have in other containers. Or
> 2. Set only the PATH in a separated ENV call where it should also work with variables from the previous call. This is more readable but it also creates one more (empty) layer in the resulting image.

I have a slight preference for option 2:
- It would be overall more similar to how normal s2i image works (i.e. 2 ENV definitions, 1 for base, 1 for python)
- The code would look cleaner (visually separated ENV definitions for base and for python)
- I don't think we care much about an empty layer in the resulting image

However, given the limitation, option 1 is also valid, so feel free to proceed how you deem best.

frenzymadness:
[test]

frenzymadness:
Thanks, let's give it a try. You are right that we already have more ENV calls than we really need so one more should be fine.

frenzymadness:
@torsava also, if you have some free cycles, please take a look at the PRs mentioned above. There are also some very simple ones in https://github.com/sclorg/django-ex/pulls

3.9-minimal/test/run
```diff
@@ -199,7 +199,12 @@ test_application() {
   # that the container starts at all
   check_result $?
 
-  test_scl_usage "python --version" "Python $VERSION." "${cid_file}"
+  # Instead of relying on VERSION variable coming from Makefile
+  # set the expected string based on the PYTHON_VERSION defined
+  # inside the running container.
+  python_version=$(docker run --rm $IMAGE_NAME /bin/bash -c "echo \$PYTHON_VERSION" 2>&1)
```
torsava, line 205: Above in this script there is `if [[ -z $VERSION ]]; ... exit 1`, so why can't we rely on `VERSION`?

In addition, if there is a reason to detect the `python_version` like this, wouldn't it be better to do it outside of `test_application` so it runs only once?

3.9-minimal/test/run
```diff
@@ -220,7 +220,8 @@ test_from_dockerfile(){
   check_result $?
 
   # mod_wsgi in micropipenv-requirements-test-app
-  sed "s@#IMAGE_NAME#@${IMAGE_NAME}@" $test_dir/from-dockerfile/mod_wsgi.Dockerfile.tpl > $test_dir/from-dockerfile/Dockerfile
+  # TODO: Find a better way how to specify full/minimal images that should work together
+  sed "s@#IMAGE_NAME#@${IMAGE_NAME}@;s@#FULL_IMAGE_NAME#@ubi8/python-39:1@" $test_dir/from-dockerfile/mod_wsgi.Dockerfile.tpl > $test_dir/from-dockerfile/Dockerfile
```
torsava: This seems a bit hardcoded. How about creating a `FULL_IMAGE_NAME` variable similar to `IMAGE_NAME`?

3.9-minimal/test/run
```diff
@@ -215,8 +215,7 @@ test_from_dockerfile(){
   info "Test from Dockerfile"
 
   # uwsgi in uwsgi-test-app
-  sed "s@#IMAGE_NAME#@${IMAGE_NAME}@" $test_dir/from-dockerfile/uwsgi.Dockerfile.tpl > $test_dir/from-dockerfile/Dockerfile
-  ct_test_app_dockerfile $test_dir/from-dockerfile/Dockerfile $test_dir/uwsgi-test-app 'Hello World from uWSGI hosted WSGI application!' app-src
+  ct_test_app_dockerfile $test_dir/from-dockerfile/uwsgi.Dockerfile.tpl $test_dir/uwsgi-test-app 'Hello World from uWSGI hosted WSGI application!' app-src
```
torsava: Why is this `sed` removed? I don't see where it's replaced by anything else.

frenzymadness:
[test]

3.9-minimal/test/run
frenzymadness: $VERSION works only if it is 3.7 or 3.8 but not for 3.9-minimal because then the $VERSION is not a part of `python -V` output. That's the main reason. Now the test also tests that the env variable is correct in the container so I don't think it's bad.

3.9-minimal/test/run
frenzymadness: It would be nice to have something like that in the future but I don't see an easy way how to do it now. How to find a proper full-size container if you only know the name of the minimal one? Some regex or replace might work in the future and it will be a part of the distgen implementation. It can then also be implemented as a distgen variable that every minimal image will have the name of the full-size image defined manually.

3.9-minimal/test/run
frenzymadness: Take a look here, it's already implemented as a part of the function: https://github.com/sclorg/container-common-scripts/blob/47b8b9cfe459a1bba3b04c44cc7da8519a3c520a/test-lib.sh#L863

frenzymadness: My plan is to change the function to sed the image name to the Dockerfile only if there is a placeholder for that and do nothing otherwise. That way, you can say whether the Dockerfile is ready for use or it needs the name of the image currently being tested.

But such change needs to check all the images where this function is already in use.

frenzymadness:
I did a rebase, removed all temporary commits, and squashed some of them. This should be ready for merge when we resolve all the open conversations. I'm gonna prepare a fix for the dirty from/FROM hack until I lose the context of it but it'll be cross-repo change so it might take some time.
[test]

frenzymadness:
The failure of the RHEL 8 job does not seem to be related to this PR. [test]

frenzymadness:
CI is completely green! :tada: @torsava wanna do a sanity check? We can merge this and I can then submit other PR to address my ToDo records.

phracek:
[test]

3.9-minimal/test/run
torsava: Ok, sounds good.

3.9-minimal/test/run
torsava: I like the distgen variable approach. I see there's a TODO note for the future, so I'm fine with leaving it as it is for now.

3.9-minimal/test/run
torsava: Ah makes sense, and I agree that would be a good addition for the future.

3.9-minimal/Dockerfile.rhel8
```diff
@@ -88,8 +99,7 @@ RUN \
     ${APP_ROOT}/bin/pip install /opt/wheels/pip-* && \
     rm -r /opt/wheels && \
     chown -R 1001:0 ${APP_ROOT} && \
-    fix-permissions ${APP_ROOT} -P && \
-    rpm-file-permissions
```
torsava: `rpm-file-permissions` are not needed anymore for some reason?

3.9-minimal/README.md
```diff
@@ -1,117 +1,61 @@
...
-The RHEL images are available in the [Red Hat Container Catalog](https://access.redhat.com/containers/),
-the CentOS images are available on [Quay.io](https://quay.io/organization/centos7),
-and the Fedora images are available in [Fedora Registry](https://registry.fedoraproject.org/).
-The resulting image can be run using [podman](https://github.com/containers/libpod) or
-[docker](http://docker.io).
+This container image is a special version of the [full Python 3.9 container image](https://github.com/sclorg/s2i-python-container/tree/master/3.9)
+provided as a [S2I](https://github.com/openshift/source-to-image) base image for your Python 3.9 applications.
 
-Note: while the examples in this README are calling `podman`, you can replace any such calls by `docker` with the same arguments
+Because the minimal and full images work simillary, we document here only the differences and limitations
```
torsava, line 7:
```suggestion
Because the minimal and full images work similarly, we document here only the differences and limitations
```

torsava:
> CI is completely green!  @torsava wanna do a sanity check? We can merge this and I can then submit other PR to address my ToDo records.

First of all, apologies for missing this ping :/

I have re-reviewed all the commits, and found only one typo and one question, see above. Overall it looks great, @frenzymadness , LGTM!

3.9-minimal/Dockerfile.rhel8
frenzymadness: I thought that it's no longer needed but it seems that it'd be better to have it here. I've added it. Thanks for spotting this!

3.9-minimal/README.md
frenzymadness: Fixed, thank you!

frenzymadness:
[test]

3.9-minimal/Dockerfile.rhel8
torsava: No worries! :+1:

frenzymadness:
[test]

frenzymadness:
I don't consider the current status of the CI a problem blocking the merging. @phracek do you want to check it? I remember that the SCLorg-CI jobs were running for days in the previous run and OpenShift 3 job is broken for quite some time.

torsava:
Woohoo! 🎉🥳

PR #470, opened 2021-08-20
Add poetry support

LenkaSeg:
It would be nice to have poetry support.

### DIFF
--- a/2.7/test/poetry-test-app/.s2i/environment
+++ b/2.7/test/poetry-test-app/.s2i/environment
@@ -1,0 +1,3 @@
+APP_MODULE=testapp:application
+ENABLE_POETRY=true
+DISABLE_SETUP_PY_PROCESSING=true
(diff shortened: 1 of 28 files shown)


frenzymadness:
Hello. Thanks for the PR! Have you tried micropipenv? We already support it and it can install dependencies from various sources including poetry.lock files.

A few notes:
* Poetry will soon stop supporting Python 2 so we should implement it for Python 3 only.
* It has to be implemented also in `src/s2i/bin/assemble` where the installation of the tools actually happens.
* It would be nice to have two commits here: One with changes made to the src and examples folders and a second one with generated content from distgen.

I can take the PR and improve it but first of all, I'd like to know whether micropipenv works for you or not and how/why.

abompard:
Hey! I didn't know about micropipenv, it seems to be a good candidate to replace our Poetry-based hack in `assemble`. However, we have Twisted as one of our indirect dependencies and we're hit by https://github.com/thoth-station/micropipenv/issues/186, since Twisted marks some dependencies as Windows-only and micropipenv tries to install them.

frenzymadness:
There is a new release of micropipenv with improved support for poetry. Could you please test it?

abompard:
Yes, that specific issue is fixed, thanks! Unfortunately we now hit another one: thoth-station/micropipenv#192 :-/

PR #471, opened 2021-09-01
Run only fedora TMT tests and not cwt tests.

phracek:
Available plans are here: https://github.com/sclorg/sclorg-testing-farm/tree/main/plans
The specification is the name without `fmf` extension.


frenzymadness:
Looks good to me although I don't understand why is the name of the second plan "fedora" when it runs OpenShift 3 tests on Centos 7. Also, should we try to run the new configuration here in this PR?

phracek:
@frenzymadness This is not possible. The updated plans are possible to run as soon as PR is merged. GitHubActions take the files from .github/workflows directory and not from PR :(

PR #472, opened 2021-09-02
This Github Action automatically pushes image into Quay.io registry

phracek:
Actions which are used are from GitHub Marketplace
https://github.com/marketplace/actions/buildah-build
https://github.com/marketplace/actions/push-to-registry

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -1,0 +1,78 @@
+name: Build and push images
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  setup-matrix:
+    runs-on: ubuntu-latest
+    # We need to store values for job build-and-push
+    outputs:
+      versions_matrix: ${{ steps.setup-job.outputs.versions_matrix }}
+      image_name: ${{ steps.setup-job.outputs.image_name }}
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Get all matrix versions and base image name
+        id: setup-job
+        run: |
+          # This command returns row with VERSIONS
+          versions_row=$(grep "VERSIONS" Makefile)
+          # Return only version string, spaces are separated by dash
+          versions_row=${versions_row/VERSIONS = /}
+          versions_row=$(echo $versions_row | tr " " "\n")
+          VERSIONS_MATRIX=""
+          for row in $versions_row; do
+            if [[ $row =~ "minimal" ]]; then
+              continue
+            fi
+            VERSIONS_MATRIX="$VERSIONS_MATRIX$row,"
+          done
+          # Remove last dash from string which now looks like X,Y,Z,
+          VERSIONS_MATRIX=${VERSIONS_MATRIX::-1}
+          # Create a string like [X,Y,Z] which is the matrix
+          VERSIONS_MATRIX="[$VERSIONS_MATRIX]"
+          # This command returns row with BASE_IMAGE_NAME
+          row=$(grep "BASE_IMAGE_NAME" Makefile)
+          # Return only base image name
+          BASE_IMAGE_ROW=${row/BASE_IMAGE_NAME = /}
+
+          echo ::set-output name=image_name::$BASE_IMAGE_ROW
+          echo ::set-output name=versions_matrix::$VERSIONS_MATRIX
+
+  build-and-push:
+    runs-on: ubuntu-latest
+    needs: setup-matrix
+    strategy:
+      matrix: ${{ needs.setup-matrix.outputs.versions_matrix }}
+    steps:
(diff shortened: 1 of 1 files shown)


phracek:
Please look at this PR which introduces automatic image pushing into Quay.io.

Propose a better BASH way if it is feasible.

.github/workflows/build-and-push.yml
```diff
@@ -0,0 +1,78 @@
...
+          versions_row=${versions_row/VERSIONS = /}
+          versions_row=$(echo $versions_row | tr " " "\n")
+          VERSIONS_MATRIX=""
+          for row in $versions_row; do
+            if [[ $row =~ "minimal" ]]; then
+              continue
+            fi
+            VERSIONS_MATRIX="$VERSIONS_MATRIX$row,"
+          done
+          # Remove last dash from string which now looks like X,Y,Z,
```
frenzymadness, line 32:
```suggestion
          # Remove last comma from string which now looks like X,Y,Z,
```

frenzymadness (commented):
I haven't found anything wrong. Given the fact that this is only for Centos 7, will there be also an action for other containers? I mean, the name of the action does not tell that it's only for Centos 7.

phracek:
> I haven't found anything wrong. Given the fact that this is only for Centos 7, will there be also an action for other containers? I mean, the name of the action does not tell that it's only for Centos 7.

Good point. Automatic pushing into RHEL is done by internal tools. Pushing into Fedora is not possible, because of karma etc.

All other containers in https://github.com/sclorg or will use the same GitHub action. But s2i-python-container is PoC.

.github/workflows/build-and-push.yml
```diff
@@ -0,0 +1,78 @@
...
+      image_name: ${{ steps.setup-job.outputs.image_name }}
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Get all matrix versions and base image name
+        id: setup-job
+        run: |
+          # This command returns row with VERSIONS
+          versions_row=$(grep "VERSIONS" Makefile)
+          # Return only version string, spaces are separated by dash
```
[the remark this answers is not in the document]
phracek: I will rewrite it a bit more.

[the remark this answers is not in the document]
phracek: Rewritten a bit more.

pkubatrh:
How about the new upstream container images, like python-39-minimal? Will they be covered by the same action?

phracek:
@pkubatrh Thanks for this question. There will be definitely another GitHub Action. I will do it as soon as this is gonna merged. Because of quay.io registry will be different. Or this can be built only with a condition https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions.

I would prefer to do it as a separate PR so we can see that this is working properly.

phracek:
Some example is here: https://www.edwardthomson.com/blog/github_actions_13_conditionals.html

phracek:
The pull request is ready for merge. redis-container already uses the same GitHub Action https://github.com/sclorg/redis-container/runs/3621309385 and https://github.com/sclorg/redis-container/runs/3621309480

phracek:
@frenzymadness The Github action for build and push is ready for merge. What do you think?

phracek:
[test]

phracek:
[test]

.github/workflows/build-and-push.yml
```diff
@@ -22,12 +24,23 @@ jobs:
           BASE_IMAGE_ROW=${row/BASE_IMAGE_NAME = /}
           echo ::set-output name=image_name::$BASE_IMAGE_ROW
 
+      - name: Checkout into generated section
+        run: |
+          git checkout --progress --force -B generated refs/remotes/origin/generated
```
frenzymadness, line 29: Do we really want to use the generated branch? I mean, I don't use it and I'm not sure there is a check that the branch is always up to date.

phracek: It is a good question. I did not test whether dist-gen works on Ubuntu, yet ;) I can try it and then we can call 'make generate-all' and take the resources correspondingly. But it needs time to investigate it.

phracek: Fix by the latest commit. We generate sources by each merge.

phracek:
> I've posted just one question. Otherwise, I'm not able to check it so we can merge it and see whether it works or not. Also, if we merge it before the minimal container, should I add it to the matrix?

The minimal container will have alone build-push action.

phracek:
[test]

frenzymadness:
We should sync about this. I don't think that generating sources via distgen is necessary but I might be missing something.

PR #477, opened 2021-10-06
Use ct pull image

phracek:
Before any OpenShift 3,4 tests execution, let's try to pull images.
If the images are not pullable, then failed, because the OpenShift 3,4 tests will fail too.

### DIFF
--- a/test/run-openshift
+++ b/test/run-openshift
@@ -15,6 +15,10 @@ source "${THISDIR}/test-lib-python.sh"
 
 set -eo nounset
 
+# Pull images before test execution and exit if docker/podman pull operation failed
+ct_pull_image "registry.access.redhat.com/ubi7/ubi" "true"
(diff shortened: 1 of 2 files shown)


phracek:
https://github.com/sclorg/container-common-scripts/pull/198

pkubatrh:
[test]

phracek:
[test-all]

phracek:
OpenShift tests are failing because of ct_pull_image function. The fix is here: https://github.com/sclorg/container-common-scripts/pull/222

phracek:
Let's have a look and test also distgen generation test suite. It might fail because `3.9-minimal` is not generated by distgen.
[test-all]

phracek:
This is solved by https://github.com/sclorg/s2i-python-container/pull/494

Issue #479, opened 2021-10-16
pip install . is not run if setup.cfg is present


yrro:
Following the [Packaging Python Projects](https://packaging.python.org/tutorials/packaging-projects/) leads you down the road of creating a `pyproject.toml` file (which declares that setuptools is used to build the project) and a `setup.cfg` file, which appears to be a newer alternative to `setup.py`.

I noticed that the `assemble` script only runs `pip install .` if `setup.py` is present, so I had to convert my `setup.cfg` file to a `setup.py` file. No big deal. But it would be nice if I could have kept the static `setup.cfg` file.

I don't know if it would be better to change the `assemble` script to check for `setup.cfg` as well, or whether it should be looking for a `pyproject.toml` file (since I think projects should put stuff in there that tells `pip` (or other package managers) how to build the project) and then fall back to `setup.py` (for projects that implicity use setuptools without declaring such in `pyproject.toml`).

frenzymadness:
What Python version and platform combination do you prefer?

yrro:
Sorry - didn't get around to trying this myself. But I am 99% sure your test reproduces what I'm trying to do exactly. Thanks :)

PR #480, opened 2021-10-18
Allow apps with setup.cfg without setup.py

frenzymadness:
Fixes: #479

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -219,7 +219,7 @@ file inside your source code repository.
 
 * **DISABLE_SETUP_PY_PROCESSING**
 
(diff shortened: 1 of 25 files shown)


frenzymadness:
[test]

dperaza4dustbit:
OK, thanks for pointing us here @gabemontero I will follow up with @frenzymadness on slack

frenzymadness:
I'm not able to test it now. Could you please provide some logs? As far as I know, there might be some issues currently caused by the latest release of pipenv.

dperaza4dustbit:
OK, going to work on disabling the test on origins first, and then will work with you @frenzymadness on slack to get you some logs.

PR #483, opened 2021-11-22
Use groupified apiVersion

yselkowitz:
Non-groupified objects were deprecated in OCP 4.7.

### DIFF
--- a/examples/django-postgresql-persistent.json
+++ b/examples/django-postgresql-persistent.json
@@ -1,6 +1,6 @@
 {
   "kind": "Template",
-  "apiVersion": "v1",
+  "apiVersion": "template.openshift.io/v1",
   "metadata": {
     "name": "django-psql-persistent",
     "annotations": {
@@ -56,7 +56,7 @@
     },
     {
       "kind": "Route",
-      "apiVersion": "v1",
+      "apiVersion": "route.openshift.io/v1",
       "metadata": {
         "name": "${NAME}",
         "annotations": {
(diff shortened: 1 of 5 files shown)


frenzymadness:
The change looks good to me. @phracek could you please also take a look? It might be something we need to do for other images as well.

[test-openshift]

frenzymadness:
In the first job, there are a lot of errors like "no RBAC policy matched". Not sure what causes this.

frenzymadness:
I thought that the problem is that this change is not compatible with OS 3 but it seems that it is and the problem has to be somewhere else.

phracek:
[test-openshift]

phracek:
[test]

phracek:
> In the first job, there are a lot of errors like "no RBAC policy matched". Not sure what causes this.

This problem should be fixed by https://github.com/sclorg/container-common-scripts/pull/220.
May be we should switch to user admin a bit early. But the OpenShift 3 tests should pass now.

phracek:
@yselkowitz As soon as you rebase common directory, then CentOS7 OpenShift 3 tests will be passing too.

frenzymadness:
@yselkowitz let me know if I should do it for you. You need to update the common submodule and then push the change here.

phracek:
[test-all]

frenzymadness:
@phracek there seems to be a problem with `rhscl/postgresql-10-rhel7` image. Do you know where this comes from or should I investigate it?

phracek:
I will investigate it a bit more soon. The issues is here: https://github.com/sclorg/s2i-python-container/issues/491

phracek:
[test-all]

phracek:
The issue is filed, let's do not block this pull request.

PR #489, opened 2021-12-22
add pip extra args

TecHunter:
Pip cache is disabled by default in s2i image, this setting allows to override this. Useful when building with Buildah:

```bash
ctr=$(buildah from registry.redhat.io/ubi8/python-38:latest)
mnt=$(buildah unshare buildah mount $ctr)
#host cache folder
mkdir ~/.cache/pip -p
[ -d $mnt/.pip-cache ] && rm $mnt/.pip-cache
#link from container  to host folder
ln -s ~/.cache/pip $mnt/.pip-cache
buildah config --env PIP_CACHE_DIR=/.pip-cache $ctr
buildah unshare buildah run --user 0 $ctr -- /usr/libexec/s2i/assemble
buildah unmount $ctr
buildah commit $ctr my-app
```

### DIFF
--- a/src/s2i/bin/assemble
+++ b/src/s2i/bin/assemble
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+pip_install_args=${PIP_INSTALL_ARGS:-}
+if [ ! -z "$PIP_CACHE_DIR" ]; then
+  pip_install_args="$pip_install_args --cache-dir $PIP_CACHE_DIR"
+fi
+
 function is_django_installed() {
   python -c "import django" &>/dev/null
 }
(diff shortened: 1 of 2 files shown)


frenzymadness:
Hello and thanks for the PR!

Have you tried to use a configuration file for pip or set what you need via environment variables? Both ways should make it possible to alter pip's behavior without the need to update all the pip calls in the scripts for all possible use cases. See https://pip.pypa.io/en/stable/topics/configuration/#environment-variables

What do you think?

TecHunter:
Hello!

I've tried with PIP_CACHE_DIR but it seems to ignore it completely. I'm new to python pip, maybe you have an idea why?

frenzymadness:
The problem is that `PIP_NO_CACHE_DIR` takes precedence if defined to a non-empty value which is the case for our containers. What you need is to specify both `PIP_CACHE_DIR` to the dir you want and `PIP_NO_CACHE_DIR` to an empty string or undefine it, if it's possible in buildah (it's not in a Dockefile). I've tested it and while having these environment variables:
```
PIP_CACHE_DIR=/tmp/cachetest
PIP_NO_CACHE_DIR=
```
everything works as expected and I have the cache in the specified folder.

TecHunter:
nice finding! thanks

frenzymadness:
If it works for you, can we close this PR?

PR #496, opened 2022-02-11
Schedule CentOS Stream 9 tests in Testing Farm

phracek:
This pull request adds only support for scheduling
tests in Testing Farm. The alone changes
regarding building, testing, and push will be
as a separate pull request.


frenzymadness (commented):
It would be nice to have HTTPS also for artifacts.dev.testing-farm.io but the certificate they use is not valid for that domain. Where can I open an issue for that?

The change itself looks good to me and should not be blocked on the previous comment.

phracek:
@frenzymadness Here: https://github.com/teemtee/tmt/issues

PR #499, opened 2022-02-15
do not end line from readlink with NUL

zmiklank:
Source: https://github.com/sclorg/s2i-perl-container/pull/216


phracek:
This affects only container tests.

[test]

zmiklank:
Hmm, there is the "Couldn't resolve host name for https://mirrors.fedoraproject.org/" bug in the Fedora tests.

frenzymadness:
Thanks for the PR. I've switched it to the draft because we use distgen in this repository so I can transfer the change to it.

frenzymadness:
So, what I did here is that I split the one commit into two. The first one is the actual change to the source code and the second one contains changes propagated by distgen to the generated folders. This brings the change also to 3.10, omits unsupported version 3.3, and contains also manual change to 3.9-minimal which is not yet covered by distgen.

[test-all]

frenzymadness:
There was no need to start OS tests and we know about the issue with Fedora Testing farm. It seems ready to be merged. Any last comments?

zmiklank:
No, thanks for adjusting the PR.

frenzymadness:
I thank you!

PR #501, opened 2022-02-28
Support test, build, and push for CentOS Stream 9

phracek:
This pull request adds support testing, building, and pushing images in CentOS Stream 9.

PR #502, opened 2022-03-02
Fix docker-tests.yml file with setting proper variables

phracek:


phracek:
This does not affect tests but blocks https://github.com/sclorg/s2i-python-container/pull/501

PR #503, opened 2022-03-02
Use testing farm as GitHub action

phracek:


.github/workflows/container-tests.yml
```diff
@@ -0,0 +1,90 @@
...
+        run: |
+          PR_URL="${{ github.event.comment.issue_url }}"
+          echo "::set-output name=PR_NR::${PR_URL##*/}"
+
+      - name: Checkout repo
+        uses: actions/checkout@v2
+        with:
+          ref: "refs/pull/${{ steps.pr_nr.outputs.PR_NR }}/head"
+
+      - name: SHA value
```
zmiklank, line 53: is this step needed? I can not see, where the `SHA` output is used.

.github/workflows/container-tests.yml
```diff
@@ -0,0 +1,90 @@
...
+        id: github_action
+        uses: sclorg/testing-farm-as-github-action@v1
+        with:
+          api_key: ${{ steps.tf_values.outputs.API_KEY }}
+          git_url: ${{ steps.tf_values.outputs.TMT_REPO }}
+          git_ref: ${{ steps.tf_values.outputs.BRANCH_NAME }}
+          tmt_plan_regex: ${{ matrix.tmt_plan }}
+          pull_request_status_name: ${{ matrix.context }}
+          variables: "REPO_URL=$GITHUB_SERVER_URL/$GITHUB_REPOSITORY;REPO_NAME=$GITHUB_REPOSITORY;PR_NUMBER=${{ steps.pr_nr.outputs.PR_NR }};OS=${{ matrix.os_test }};TEST_NAME=test"
+          secrets: "QUAY_USERNAME=${{ secrets.QUAY_IMAGE_SCLORG_BUILDER_USERNAME }};QUAY_TOKEN=${{ secrets.QUAY_IMAGE_SCLORG_BUILDER_TOKEN }}"
```
zmiklank, line 87: Doesn't Fedora require `QUAY_IMAGE_BUILDER_{USERNAME,TOKEN}` instead of the sclorg one?

.github/workflows/container-tests.yml
phracek: This is not needed.

.github/workflows/container-tests.yml
phracek: Good point. Only `c9s` is valid. I would leave it as it is. RHEL7 and RHEL8 do not need it as well. But two the same actions, does not make sense

zmiklank: And what will be the effect on Fedora tests, if we leave it as it is?

phracek: Nothing, the secrets are hidden. Only c9s.yml use it https://github.com/sclorg/sclorg-testing-farm/blob/main/plans/c9s.fmf#L22

zmiklank: I would rather doublecheck if [this line](https://github.com/sclorg/sclorg-testing-farm/blob/c2778a0a8fb2ea9120576c3cccde76892e389779/plans/c9s.fmf#L22) is really needed in the tmt plan before merging this PR.

phracek: It is not needed at all. I will remove it.

phracek: https://github.com/sclorg/sclorg-testing-farm/pull/13

zmiklank: Thanks.

zmiklank (approved):
LGTM.

zmiklank:
[test]

PR #504, opened 2022-03-22
use composite build-and-push action

zmiklank:
- move contents of -minimal.yml to build-and-push.yml
- schedule rebuild on every Wednesday
- allow running job only for sclorg repositories, not for forks


phracek (approved):
LGTM. As soon as build-and-push action is fixed because of .exclude files, we can merge it.

PR #506, opened 2022-04-27
Use Fedora compose for tests and update Testing Farm as GitHub Action

phracek:


phracek:
[test-all]

phracek:
Tests are still using the old 'CentOS-7' compose and not this one :(. Which is correct. After merging, it will use Fedora-latest.

frenzymadness:
If the only way to test this is to merge this, let's merge this :)

phracek:
As soon as #507 is merged, than also Fedora tests should be passing.

PR #510, opened 2022-05-10
Unstable tests

frenzymadness:
@phracek with the list of possible unstable tests and `IGNORE_UNSTABLE_TESTS` env variable, we can make the downstream CI more stable.

### DIFF
--- a/2.7/test/run
+++ b/2.7/test/run
@@ -8,6 +8,13 @@
 #
 declare -a WEB_APPS=({gunicorn-config-different-port,gunicorn-different-port,django-different-port,standalone,setup,setup-cfg,setup-requirements,django,numpy,app-home,npm-virtualenv-uwsgi,locale,mod-wsgi,pipenv,pipenv-and-micropipenv-should-fail,pin-pipenv-version,app-module}-test-app)
 
+# Some tests, like the one using the latest pipenv, might be unstable
+# because new upstream releases tend to break our tests sometimes.
+# If a test is in UNSTABLE_TESTS and IGNORE_UNSTABLE_TESTS env
+# variable is defined, a result of the test has no impact on
+# the overall result of the test suite.
+declare -a UNSTABLE_TESTS=(pipenv-test-app)
+
 # TODO: Make command compatible for Mac users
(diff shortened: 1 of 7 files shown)


frenzymadness:
[test-all]

phracek (approved):
LGTM. Test are passing.

phracek:
Thanks. Cool.

zmiklank:
Do we want this to be python specific, or for all sclorg's containers?
If we want to use it everywhere, we can consider adding it to the common-scripts:
https://github.com/sclorg/container-common-scripts/blob/f655710e567a22fc653679e9496ef7de639f6ad3/test-lib.sh#L1091
(Or maybe make the s2i-python use the function from common-scripts.)

frenzymadness:
@phracek do you know about any reason why we don't use `ct_run_tests_from_testset` in the python container?

zmiklank:
I know, that I am not @phracek, but I think I know the answer.
We do not use it here, because the `ct_run_tests_from_testset` has been written only after the code in `test/run` has been written and used. The `ct_run_tests_from_testset` is used for (and has been written because of) openshift tests because they did not have such test mechanism before. I see, that the python container has not been yet rewritten to use it even in the openshift tests. I will create a PR for that.

Anyway, I do not see a problem in using it even in container tests.

frenzymadness:
@zmiklank interesting, thanks. Do you have some free cycles to implement this change in common submodule and then switch this container image to the mentioned function? If not, I can try to do it once I return from my PTO.

zmiklank:
With https://github.com/sclorg/s2i-python-container/pull/515 merged, I would say, that we can continue with support of unstable test, and define unstable tests for python container images.

frenzymadness:
This should now work as expected.

zmiklank:
[test-all]

phracek:
OpenShift 4 failures are fixed by https://github.com/sclorg/s2i-python-container/pull/525

phracek (approved):
LGTM. OpenShift failures are fixed by https://github.com/sclorg/s2i-python-container/pull/525

PR #511, opened 2022-05-12
run all openshift4 tests, even if some fail

zmiklank:


phracek:
The code LGTM. But you need to do two things in the pull request.
1) Modify `test/test-openshift-remote-cluster` and `test/test-lib-python.sh`
2) Regenerate source by this way https://github.com/sclorg/s2i-python-container#contributing

zmiklank:
Oh, right! Thanks for noticing, will do it.

zmiklank:
@phracek, the files, which I have changed do not seem to be generated by the distgen, they are only symlinks.
What should be done in this case?

frenzymadness:
Try this: `make clean-versions && make generate-all` and there should be nothing to commit after that.

zmiklank:
Thanks, @frenzymadness 
There is a change, which you already addressed here: https://github.com/sclorg/s2i-python-container/pull/510/commits/e42e04ea21145d6b59fbc113f8d339350ba0362b

And there is deleted the whole generated 3.9-minimal, which could be related to the fact, that `Makefile` sets `SKIP_GENERATOR_FOR="3.9-minimal"`.

But there are no changes to commit, that could be caused by this PR.

frenzymadness:
I've pushed the fix for the distgen inconsistency to the master branch so don't have to deal with it in all active PRs.

phracek (approved):
I don't see any problem here: LGTM. Thanks for handling this pull request.

phracek:
[test-all]

phracek:
RHEL8 failures are updated by this PR https://github.com/sclorg/s2i-python-container/pull/519. RHEL-8.3.1 was deleted from Testing Farm and marked as EOL.

PR #514, opened 2022-05-16
Newer Fedora with older Pythons

frenzymadness:
I've realized that because we maintain older Pythons in all Fedora releases, there is no reason to use only combinations of Fedora and the main python3 package in every release. We can always use the latest stable Fedora and whatever python3.X package is available in it.

I'm using it to switch Python 3.9 from Fedora 34 which will soon be EOL to Fedora 35 but we can also restore even older Pythons without worrying about what happens during the next Fedora update (soon 35 → 36).

Do you see any problem with this approach?

Cc: @torsava @hroncok

### DIFF
--- a/3.9/Dockerfile.fedora
+++ b/3.9/Dockerfile.fedora
@@ -1,6 +1,6 @@
 # This image provides a Python 3.9 environment you can use to run your Python
 # applications.
-FROM quay.io/fedora/s2i-base:34
+FROM quay.io/fedora/s2i-base:35
 
 EXPOSE 8080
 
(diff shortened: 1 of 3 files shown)


frenzymadness:
[test]

phracek:
[test-openshift]

phracek (approved):
Thanks for this important update. LGTM.

hroncok:
I've talked to @frenzymadness about this. We do not support alternative Python versions for deployments, only for testing. We regularly decided not to fix CVEs in them because of this. Please, don't merge this.

PR #520, opened 2022-05-30
do not build python3.9 centos7 image in build-and-push

zmiklank:
The Dockerfile for centos7, python 3.9 does not exist, so the build currently fails.


phracek (approved):
LGTM.

PR #524, opened 2022-06-08
Exit oc4  tests if login fails

zmiklank:
The nounset shell option needs to be set after the OC login attempt,
because (from man shopt):
```
-u
  If expansion is attempted on an unset  variable  or
  parameter,  the  shell prints  an  error  message, and, if not
  interactive exits with a non-zero status.
```
That is in this matter inconvenient, as we do not really know from the logs
what has happened if the oc login fails and moreover the test suite ends
with success, as only the first failed test sets TESTSUITE_RESULT to 1.
If we check login with the ct_os_check_login function, it handles the
situation conveniently.

### DIFF
--- a/test/run-openshift-remote-cluster
+++ b/test/run-openshift-remote-cluster
@@ -20,7 +20,6 @@ test_python_imagestream
 test_python_s2i_templates
 "
 
-set -u
 # Pull image before going throw tests
 # Exit in case of failure
 ct_pull_image "quay.io/centos7/postgresql-10-centos7" "true"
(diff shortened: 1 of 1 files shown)


zmiklank:
[test-all]

frenzymadness:
The failure of OS 4 is present also in another PR os it should be unrelated to this change. I like it. @phracek what do you think?

zmiklank:
I force-pushed the commit with the corrected commit message just now, as some words got lost from it somehow and it did not make sense :).

phracek:
OpenShift 4 failures are fixed by https://github.com/sclorg/s2i-python-container/pull/525

[test-all]

zmiklank:
Rebased on master.
[test-all]

zmiklank:
There is some problem with test infra. Will rerun the tests now:
[test-all]

zmiklank:
There is some problem with test infra. Will rerun the tests now:
[test-all]

phracek:
[test-all]

phracek (approved):
LGTM. Thanks for driving it. Let's merge it as soon as tests are passing.

PR #526, opened 2022-06-14
Enable testing s2i-python-container on RHEL9 host

phracek:
Enable testing s2i-python-container on RHEL9 host.

Remove also obsoleted variables.
As soon as s2i-python-container will have Dockerfile.rhel9 files, then it will be really tested.


phracek:
This does not need review. It does not affect python code at all.

PR #528, opened 2022-06-27
Support building testing rhel9

phracek:
This pull request adds support for building and testing python-3.9 container on RHEL9 host.

It also updates documentation with references.


phracek:
[test-all]

phracek:
It looks like we have a problem with the subscription. Let's try it a bit more later.

phracek:
[test-all]

phracek:
[test-all]

frenzymadness:
The diff looks good to me. Unfortunately, the tests for RHEL 9 don't work yet.

phracek:
I am solving issues with subscriptions to RHEL9. Please be patient.

phracek:
[test-all]

phracek:
By this pull request https://github.com/sclorg/s2i-python-container/pull/529 s2i-python-container should be able to be tested on RHEL9 as well.

[test-all]

frenzymadness:
I'm gonna investigate it.

frenzymadness:
The problem causing the tests to fail is that the new image does not contain code for virtualenv activation. A fix is on the way.

pkubatrh:
>How is that possible that it's possible to build the container image in CI even one of the packages is not available?

The CI does not use only UBI repositories, but regular RHEL repositories as well.

If we can remove enchant now that would be the best outcome, otherwise we can ask for adding enchant to UBI.

frenzymadness:
> The CI does not use only UBI repositories but regular RHEL repositories as well.

And is that correct? I mean, the build is not reproducible if you don't have an access to the RHEL repos so if you just clone this repo the build will fail.

hhorak:
> It seems that pyenchant provides binary wheels so, for Python 3, we might not need the enchant anymore. I also cannot say whether enchant2 is just a replacement for enchant.

Does it provide them on all arches we need for RHEL?

frenzymadness:
> > It seems that pyenchant provides binary wheels so, for Python 3, we might not need the enchant anymore. I also cannot say whether enchant2 is just a replacement for enchant.
> 
> Does it provide them on all arches we need for RHEL?

No, the wheels are noarch containing only Python code. But it does not matter in the end. Pyenchant is only a Python wrapper around C lib provided by the enchant package.

I wouldn't block this on the missing package. We can continue without it and wait until somebody requests it or be proactive and request it to be added to UBI repos in advance. Is it considered to be a backward incompatibility if a UBI9 container image lacks a functionality available in UBI8? I don't think so.

frenzymadness:
I've created a new PR for the test of pyenchant: #530 We can merge it once the enchant package is available on all platforms.

This should now pass all the tests: [test-all]

The remaining question is whether we should request the inclusion of enchant to UBI 9 repos proactively or not.

frenzymadness:
The CI failures seem to be unrealted and the PR changes only blank lines in RHEL 7 Dockerfile so this doesn't have to be a blocker.

phracek:
OpenShift 4 cluster should be up and running again.
[test-all]

pkubatrh:
>The remaining question is whether we should request the inclusion of enchant to UBI 9 repos proactively or not.

We do need to do that either way - the existing released images already have enchant inside. I will file a ticket for it.

phracek (commented):
Everything LGTM from my point of view.

frenzymadness:
All green. I'm gonna merge this. Thanks @phracek! @pkubatrh cc me to the issue please or link it here, thank you.

PR #530, opened 2022-07-13
Test installation and import of pyenchant package

frenzymadness:
enchant RPM should always be installed to support pyenchant Python wrapper. The package is not available in UBI 9 repos so this test has to wait for it.

### DIFF
--- a/3.9-minimal/test/from-dockerfile/mod_wsgi.Dockerfile.tpl
+++ b/3.9-minimal/test/from-dockerfile/mod_wsgi.Dockerfile.tpl
@@ -19,9 +19,9 @@ from #IMAGE_NAME#
 # Copy app sources together with the whole virtual environment from the builder image
 COPY --from=builder $APP_ROOT $APP_ROOT
 
-# Install httpd package - runtime dependency of our application
+# Install httpd and echant packages - runtime dependencies of our application
 USER 0
-RUN microdnf install -y httpd
+RUN microdnf install -y httpd enchant
 USER 1001
 
 # Set the default command for the resulting image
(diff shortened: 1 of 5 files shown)


frenzymadness:
When enchant package will be available in UBI 9 repos, we can revert this commit https://github.com/sclorg/s2i-python-container/commit/8d1411b9d1259a87f5611d37313bf74a10c1d0df and add the test proposed here.

@pkubatrh could you please link the request for inclusion of the package here?

frenzymadness:
The package should be available soon so this PR basically adds a test for it. In the future, we can move to enchant2 which is also supported by pyenchant package. [test-all]

phracek:
Let's test RHEL9 unsubscribed tests
[test]

frenzymadness:
I expect it to pass because the package is already available in UBI9.

frenzymadness:
Let's use this PR to test the new test. I'm gonna resolve the conflict and then I'll try to find a package that is available only via subscription.

frenzymadness:
[test]

frenzymadness:
I've used python3-requests-gssapi which seems to be available in RHEL repos but not in UBI and it seems to work well. RHEL9 job is green but RHEL9 - Unsubscribed host has failed during the installation of the mentioned package.

I'm gonna remove the last commit and run tests again.

frenzymadness:
[test]

frenzymadness:
The problem reported in CI is real and I'd like to use this PR to track it. The problem is probably in micropipenv/setuptools and I'm gonna workaround it and release a new version.

frenzymadness:
[test]

frenzymadness:
[test]

frenzymadness:
All green!

PR #535, opened 2022-07-27
Remove EOL F34/Py 3.9 and update Py 3.10 from F35 to F36

frenzymadness:
SSIA


frenzymadness:
[test]

frenzymadness:
One last thing - update the bundled pip to the version from the latest stable Fedora. [test-all]

frenzymadness:
[test]

frenzymadness:
There is a problem. pip 9 respects metadata about Python versions compatibility but pip 7 we have in RHEL 7 does not.

frenzymadness:
[test]

frenzymadness:
The failure of Fedora CI can be ignored, for now, the problem is in uWSGI not in the image. The best is to review this commit by commit. @torsava or @phracek do you want to take a look?

src/s2i/bin/assemble
```diff
@@ -72,10 +72,20 @@ fix-permissions /opt/app-root -P
 # * pip < 9 does not support different packages' versions for Python 2/3
 # * pip < 19.3 does not support manylinux2014 wheels. Only manylinux2014 (and later) wheels
 #   support platforms like ppc64le, aarch64 or armv7
+#
+# Also pip 9 we have in RHEL 8 can use metadata about supported Python version
+# so we have to limit the version of pip only for RHEL 7 where the default pip
+# version is 7. And because pip still supports Python 3.8, the limit makes sense
+# only for Python 2.7.
```
torsava, line 79: I'm not sure I understand the latest sentence. I thought this makes sense only for Python 2.7 in RHEL 7 is because there the pip version is too old to use the metadata. What does it have to do with pip supporting Python 3.8?

frenzymadness:
Thanks for the review. Let's do the removal first so we can test this update without Py2.7/RHEL7. See https://github.com/sclorg/s2i-python-container/pull/537

frenzymadness:
Rebased, fix for EL7/py2 removed, [test-all]

torsava:
> Rebased, fix for EL7/py2 removed, [test-all]

LGTM!

PR #539, opened 2022-08-11
New generator for distgen working with minimal images

frenzymadness:
Python replacement for the black magic in https://github.com/sclorg/container-common-scripts/blob/233496946a028218958d8fbb1857f921ad8a99ef/common.mk#L147-L164 and in https://github.com/sclorg/container-common-scripts/blob/master/generate.sh

It might be possible to merge some of the files specific for the minimal container to the original one but that can be done in the future. This does not apply to README.md which is completely different.


frenzymadness:
[test]

pkubatrh:
I am wondering, can we add the black magic replacements to container-common-scripts, or are they too specific for the python repository to be used elsewhere?

frenzymadness:
> I am wondering, can we add the black magic replacements to container-common-scripts, or are they too specific for the python repository to be used elsewhere?

I'd like to replace what we have in the common scripts with this after some testing here. How many images use distgen? I know only about postgres.

generator.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def filename_to_distro_config(filename, version, mapping):
+    """Find distgen distro config from a filename.
+
+    This is usually needed only for dockerfiles.
+    - Dockerfile.rhelXX → rhel-XX-x86_64.yaml
+    - Dockerfile.cXXs → centos-stream-XX-x86_64.yaml
+    - Dockerfile.fedora → the newest fedora-XX-x86_64.yaml
+    - Dockerfile → centos-7-x86_64.yaml
```
torsava, line 62: Food for thought: "centos-7" might not be a good default anymore. How about creating a suffix for centos 7 (e.g. `centos7`) and error out when no suffix is given?

generator.py
```diff
@@ -0,0 +1,180 @@
...
+    elif m := re.match(r".*\.c(\d+)s$", filename):
+        config = f"centos-stream-{m.group(1)}-x86_64.yaml"
+    elif filename.endswith(".fedora"):
+        sorted_configs = sorted(
+            c for c in mapping[version] if c.startswith("fedora")
+        )
+        if sorted_configs:
+            config = sorted_configs[-1]
+        else:
+            config = None
```
torsava, line 79: Let's error out when there are multiple Fedora versions available, as there should be only one?

torsava:
Meta thought from meeting with Lumir: Currently in multispec it's necessary to exclude version/distro combinations that are unwanted. How about switching it around and only declaring version/distro combinations that are *wanted*? It feels more natural and increases readability.

src/cccp.yml
```diff
@@ -1,3 +0,0 @@
-# This is for the purpose of building this container
-# on the centos container pipeline.
-job-id: python-{{ spec.short_ver }}-centos7
```
torsava, line 3: Как ты мог, това́рищ?!

specs/multispec.yml
```diff
@@ -111,16 +112,29 @@ specs:
...
       pkg_prefix: "python39"
       main_image: "registry.access.redhat.com/ubi8/python-39"
 
+    "3.9-minimal":
+      version: "3.9"
+      short_ver: "39"
+      minimal: true
+      base_img_version: "1"
+      python_img_version: "1"
+      scl: "rh-python39"
```
torsava, line 129: Is this relevant? There's no rh-python39 collection at all. Maybe even remove it from the `3.9` config.

frenzymadness:
> Meta thought from meeting with Lumir: Currently in multispec it's necessary to exclude version/distro combinations that are unwanted. How about switching it around and only declaring version/distro combinations that are _wanted_? It feels more natural and increases readability.

This is not possible in distgen, see: https://github.com/devexp-db/distgen/blob/main/docs/spec_multispec.rst#multispecs

But, we might be able to implement it outside of distgen. The current workflow takes all the possible combinations from multispec.yaml (generating all of them, excluding specified). But we use them one by one only for files for which we need them for so we might also come with some external yaml config containing all the wanted combinations and remove the matrix section from distgen. Or, we might try to implement this in distgen itself.

test/run
```diff
@@ -6,7 +6,11 @@
...
 # The image has to be available before this script is executed.
 #
-declare -a WEB_APPS=({gunicorn-config-different-port,gunicorn-different-port,django-different-port,standalone,setup,setup-cfg,setup-requirements,django,numpy,app-home,npm-virtualenv-uwsgi,locale,mod-wsgi,pipenv,pipenv-and-micropipenv-should-fail,pin-pipenv-version,app-module{% if spec.version.startswith("3.") %},micropipenv,micropipenv-requirements{% endif %}}-test-app)
+{% if spec.minimal %}
+declare -a WEB_APPS=({gunicorn-config-different-port,gunicorn-different-port,django-different-port,standalone,setup,setup-requirements,django,numpy,app-home,locale,pipenv,pipenv-and-micropipenv-should-fail,pyuwsgi-pipenv,app-module,micropipenv}-test-app)
+{% else %}
+declare -a WEB_APPS=({gunicorn-config-different-port,gunicorn-different-port,django-different-port,standalone,setup,setup-cfg,setup-requirements,django,numpy,app-home,npm-virtualenv-uwsgi,locale,mod-wsgi,pipenv,pipenv-and-micropipenv-should-fail,pin-pipenv-version,app-module,pyuwsgi-pipenv{% if spec.version.startswith("3.") %},micropipenv,micropipenv-requirements{% endif %}}-test-app)
+{% endif %}
```
torsava, line 13: For readability and easier maintainability, I think it'd be better to separate it into:

1. Tests for all images
2. Tests for full (non-minimal) images.

And then for full images just combine 1 & 2.

test/run
```diff
@@ -249,6 +261,20 @@ test_from_dockerfile(){
...
+  ct_test_app_dockerfile $test_dir/from-dockerfile/uwsgi.Dockerfile.tpl $test_dir/uwsgi-test-app 'Hello World from uWSGI hosted WSGI application!' app-src
+  check_result $?
+
+  # mod_wsgi in micropipenv-requirements-test-app
+  # TODO: Find a better way how to specify full/minimal images that should work together
+  sed "s@#IMAGE_NAME#@${IMAGE_NAME}@;s@#FULL_IMAGE_NAME#@ubi{{ spec.el_version }}/python-{{ spec.short_ver }}:1@" $test_dir/from-dockerfile/mod_wsgi.Dockerfile.tpl > $test_dir/from-dockerfile/Dockerfile
+  ct_test_app_dockerfile $test_dir/from-dockerfile/Dockerfile $test_dir/micropipenv-requirements-test-app 'Hello World from mod_wsgi hosted WSGI application!' app-src
+  check_result $?
+}
+
```
torsava, line 277: `test_from_dockerfile` and `test_from_dockerfile_minimal` look very different, can you explain why?

3.9-minimal/test/run
```diff
@@ -6,13 +6,20 @@
 # IMAGE_NAME specifies a name of the candidate image used for testing.
 # The image has to be available before this script is executed.
 #
-declare -a WEB_APPS=({gunicorn-config-different-port,gunicorn-different-port,django-different-port,standalone,setup,setup-requirements,django,numpy,app-home,locale,pipenv,pipenv-and-micropipenv-should-fail,pin-pipenv-version,app-module,micropipenv}-test-app)
```
torsava, line 9: Why did `pin-pipenv-version` test disappear?

Makefile
```diff
@@ -10,3 +10,13 @@ DOCKER_BUILD_CONTEXT = ..
 .PHONY: $(shell test -f common/common.mk || echo >&2 'Please do "git submodule update --init" first.')
 
 include common/common.mk
+
+.PHOHY: generate-new
+generate-new:
+	for version in ${VERSIONS} ; do \
+		if [[ "$$version" == *-minimal ]]; then \
+			python3 ./generator.py -v $$version -m manifest-minimal.yml -s specs/multispec.yml ; \
```
torsava, line 18: Why not add a shebang to `generator.py` instead?

Makefile
```diff
@@ -10,3 +10,13 @@ DOCKER_BUILD_CONTEXT = ..
...
+		else \
+    		python3 ./generator.py -v $$version -m manifest.yml -s specs/multispec.yml ; \
```
torsava, line 20: This line needs one more indentation.

torsava:
First review pass finished. Overall it's looking much better than the previous machinery, I like it!

torsava:
> > Meta thought from meeting with Lumir: Currently in multispec it's necessary to exclude version/distro combinations that are unwanted. How about switching it around and only declaring version/distro combinations that are _wanted_? It feels more natural and increases readability.
> 
> This is not possible in distgen, see: https://github.com/devexp-db/distgen/blob/main/docs/spec_multispec.rst#multispecs
> 
> But, we might be able to implement it outside of distgen. The current workflow takes all the possible combinations from multispec.yaml (generating all of them, excluding specified). But we use them one by one only for files for which we need them for so we might also come with some external yaml config containing all the wanted combinations and remove the matrix section from distgen. Or, we might try to implement this in distgen itself.

Good to know. That'd be a great idea for the future, but no need to burden this PR with it.

frenzymadness:
> > > Meta thought from meeting with Lumir: Currently in multispec it's necessary to exclude version/distro combinations that are unwanted. How about switching it around and only declaring version/distro combinations that are _wanted_? It feels more natural and increases readability.
> > 
> > This is not possible in distgen, see: https://github.com/devexp-db/distgen/blob/main/docs/spec_multispec.rst#multispecs
> > 
> > But, we might be able to implement it outside of distgen. The current workflow takes all the possible combinations from multispec.yaml (generating all of them, excluding specified). But we use them one by one only for files for which we need them for so we might also come with some external yaml config containing all the wanted combinations and remove the matrix section from distgen. Or, we might try to implement this in distgen itself.
> 
> Good to know. That'd be a great idea for the future, but no need to burden this PR with it.

I've filled an issue in distgen and it seems that they're open to this idea. Will see how hard it'll be.

test/run
frenzymadness: Both tests call ct_test_app_dockerfile twice for two applications - where do you see a big difference?

3.9-minimal/test/run
frenzymadness: It's still there but just for full-size container images. Minimal has a new test - pyuwsgi-pipenv.

Makefile
frenzymadness: It seems that there is a combination of spaces and tabs on line 20.

Makefile
frenzymadness: How is shebang better?

test/run
frenzymadness: Done. Hope it'll work in Bash.

generator.py
frenzymadness: Done. Hope it's not Python-specific.

generator.py
frenzymadness: Done. It requires handling of dead symlinks but that isn't complicated.

frenzymadness:
See the new commits - some of them are fixup but it's better for review. [test]

2.7/test/run
```diff
@@ -247,6 +255,20 @@ test_from_dockerfile(){
   check_result $?
 }
 
+test_from_dockerfile_minimal(){
+  info "Test from Dockerfile"
+
```
torsava, line 260:
```suggestion

The following tests are for multi-stage builds. These technically also work on full images, but there is no reason to do multi-stage builds with full images.

```

test/run
torsava: Added a comment suggestion instead.

test/run
```diff
@@ -249,6 +264,20 @@ test_from_dockerfile(){
   check_result $?
 }
 
+test_from_dockerfile_minimal(){
+  info "Test from Dockerfile"
+
```
torsava, line 269:
```suggestion

The following tests are for multi-stage builds. These technically also work on full images, but there is no reason to do multi-stage builds with full images.

```

Makefile
torsava: The fact the script is in Python is an irrelevant implementation detail in this case, thus shebang makes sense IMHO.

generator.py
```diff
@@ -139,6 +141,10 @@ def main():
             elif section == "SYMLINK_RULES":
                 print(f"LN\t{spec['src']} → {spec['dest']}")
                 symlink(spec["src"], spec["dest"])
+                # Remove dead symlinks
+                if not spec["dest"].exists():
+                    print(f"WARN: {spec['dest']} is a dead symlink, removed.")
+                    unlink(spec["dest"])
```
torsava, line 149: This should be done after all symlinks are already processed, to detect chain symlinks correctly.

frenzymadness: Well. Dead symlinks are covered because symlinks should be created after all other files are copied or generated by distgen. That's the reason why the symlink section is the last one in the manifest config. Going one level deeper, if you want to create a symlink pointing to symlink, you just have to make sure that they're in the manifest config file in the correct order.
We don't need that functionality now and if you use it that way and without the correct order, the generator gives you a warning about it so you'd know what is happening. I think that's enough for now.

frenzymadness:
Rebased. I'm happy with how the commit history looks now.

frenzymadness:
[test]

phracek:
This pull request affects also https://github.com/sclorg/container-common-scripts/pull/275 and `varnish-container` and `postgresql-container` which uses manifest.sh

generator.py
torsava: Sounds good.

torsava:
I've done a sanity check of all the commits again. All the previous discussions were successfully addressed. I've pinpointed one typo, but besides that all looks good to me!

However, be warned that if you remove CCCP, Putin might attack you in the next ~30 years. Beware! :smiley:

PR #544, opened 2022-09-17
Set VIRTUAL_ENV in venv_setup

michael131468:
If poetry is used to install dependencies in a Dockerfile, it will not obey the virtualenv set up in the container because it does not detect it properly. The dependencies end up installed into the system python paths (eg. /usr/local/lib/python3.10/site-packages/).

This shows differing behaviour to pip which if used in the Dockerfile to install dependencies will install them correctly into the virtual env path (eg. /opt/app-root/lib/python3.10/site-packages/).

Setting VIRTUAL_ENV manually in the Dockerfile gives Poetry the needed info to obey the virtual env paths.

### DIFF
--- a/2.7/Dockerfile.rhel8
+++ b/2.7/Dockerfile.rhel8
@@ -72,6 +72,7 @@ RUN \
 # so virtualenv needs to be activated this way
 ENV BASH_ENV="${APP_ROOT}/bin/activate" \
     ENV="${APP_ROOT}/bin/activate" \
+    VIRTUAL_ENV="${APP_ROOT}" \
     PROMPT_COMMAND=". ${APP_ROOT}/bin/activate"
 
 USER 1001
--- a/3.10/Dockerfile.fedora
+++ b/3.10/Dockerfile.fedora
@@ -81,6 +81,7 @@ fix-permissions ${APP_ROOT} -P
 # so virtualenv needs to be activated this way
 ENV BASH_ENV="${APP_ROOT}/bin/activate" \
     ENV="${APP_ROOT}/bin/activate" \
+    VIRTUAL_ENV="${APP_ROOT}" \
     PROMPT_COMMAND=". ${APP_ROOT}/bin/activate"
 
 USER 1001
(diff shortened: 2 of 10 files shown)


michael131468:
I didn't test this too deeply, just using the fedora container. I'm also unsure of any negatives by manually setting this variable in the Dockerfile. Setting this variable seems to fix usage of Poetry with the ubi images but since I'm unsure of any unintended side-effects as a result so open to discuss if it's the right fix or not.

frenzymadness:
Hello and thanks for the contribution! Could you please describe your workflow in more detail? AFAIK we currently don't support Poetry (at least directly) so I'm interested in how you adjust your dockerfiles. Do you create your own Dockerfile and use the S2I Python as a base?

The virtual environment in the container is activated by: https://github.com/sclorg/s2i-python-container/blob/a44abddb965cee38c1acb46411b9f366dcd3b1a5/3.9/Dockerfile.rhel9#L85-L89 With these lines, virtualenv should be activated every time a bash session starts.

Also, do you know [micropipenv](https://github.com/thoth-station/micropipenv)? It's a minimalistic container-ready tool able to install Python packages from poetry.lock Pipfile.lock or requirements files.

frenzymadness:
As I'm thinking about this more, I see an inconsistency in this approach. You have `VIRTUAL_ENV` containing the path to the virtual environment but the `PATH` variable does not contain the path to `$APP_ROOT/bin`. So the virtual environment might look like it's activated but when you run `python` or `pip` their system versions from `/usr/bin/` are executed.

This might confuse other tools or users I'm afraid.

Is it possible for you to source the activation script as the first step in your workflow?

frenzymadness:
I'm closing this now but feel free to reopen it if you want to continue with the discussion.

michael131468:
Hi @frenzymadness, sorry, personal stuff popped up and this fell behind on my todo's. The workflow I am using is taking the rhel ubi python image as a base for a docker containerfile and then extending it in the Containerfile to install (using poetry) a source base. It was there I noticed this issue and when I dug in and looked at your s2i-python-container images I saw the same behaviour (while pip was working as expected).

Regarding your comment about PATH, this is set in the dockerfile already: https://github.com/sclorg/s2i-python-container/blob/master/3.9/Dockerfile.rhel8#L8 

I was thinking about your comment to instead source the environment during the run command and I think that makes sense and I'll have to try it. 

I'll leave this PR closed as after thinking further I believe we don't need to support this workflow in your project.

frenzymadness:
It should also be possible to use the Python in the virtual env directly os instead of calling `poetry install` you should be able to call `<path_to_venv>/bin/python -m poetry` and that should use the main virtualenv without explicit activation.

PR #547, opened 2022-10-05
use newest build-and-push action

zmiklank:
Main change is, that the distgen will no longer be called before build of the image.

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -14,36 +14,32 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - dockerfile_path: "3.8"
-            dockerfile: "Dockerfile"
+          - dockerfile: "3.8/Dockerfile"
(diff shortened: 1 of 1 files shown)


frenzymadness:
IIRC the distgen is not called before build here in Python container image. That's the reason why the new generator does not fail for Python in https://github.com/sclorg/container-common-scripts/pull/294

zmiklank:
@frenzymadness 

This is part of a different workflow; Unfortunately in build-and-push action was distgen called independently from common's Makefiles. See https://github.com/sclorg/build-and-push-action/commit/034c7fb9d6a72318794921c5ddf197e4518624d6.

frenzymadness:
You are right, thanks for the context. I trust you here so feel free to merge this as you wish.

zmiklank:
Ok, lets see :).

zmiklank:
It seems to work.

PR #549, opened 2022-10-13
ubi9: Pull in krb5-devel

frantisekz:
Otherwise, gssapi (needed by requests) fails to build from pypi with:

```
Downloading gssapi-1.8.1.tar.gz (94 kB)
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 94.3/94.3 kB 293.6 MB/s eta 0:00:00
Installing build dependencies: started
Installing build dependencies: finished with status 'done'
Getting requirements to build wheel: started
Getting requirements to build wheel: finished with status 'error'
error: subprocess-exited-with-error
× Getting requirements to build wheel did not run successfully.
│ exit code: 1
╰─> [21 lines of output]
/bin/sh: line 1: krb5-config: command not found
```


frenzymadness:
Thanks @frantisekz for the contribution. I'll take a look at it soon. I'll switch this PR to draft because we don't change the versioned folders directly but their content is generated by distgen. See: https://github.com/sclorg/s2i-python-container#contributing

You don't have to do it now. If we decide to include the new package, I'll prepare it.

frantisekz:
Uff, thanks! and sorry, I could have read through the readme at least :(

pkubatrh:
The package is already in UBI and the installed size is rather small (DNF says 1MB) so it should not increase the base image by a lot. Makes sense to me to include it, for the full sized image at least.

frenzymadness:
Let's see what happens if we add the package everywhere [test-all]

I'll also add a test for it later.

frenzymadness:
Interesting is that the gsspai maintainers provide pre-compiled binary wheels for windows and mac but not linux. If you want to know why read this: https://github.com/pythongssapi/python-gssapi/issues/200 

I'm almost done fixing almost all the CI jobs in #553 . After that, I'm gonna rebase this PR and add gssapi package to one of the example applications to test its installation and we should be soon ready to merge this.

frenzymadness:
[test-all]

frenzymadness:
The problems in CI with Centos 7 are expected and nothing we should worry about. @frantisekz thanks for the contribution! Do you want to test it or take a look at it before I merge it?

frantisekz:
> The problems in CI with Centos 7 are expected and nothing we should worry about. @frantisekz thanks for the contribution! Do you want to test it or take a look at it before I merge it?

Looks good and gets the job done, thanks!

PR #557, opened 2022-11-21
create separate job for distgen check in container-tests

zmiklank:
the testing job itself would then depend on the distgen-check job, and will not run, if distgen-check fails.
Also status report for distgen check job is added.

### DIFF
--- a/.github/workflows/container-tests.yml
+++ b/.github/workflows/container-tests.yml
@@ -3,7 +3,56 @@ on:
     types:
       - created
 jobs:
+  distgen-check:
+    name: "Check distgen generated files"
+    runs-on: ubuntu-20.04
+    if: |
+      github.event.issue.pull_request
+      && (contains(github.event.comment.body, '[test]') || contains(github.event.comment.body, '[test-all]'))
+      && contains(fromJson('["OWNER", "MEMBER"]'), github.event.comment.author_association)
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v2
+        with:
+          ref: "refs/pull/${{ github.event.issue.number }}/head"
+          submodules: true
+
+      - name: Check distgen generated files
+        id: check
+        shell: bash
+        run: |
+          sudo apt update && sudo apt -y install python3-pip
+          pip3 install pyyaml distgen Jinja2
+          result="passed"
+          ./common/tests/check_distgen_generated_files.sh || result="failed"
+          echo "result=$result" >> "$GITHUB_OUTPUT"
+
+      - name: Update status of the distgen check to the PR
+        shell: bash
+        run: |
+          _sha=$(git rev-parse HEAD)
+          cat << EOF > result.json
+          {
+            "sha": "$_sha",
+            "state": "pending",
+            "context": "Distgen check",
+            "description": "${{ steps.check.outputs.result }}",
+          }
+          EOF
+          cat result.json
+          curl -X POST \
+            -H "Authorization: Bearer ${{ inputs.github_token }}" \
+            -H "Accept: application/vnd.github.v3+json" \
+            https://api.github.com/repos/$GITHUB_REPOSITORY/statuses/"$_sha" }} \
+            --data @result.json > update.json
+          cat update.json
+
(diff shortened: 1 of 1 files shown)


.github/workflows/container-tests.yml
```diff
@@ -3,7 +3,56 @@ on:
...
+        with:
+          ref: "refs/pull/${{ github.event.issue.number }}/head"
+          submodules: true
+
+      - name: Check distgen generated files
+        id: check
+        shell: bash
+        run: |
+          sudo apt update && sudo apt -y install python3-pip
+          pip3 install pyyaml distgen Jinja2
```
frenzymadness, line 25: distgen depends on pyyaml and Jinja2 so why do we need to list them here?

frenzymadness (commented):
Just one note from me. Feel free to merge it.

.github/workflows/container-tests.yml
zmiklank: Hmm, I have added Jinja2, because distgen generation has failed without explicitly specifying it...
But, as you say it, I am not sure why this happened...
However I would rather leave it here explicitly, if you do not mind, as it does not break anything.

.github/workflows/container-tests.yml
```diff
@@ -3,7 +3,56 @@ on:
...
+        uses: actions/checkout@v2
+        with:
+          ref: "refs/pull/${{ github.event.issue.number }}/head"
+          submodules: true
+
+      - name: Check distgen generated files
+        id: check
+        shell: bash
+        run: |
+          sudo apt update && sudo apt -y install python3-pip
```
phracek, line 24: Are you sure `sudo apt update` does not need `-y`?

.github/workflows/container-tests.yml
```diff
@@ -3,7 +3,56 @@ on:
...
+            "state": "pending",
+            "context": "Distgen check",
+            "description": "${{ steps.check.outputs.result }}",
+          }
+          EOF
+          cat result.json
+          curl -X POST \
+            -H "Authorization: Bearer ${{ inputs.github_token }}" \
+            -H "Accept: application/vnd.github.v3+json" \
+            https://api.github.com/repos/$GITHUB_REPOSITORY/statuses/"$_sha" }} \
```
phracek, line 46: `$_sha` is GitHub variable? I have missed that.

.github/workflows/container-tests.yml
```diff
@@ -3,7 +3,56 @@ on:
...
+            --data @result.json > update.json
+          cat update.json
```
phracek, line 48: Any reason for this?

.github/workflows/container-tests.yml
```diff
@@ -3,7 +3,56 @@ on:
     types:
       - created
 jobs:
+  distgen-check:
```
phracek, line 6: Why not separate GitHub Action?

phracek (changes requested):
Thanks for bringing this PR. 
I would prefer to have it as a separate GitHub action and not in the same as `container-tests.yaml`
See my proposal.

.github/workflows/container-tests.yml
zmiklank: I used the same comman as is used in TFaGA: https://github.com/sclorg/testing-farm-as-github-action/blob/09d7e30a45da45249d0688a1e3075e738d3b5316/action.yml#L98

.github/workflows/container-tests.yml
zmiklank: On the line 33.

.github/workflows/container-tests.yml
zmiklank: Only debug reasons.

.github/workflows/container-tests.yml
zmiklank: Because the is no possibility currently to make multiple Github Actions depend on each other.
And we want to run container tests only when the distgen check passes.

zmiklank: But I could make a composite GHA from this, if this works.

PR #558, opened 2022-11-21
create separate job for distgen check in openshift-tests

zmiklank:
The distgen check in container tests seems to run, so I am adding also check for the openshift tests.

The job for openshift tests will not create a status report for the PR, to avoid duplicate status. The meaning of this is not to run the openshift tests if the files are not regenerated properly.


phracek (approved):
Thanks for the PR! I did not hit any issues. LGTM.

PR #575, opened 2022-12-19
Install also projects with pyproject.toml

### LINKED ISSUE #560: Poetry backed project is not installed
Hi, thanks for this awesome repository.

Recently within our team, we switched to use poetry.

Now, when using the s2i on our openshift environment (coupled with env var USE_MICROPIPENV=1) the application builds properly, but once the run command is triggered, we receive a `ModuleNotFoundError` stating that the root package is not found.

After some debugging, I noticed that the problem here might be [this line in the assemble script](https://github.com/sclorg/s2i-python-container/blob/master/3.9/s2i/bin/assemble#L88). It only runs `pip install .` whenever a setup.cfg or setup.py file is found. Since the addition of pyproject.toml, pip can also handle that.

Therefore, I think the assemble script could be expanded to not only check for the presence of setup.py or setup.cfg, but also include pyproject.toml

If you see any issues with this proposal, do you have a way for us to run the `pip install .` right after calling assemble, but before calling run? When attempting to override the dockerfile and adding the `pip install .`, this command is run BEFORE calling the assemble script, so that is not a solution either.

INFO: We use the python39 ubi8

If there is further information I can provide, feel free to ask for it :) 
Hope to hear from you soon.

Kind regards

frenzymadness:
Fixes #560

@torsava and @befeleme, do you wanna take a look?

### DIFF
--- a/2.7/README.md
+++ b/2.7/README.md
@@ -217,9 +217,9 @@ file inside your source code repository.
     Set this variable to a non-empty value to inhibit the execution of
     'manage.py collectstatic' during the build. This only affects Django projects.
 
-* **DISABLE_SETUP_PY_PROCESSING**
+* **DISABLE_SETUP_PY_PROCESSING** / **DISABLE_PYPROJECT_TOML_PROCESSING**
 
-    Set this to a non-empty value to skip processing of setup.{py,cfg} script if you
+    Set this to a non-empty value to skip processing of setup.{py,cfg} or pyproject.toml script if you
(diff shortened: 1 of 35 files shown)


shifqu:
Original author of issue #560 here.

LGTM, solution is fairly simple and I like the fact that it is configurable through the `DISABLE_PYPROJECT_TOML_PROCESSING` flag.
Thanks again for this PR

PR #579, opened 2023-02-09
3.9 minimal on C9S

frenzymadness:
This PR adds Python 3.9 minimal image for Centos stream 9.


frenzymadness:
[test-all]

frenzymadness:
@zmiklank hi, the tests for the new image are pretty good. There is only one I need to fix but there seems to be a problem with `$IMAGE_NAME` variable. I have no idea why is this variable set to `quay.io/ubi9/python-39-minimal:1` when it should be `quay.io/sclorg/python-39-minimal-c9s:0`. Any ideas?

pkubatrh:
>I have no idea why is this variable set to quay.io/ubi9/python-39-minimal:1 when it should be quay.io/sclorg/python-39-minimal-c9s:0. Any ideas?

My guess is its because you have it named that way:
https://github.com/sclorg/s2i-python-container/pull/579/files#diff-34d2b8de972391dd38249223ece3dcfed032e6bc887cd280c37d945a1e273c68R50

frenzymadness:
[test]

frenzymadness:
[test]

frenzymadness:
[test]

frenzymadness:
[test]

frenzymadness:
[test]

frenzymadness:
@zmiklank is the CI config enough to publish this image to quay?

@pkubatrh @torsava I'd appreciate a review.

zmiklank:
> @zmiklank is the CI config enough to publish this image to quay?

Yes, it is sufficient for publishing, but still the repository on quay.io needs to be created for `python-39-minimal-c9s` image. I will do that.

However, the tests are not enabled (this is to be done in `.github/workflows/container-tests.yml`). I am not sure if this is intended or not.

frenzymadness:
> However, the tests are not enabled (this is to be done in `.github/workflows/container-tests.yml`). I am not sure if this is intended or not.

Are you sure? I see Testing Farm - CentOS Stream 9 - 3.9-minimal job in the CI results. The file probably already contains the combination I'm adding here.

zmiklank:
> Are you sure? I see Testing Farm - CentOS Stream 9 - 3.9-minimal job in the CI results. The file probably already contains the combination I'm adding here.

You are right, they are already present. Sorry for the confusion.

src/Dockerfile-minimal.template
```diff
@@ -49,7 +49,7 @@ LABEL summary="$SUMMARY" \
       io.openshift.expose-services="8080:http" \
       io.openshift.tags="builder,python,python{{ spec.short_ver }},python-{{ spec.short_ver }},rh-python{{ spec.short_ver }}" \
       com.redhat.component="python-{{ spec.short_ver }}-container" \
-      name="{{ spec.org }}/python-{{ spec.short_ver }}-minimal" \
+      name="{{ spec.org }}/python-{{ spec.short_ver }}-{% if spec.org == "sclorg" %}{{ spec.prod }}-{% endif %}minimal" \
```
torsava, line 52: Ad `python-39-c9s-minimal`: Here's where it should be changed if we decide to do so.

test/run
```diff
@@ -279,11 +279,15 @@ test_from_dockerfile_minimal(){
...
-  check_result $?
+  if ct_pull_image "$FULL_IMAGE_NAME"; then
+    # mod_wsgi in micropipenv-requirements-test-app
+    sed "s@#IMAGE_NAME#@${IMAGE_NAME}@;s@#FULL_IMAGE_NAME#@${FULL_IMAGE_NAME}@" $test_dir/from-dockerfile/mod_wsgi.Dockerfile.tpl > $test_dir/from-dockerfile/Dockerfile
+    ct_test_app_dockerfile $test_dir/from-dockerfile/Dockerfile $test_dir/micropipenv-requirements-test-app 'Hello World from mod_wsgi hosted WSGI application!' app-src
+    check_result $?
+  else
+    echo "[SKIP] Multistage build from Dockerfile - $FULL_IMAGE_NAME does not exists."
+  fi
+
```
torsava, line 290: Commit title: If full image does not exists, skip the multistage build test

My question is: If the full image does not exist, then there's likely to be some problem with the way we detect the FULL_IMAGE_NAME. Or why do you suppose the full image would not exist?

torsava (commented):
Finished first review, opened some comments.

specs/multispec.yml
```diff
@@ -204,6 +205,7 @@ matrix:
       version: "3.9"
     - distros:
         - rhel-8-x86_64
+        - centos-stream-9-x86_64
```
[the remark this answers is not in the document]
frenzymadness: Glad it's useful.

3.9-minimal/Dockerfile.c9s
```diff
@@ -47,7 +47,7 @@ LABEL summary="$SUMMARY" \
       io.openshift.expose-services="8080:http" \
       io.openshift.tags="builder,python,python39,python-39,rh-python39" \
       com.redhat.component="python-39-container" \
-      name="sclorg/python-39-minimal" \
+      name="sclorg/python-39-c9s-minimal" \
```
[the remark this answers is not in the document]
frenzymadness: 1. You are right for rhel7 and centos7, ubi8 and 9 have it at the beginning before the first slash. From the point of view that we're adding a new component to the name, I'd add it as the last one.
2. It does not matter because all minimal Python containers are and always will be based on the minimal base containers. And from my point of view, it does not make sense to build a full container image on a minimal base image or vice versa. That's the reason I was waiting for the minimal C9S container to become available.

test/run
frenzymadness: For example, right now, we are preparing the minimal Python container image based on C9S (per request) but the full image for the same combination of Python version and platform does not exist. I see this as a possible future if the minimal images will be popular, we might start producing them more and more without their full counterparts.

3.9-minimal/Dockerfile.c9s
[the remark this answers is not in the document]
pkubatrh: For what its worth, Red Hat naming policies for container images say the platform should always be a suffix to the name of the image if not already present in the namespace (like in the case of ubiX/ and rhelX/) images. So I would advocate for `python-39-minimal-c9s` as well.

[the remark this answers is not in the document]
zmiklank: Also the repository I've created on quay.io is named `python-39-minimal-c9s` because that is also the image name you specified in the build and push action. This can be changed though.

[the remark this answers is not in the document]
frenzymadness: The name is fixed to the final version `python-39-minimal-c9s`. Thank you all for the inputs!

frenzymadness:
[test]

zmiklank:
> @zmiklank hi, the tests for the new image are pretty good. There is only one I need to fix but there seems to be a problem with `$IMAGE_NAME` variable. I have no idea why is this variable set to `quay.io/ubi9/python-39-minimal:1` when it should be `quay.io/sclorg/python-39-minimal-c9s:0`. Any ideas?

I did not notice this comment. Is this still a problem?

test/run
torsava: However, if we don't have the full counterparts, the multistage build feature won't be possible, and they might decrease in popularity.

torsava: Discussed in person: Let's try to release just the minimal for c9s and see the reactions.

pkubatrh: Note that c9s (and c8s) full sized images are not planned to be created for Python, as the UBI images should cover most of the use-cases. So if the multistage build is important for the Python images, we should likely make it work with the fullsized ubi images in future.

test/run
```diff
@@ -267,17 +267,27 @@ test_from_dockerfile(){
...
   ct_test_app_dockerfile $test_dir/from-dockerfile/uwsgi.Dockerfile.tpl $test_dir/uwsgi-test-app 'Hello World from uWSGI hosted WSGI application!' app-src
   check_result $?
 
-  # mod_wsgi in micropipenv-requirements-test-app
-  # TODO: Find a better way how to specify full/minimal images that should work together
-  sed "s@#IMAGE_NAME#@${IMAGE_NAME}@;s@#FULL_IMAGE_NAME#@ubi{{ spec.el_version }}/python-{{ spec.short_ver }}:1@" $test_dir/from-dockerfile/mod_wsgi.Dockerfile.tpl > $test_dir/from-dockerfile/Dockerfile
-  ct_test_app_dockerfile $test_dir/from-dockerfile/Dockerfile $test_dir/micropipenv-requirements-test-app 'Hello World from mod_wsgi hosted WSGI application!' app-src
-  check_result $?
+  # So far, for all the minimal images, the name of the full container image counterpart
+  # is the same just without -minimal suffix.
```
torsava, line 277: Minor nitpick: Now the `-minimal` is an "infix".

test/run
torsava: Sounds good. However, I don't consider this a blocker for this PR. We can implement it later if/when we get some feedback wanting multistage builds.

torsava:
Besides the minor nitpick, the PR looks good to me!

frenzymadness:
[test]

Issue #580, opened 2023-02-14
3.9 Readme Instructions unclear


Joshua-Beha:
Building the sample app in OpenShift with the command specified by the readme (`oc new-app python:3.9~https://github.com/sclorg/django-ex.git`) fails with error `error: the image match "python-ubi/python:3.9" for source repository "https://github.com/sclorg/django-ex" does not appear to be a source-to-image builder.`. 
Curling the image registry (`curl -X GET -u user:passwd https://<registry-host>/v2/openshift/python/tags/list`) returns the following tags:
```
{
  "name": "openshift/python",
  "tags": [
    "3.8",
    "3.8-ubi8",
    "3.9-ubi9",
    "3.9-ubi8",
    "latest",
    "2.7",
    "2.7-ubi7",
    "2.7-ubi8",
    "3.6-ubi8",
    "3.8-ubi7"
  ]
}
```
Notably, this does *not* include the `3.9` tag.
The cluster has not had any changes to the image registry to the best of my knowledge. Could the README be updated to use a command that would work by default such as `oc new-app python:3.9-ubi8~https://github.com/sclorg/django-ex.git`? Alternatively, a further explanation could be provided as to why the `python:3.9` tag is not available by default.

pkubatrh:
The default image streams are afaik tied to the release of OCP/Openshift. So you might have an older version that does not have it yet, or the latest version of Openshift did not integrate it yet.

The README expects you have the image stream available. If you do not, you can import ones maintained in this repository:  https://github.com/sclorg/s2i-python-container/tree/master/imagestreams

But I agree it makes sense to explicitly mention this in the README

frenzymadness:
I'm sorry, but my knowledge of OpenShift is limited. Could please any of you open a PR here if there is anything to fix in the readme?

pkubatrh:
Looking at this a second time (after getting some coffee) I see that the issue is not in the version of Openshift, but that the imagestream does not provide the `3.9` stream, but does provide `3.9-ubi8` and `3.9-ubi9` streams. I vaguely remember that this was done to actually force the users to differentiate between what platform they want to use.
So yes, we should update the README with the correct stream names, so that it works out of the box.

Joshua-Beha:
@pkubatrh Forcing users to choose a platform would make sense. Any chance you would have a link to documentation explaining this? I can create the PR but I think having some documentation of the change on the OpenShift end would help to justify the change.

Joshua-Beha:
[...]

PR #582, opened 2023-02-20
Update python 3.9 imagestream tag

Joshua-Beha:
Change the imagestream tag such that the OpenShift command works out of the box as discussed in [issue 580](https://github.com/sclorg/s2i-python-container/issues/580)

### DIFF
--- a/3.9/README.md
+++ b/3.9/README.md
@@ -32,12 +32,12 @@ the nodejs itself is included just to make the npm work.
 Usage in Openshift
 ------------------
 
-For this, we will assume that you are using the supported image, available via `python:3.9` imagestream tag in Openshift.
+For this, we will assume that you are using one of the supported images, available via `python:3.9-ubi8` or `python:3.9-ubi9` imagestream tag in Openshift.
 Building a simple [python-sample-app](https://github.com/sclorg/django-ex.git) application
 in Openshift can be achieved with the following step:
 
     ```
-    oc new-app python:3.9~https://github.com/sclorg/django-ex.git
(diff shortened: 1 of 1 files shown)


frenzymadness:
Thank you @Joshua-Beha for the contribution. The file `3.9/README.md` is generated from the readme template in `src` file. I'll handle that. @pkubatrh could you please verify that this is okay and make sense to add it to all readmes?

Joshua-Beha:
@pkubatrh just looking to follow up here

3.9/README.md
```diff
@@ -32,12 +32,12 @@ the nodejs itself is included just to make the npm work.
 Usage in Openshift
 ------------------
 
-For this, we will assume that you are using the supported image, available via `python:3.9` imagestream tag in Openshift.
+For this, we will assume that you are using one of the supported images, available via `python:3.9-ubi8` or `python:3.9-ubi9` imagestream tag in Openshift.
```
pkubatrh, line 35: Let's just use a single example here rather than a list. The users can check themselves which images is available to them.
So something like:
```suggestion
For this, we will assume that you are using one of the supported images available via imagestream tags in Openshift, eg. `python:3.9-ubi9`.
```

frenzymadness:
Thank you @pkubatrh and @Joshua-Beha. I've prepared #591 with one small modification of Petr's proposal - I use ubi8 there because almost all Python versions we currently have are available in RHEL 8 - the only one where this info is not correct is Python 3.10 and we'll remove that version when Fedora 36 goes EOL which will happen in 3 months.

Could you please take a look at #591 ?

PR #588, opened 2023-02-27
Proper cleanup

zmiklank:
Use `ct_init` function for preparing the test environment.
Use generic function for checking the testcase result.


frenzymadness:
I trust your expertise and I like the changes so if they work, feel free to ship them. Should we run tests first?

zmiklank:
> I trust your expertise and I like the changes so if they work, feel free to ship them. Should we run tests first?

Yes, I'll trigger them.

[test]

PR #589, opened 2023-02-27
tests: use tfaga wrapper

zmiklank:
Tfaga wrapper is a wrapper for sclorg's testing farm workflows. [1]

This was already tested in my fork and merged to nginx-container [2].
However, the frequency of PRs in the python container is higher so might be a better candidate for onboarding.
I think the problems were caught during testing in fork, but some issues can come through also now.

[1] https://github.com/sclorg/tfaga-wrapper
[2] https://github.com/sclorg/nginx-container/pull/235

### DIFF
--- a/.github/workflows/container-tests.yml
+++ b/.github/workflows/container-tests.yml
@@ -1,3 +1,5 @@
+---
+
 on:
   issue_comment:
     types:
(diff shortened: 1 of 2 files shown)


frenzymadness:
Same here. You know what you're doing so if it works, ship it.

PR #591, opened 2023-02-28
Update imagestream tag in README

frenzymadness:
Based on #582


pkubatrh:
The change to ubi8 example is ok in my book. the only reason we used ubi9 in the README was that RHEL9 will be with us a bit longer. :)

Joshua-Beha:
@frenzymadness looks good to me!

frenzymadness:
Thank you all for your help!

Issue #593, opened 2023-03-13
tests: test case should fail early when the prepare function fails


pkubatrh:
I encountered this specifically when the image was not available (forgot to pull it from registry). The test suite prints out an error (https://github.com/sclorg/s2i-python-container/blob/master/test/run#L85) but continues running since the return value is never checked.

frenzymadness:
PRs welcome :wink:

zmiklank:
Yes, currently is not possible to simply end the testsuite, if the preparation is not possible.
Therefore the prepare function return 1 and it gets [FAILED] flag in the results.
However, the bug that I see here is, that it does not TESTCASE_RESULT to 1 so in theory the testsuite could still succeed. (Is this also the core of issue you encountered?)

However, as the app image could not be properly build I think the rest of the tests would fail anyway, and also fail the testsuite. Or haven't they?

I agree, that we should fix this. But we should, IMO, do it properly and do not run more tests for this app if the image build fails.

zmiklank:
We should also find out, if this is python-specific, or we should adjust also some other containers. I did not look into that yet.

PR #595, opened 2023-03-21
fail tests for specific app when prep. fails

zmiklank:
Fixes: https://github.com/sclorg/s2i-python-container/issues/593

Depends on: https://github.com/sclorg/container-common-scripts/pull/338

Also I have a question. Could it make sense to have the test/run files generated by distgen only as symlinks?

### DIFF
--- a/2.7/test/run
+++ b/2.7/test/run
@@ -303,19 +303,23 @@ for app in ${@:-${WEB_APPS[@]}}; do
   fi
 
   prepare ${app}
+  if [ $? -ne 0 ]; then
+    ct_update_test_result "[FAILED]" "${app}" "preparation"
+    TESTSUITE_RESULT=1
+    continue
+  fi
   run_s2i_build ${app}
   RESULT=$?
-  msg_run_s2i_build="'${app}' run_s2i_build"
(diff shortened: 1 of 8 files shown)


frenzymadness:
> Also I have a question. Could it make sense to have the test/run files generated by distgen only as symlinks?

It's not possible right now because the files look different for full and minimal container images, and for Python 2 and 3+ images etc. (search for `{%` to see all templating done in the source file) but that might be something we can solve by some bash logic instead of the logic in the template we have there now. What would be the benefit of doing that?

zmiklank:
> > Also I have a question. Could it make sense to have the test/run files generated by distgen only as symlinks?
> 
> It's not possible right now because the files look different for full and minimal container images, and for Python 2 and 3+ images etc. (search for `{%` to see all templating done in the source file) but that might be something we can solve by some bash logic instead of the logic in the template we have there now. What would be the benefit of doing that?

Right, if there is some reason behind it I am OK with that, thanks for explanation.

zmiklank:
[test-all]

Lupphes (approved):
LGTM, tests passed

Issue #596, opened 2023-03-30
Incorrect py-3 image


teplyakoff:
### Container platform

Podman/Docker

### Version

latest

### OS version of the container image

RHEL 8

### Description

Image doesn't contain actual code for /s2i/bin/assemble
`$ docker pull registry.access.redhat.com/ubi8/python-39 && docker run registry.access.redhat.com/ubi8/python-39 cat /usr/libexec/s2i/assemble`

You can check that container doesn't contain this code in /s2i/bin/assemble:
```
if [[ -f pyproject.toml && -z "$DISABLE_PYPROJECT_TOML_PROCESSING" ]]; then
  echo "---> Installing application (via pyproject.toml)..."
  pip install .
fi
```

Same issue with **python 3.10** (https://github.com/sclorg/s2i-python-container/tree/master/3.10#3-prepare-an-application-inside-a-container)
`$ docker pull registry.fedoraproject.org/f35/python3 && docker run registry.fedoraproject.org/f35/python3 cat /usr/libexec/s2i/assemble`

python 3.11 doesn't work at all.. (https://github.com/sclorg/s2i-python-container/tree/master/3.11#3-prepare-an-application-inside-a-container)
```
docker pull registry.fedoraproject.org/f37/python3
Using default tag: latest
Error response from daemon: manifest for registry.fedoraproject.org/f37/python3:latest not found: manifest unknown: manifest unknown
```

### Reproducer

docker pull registry.access.redhat.com/ubi8/python-39 && docker run registry.access.redhat.com/ubi8/python-39 cat /usr/libexec/s2i/assemble

frenzymadness:
_I was writing my response when Petr closed the issue so I'm sending it anyway. Petr's reply still applies for UBI container images._

Thanks a lot for the report!

Let me take a look at what happened here. It seems that the code is there for 3.9: https://github.com/sclorg/s2i-python-container/blob/1f10d8205fd66a5ccbfe651307f9860893e930b2/3.9/s2i/bin/assemble#L93-L96 So the images are outdated for some reason.

The change is 2 months old and:
* ubi8/python-39 container image is 3 days old so it should be there but isn't.
* ubi9/python-39 is one month old so it should be there as well but isn't.

As Petr mentioned, it takes some time for the internal images to get rebuilt. The container images based on Centos Stream or Fedora are much newer so you can use those.

[...]

PR #597, opened 2023-03-30
Update README with actual info about image versions and locations

zmiklank:
Resolves: https://github.com/sclorg/s2i-python-container/issues/596


README.md
```diff
@@ -5,8 +5,8 @@ Python container images
 Images available on Quay are:
 * CentOS 7 [python-38](https://quay.io/repository/centos7/python-38-centos7)
 * RHEL 8 [python-39-minimal](https://quay.io/repository/sclorg/python-39-minimal-el8)
-* CentOS Stream 9 [python-39-c9s](https://quay.io/repository/sclorg/python-39-c9s)
```
frenzymadness, line 8: It seems that we have that Dockerfile in 3.9 so we can either remove it there as well (from multispec) or add it to build and push action. What do you think?

frenzymadness (commented):
I'm interested in your opinion about C9S with Python 3.9. Python 3.9 is something we'll maintain for a long time because it's the main Python in RHEL 9 so I'd keep it here also as C9S image.

README.md
frenzymadness: The biggest advantage I see now here is that we have 3.9-minimal based on C9S so the full image on the same platform can be used as a build base image which is something we have documented and supported.
I'd keep it. Could you please add it to the build and push GH action and keep it in the readme?

specs/multispec.yml
```diff
@@ -128,7 +128,7 @@ specs:
       python_img_version: "1"
       module_stream: "python311:3.11"
       pkg_prefix: "python311"
-      main_image: "registry.fedoraproject.org/f37/python3"
+      main_image: "quay.io/fedora/python-311"
```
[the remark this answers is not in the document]
frenzymadness: For what? We publish that image for months now. See https://quay.io/repository/fedora/python-311?tab=tags&tag=latest

README.md
pkubatrh: Yes, we opted to not build c9s images when there is a UBI version available, since they are used for the same use case. There is no reason to stop building it other than that (anyone in the community can build their own anyway), so if you are ok with maintaining it on top of the UBI image, then I am ok with that.

frenzymadness:
Thank you!

PR #602, opened 2023-05-18
Prevent multiple virtual environment activation

frenzymadness:
It seems it's not necessary anymore.

### DIFF
--- a/2.7/Dockerfile.rhel8
+++ b/2.7/Dockerfile.rhel8
@@ -66,7 +66,12 @@ RUN \
     rm -r /opt/wheels && \
     chown -R 1001:0 ${APP_ROOT} && \
     fix-permissions ${APP_ROOT} -P && \
-    rpm-file-permissions
+    rpm-file-permissions && \
+    # The following echo adds the unset command for the variables set below to the \
+    # venv activation script. This is inspired from scl_enable script and prevents \
+    # the virtual environment to be activated multiple times and also every time \
+    # the prompt is rendered. \
+    echo "unset BASH_ENV PROMPT_COMMAND ENV" >> ${APP_ROOT}/bin/activate
 
 # For RHEL/Centos 8+ scl_enable isn't sourced automatically in s2i-core
 # so virtualenv needs to be activated this way
--- a/3.11/Dockerfile.c8s
+++ b/3.11/Dockerfile.c8s
@@ -76,7 +76,12 @@ RUN python3.11 -m venv ${APP_ROOT} && \
 ${APP_ROOT}/bin/pip install /opt/wheels/pip-* && \
 rm -r /opt/wheels && \
 chown -R 1001:0 ${APP_ROOT} && \
-fix-permissions ${APP_ROOT} -P
+fix-permissions ${APP_ROOT} -P && \
+# The following echo adds the unset command for the variables set below to the \
+# venv activation script. This is inspired from scl_enable script and prevents \
+# the virtual environment to be activated multiple times and also every time \
+# the prompt is rendered. \
+echo "unset BASH_ENV PROMPT_COMMAND ENV" >> ${APP_ROOT}/bin/activate
 
 # For Fedora scl_enable isn't sourced automatically in s2i-core
 # so virtualenv needs to be activated this way
--- a/3.11/Dockerfile.c9s
+++ b/3.11/Dockerfile.c9s
@@ -76,7 +76,12 @@ RUN python3.11 -m venv ${APP_ROOT} && \
 ${APP_ROOT}/bin/pip install /opt/wheels/pip-* && \
 rm -r /opt/wheels && \
 chown -R 1001:0 ${APP_ROOT} && \
-fix-permissions ${APP_ROOT} -P
+fix-permissions ${APP_ROOT} -P && \
+# The following echo adds the unset command for the variables set below to the \
+# venv activation script. This is inspired from scl_enable script and prevents \
+# the virtual environment to be activated multiple times and also every time \
+# the prompt is rendered. \
+echo "unset BASH_ENV PROMPT_COMMAND ENV" >> ${APP_ROOT}/bin/activate
 
 # For Fedora scl_enable isn't sourced automatically in s2i-core
 # so virtualenv needs to be activated this way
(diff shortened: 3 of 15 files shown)


frenzymadness:
[test-all]

frenzymadness:
I'm testing the `$VIRTUAL_ENV` variable here because python from `/opt/app-root/bin` is always first in the `$PATH` with or without the active virtual environment because the path is hardcoded in s2i-base.

app.sh
```
#!/bin/bash

echo "Running bash script…"
echo "${VIRTUAL_ENV:-NOTSET}"
```

app.py
```
#!/usr/bin/env python

import os

print("Running Python app…")
print(os.environ.get("VIRTUAL_ENV", "NOTSET"))
```

## Old container image build from master branch (Python 3.9, RHEL 8)

### The following commands work fine and print the correct value:
```
podman run --rm -it test-old ./app.py
podman run --rm -it test-old ./app.sh

echo "" | podman run --rm -it test-old ./app.py
echo "" | podman run --rm -it test-old ./app.sh

# These two work fine but have "(app-root) (app-root)" in PS1:
podman run --rm -it test-old bash
podman run --rm -it test-old bash -l

# With bash running in named container…
podman exec -ti test ./app.sh
podman exec -ti test bash
podman exec -ti test bash -l
````

### These commands don't print the correct value:
```
podman run --rm -it test-old bash -c "echo ${VIRTUAL_ENV:-NOTSET}"

# With bash running in named container …
podman exec -ti test ./app.py
podman exec -ti test bash -c "echo ${VIRTUAL_ENV:-NOTSET}"
```

## New image without $PROMPT_COMMAND

### The following commands were working and still works:
```
podman run --rm -it test-new ./app.py
podman run --rm -it test-new ./app.sh

echo "" | podman run --rm -it test-new ./app.sh
echo "" | podman run --rm -it test-new ./app.py

# These two work fine and PS1 is fixed
podman run --rm -it test-new bash
podman run --rm -it test-new bash -l

# With bash running in named container…
podman exec -ti test ./app.sh
```

### These commands weren't working and nothing changed:
```
podman run --rm -it test-new bash -c "echo ${VIRTUAL_ENV:-NOTSET}"

# With bash running in named container …
podman exec -ti test ./app.py
podman exec -ti test bash -c "echo ${VIRTUAL_ENV:-NOTSET}"
```

### And the most important category: these commands were working and now they don't:
```
# With bash running in named container …
podman exec -ti test bash
podman exec -ti test bash -l
```

Question n1 is why is the second running bash instance in the same container ignores BASH_ENV/ENV variables (they are all set properly). Question n2 is whether this is a problem for us because python is properly set anyway.

In other words, could you image an application being run by one of the ways mentioned in the last category relying on the `VIRTUAL_ENV` env variable?

My opinion is that this is good to be merged as it fixes the issue with doubled activation and wrong PS1.

hhorak:
The `exec` is used in OpenShift to debug running containers -- `oc exec`, so the `exec` itself is important. Whether we are fine not to have `VIRTUAL_ENV` set there, I don't know. What would it mean for a developer, joining to the running container using `exec` either in `podman` or in OpenShift and not having the virtual env. set properly?

zmiklank:
>     3. Find a way how to prevent multiple activations of the same environment.

In `test-lib.sh` we use kind of ifdef approach, see [here](https://github.com/sclorg/container-common-scripts/blob/master/test-lib.sh#L14-L18). It could be used for activating virtualenv too in form of some wrapper or short script. I, however, would see it more as a hotfix in this case.

zmiklank:
>     2. Find a completely different way how to activate the environment.

Could not one way of activating virtualenv be to source the `bin/activate` manually in a RUN step when building the container image? I checked, and all the variables which are set in the `bin/activate` are exported to the environment.
Thus this solution should afaik work either for interactive (login and non-login) and non-interactive bash. Not tested in container though. Please correct me if otherwise.

zmiklank:
> ### These commands weren't working and nothing changed:
> 
> ```
> podman run --rm -it test-new bash -c "echo ${VIRTUAL_ENV:-NOTSET}"
> 
> # With bash running in named container …
> podman exec -ti test ./app.py
> podman exec -ti test bash -c "echo ${VIRTUAL_ENV:-NOTSET}"
> ```

Indeed, it does not seem that `$PROMPT_COMMAND` have any affect on non-interactive shells.

frenzymadness:
> > ```
> > 2. Find a completely different way how to activate the environment.
> > ```
> 
> Could not one way of activating virtualenv be to source the `bin/activate` manually in a RUN step when building the container image? I checked, and all the variables which are set in the `bin/activate` are exported to the environment. Thus this solution should afaik work either for interactive (login and non-login) and non-interactive bash. Not tested in container though. Please correct me if otherwise.

I'm not sure I understand you here but having `RUN source ${APP_ROOT}/bin/activate` does not export any of the variables and the virtual environment is not active when you run a container build with that in Dockerfile. Is that what you meant?

frenzymadness:
So, after testing multiple different solutions and playing around with them I end up with the solution inspired by scl_enable script. The solution adds `unset BASH_ENV PROMPT_COMMAND ENV` to the very end of the `bin/activate` script. With that `unset`, no matter which of these variables causes the environment activation, it should not happen more than once.

I've tested all commands mentioned earlier and all of them work the same as with the current state of things and the problem with `(app-root) (app-root)` being in PS1 is gone.

I wanted to implement something nicer without the need to alter the script created by venv/virtualenv modules but wasn't able to do so. I'm still open to better ideas if you have some.

frenzymadness:
[test]

zmiklank:
> > > ```
> > > 2. Find a completely different way how to activate the environment.
> > > ```
> > 
> > 
> > Could not one way of activating virtualenv be to source the `bin/activate` manually in a RUN step when building the container image? I checked, and all the variables which are set in the `bin/activate` are exported to the environment. Thus this solution should afaik work either for interactive (login and non-login) and non-interactive bash. Not tested in container though. Please correct me if otherwise.
> 
> I'm not sure I understand you here but having `RUN source ${APP_ROOT}/bin/activate` does not export any of the variables and the virtual environment is not active when you run a container build with that in Dockerfile. Is that what you meant?

Yes, you are right.

frenzymadness:
I did a rebase and will run one more round of testing in the CI. If nobody has a better idea of how to solve this issue, I'm going to merge it and we can improve it later.
[test]

frenzymadness:
[test-openshift]

frenzymadness:
[test-openshift]

frenzymadness:
The failures in the CI don't seem to be related to this change. Merging.

PR #629, opened 2023-08-08
stop rebuilding and testing of CentOS7 images

zmiklank:
The CentOS registry has been decommissioned[1]. Alternative registries provide only old versions of CentOS base image. Because there is no known possibility currently to obtain fresh CentOS7 base container image (we do not want to update old images with yum -y update) we are stopping rebuilds of the SCLorg's s2i-images too. Already built images will be still available on quay.io.

[1] https://lists.centos.org/pipermail/centos-devel/2023-May/142956.html


zmiklank:
> One more thing you can remove here is the centos7 distro config from the specfile. Otherwise, it looks good to me. Thank you!

Thank you for the review.

In non-distgen repos we 'deprecated' CentOS7 with adding .exclude- files. That way it would be easier to reenable building and testing of the images, since official EOL of CentOS7 is next year. 
We were thinking about alternative for this step in distgen repos, and the approach you see in this PR was closest to the .exclude- files system. Here is the discussion: https://github.com/sclorg/postgresql-container/pull/520.
The idea behind is, that it could be easy enough to revert the change in the future, if needed.
Would you agree with this approach with these arguments?

If not I am OK also with removing the conf file.

frenzymadness:
I 100 % understand you and I don't have a strong opinion here - just a note: if we ever revert this change, we probably do it on the commit level (`git revert`), and the revert can simply reintroduce the centos 7 section in the multispec as well. I'd prefer that over manually adding it back to the matrix. If we don't do it, we'll need one more cleanup PR/commit for the section. Does that make sense?

zmiklank:
> if we ever revert this change, we probably do it on the commit level (git revert), and the revert can simply reintroduce the centos 7 section in the multispec as well.

That is great argument, I agree.
This approach was chosen on postresql container, where more CentOS7 Dockerfiles were removed. In that case the revert can be also on file level and not commit level. In python case this does not make sense, as there is only one CentOS7 python image currently supported.

If you do not have strong opinion I would rather leave this PR it as it is, so we have the container repositories united in the approaches chosen and avoid unnecessary chaos -- I will be adding additional commits regarding this change next year to all of other repos anyway.

frenzymadness:
All right. Feel free to merge this then.

PR #561, opened 2022-12-06
Add Python 3.11 and Fedora 37 + some small updates

frenzymadness:

PR #643, opened 2023-10-11
Add Python 3.12 and Fedora 39

befeleme:
Opening for the CI.
I expect at least problems with pipenv, the Fedora's version breaks on missing setuptools.


befeleme:
[test]

befeleme:
I guess the CI has to be triggered by someone with the correct repository rights.

frenzymadness:
[test]

frenzymadness:
[test]

befeleme:
I found this discussion and realized that CI changes need to come first: https://github.com/sclorg/s2i-python-container/pull/561 to actually test the container. Let me rework this.

befeleme:
Let's start with https://github.com/sclorg/s2i-python-container/pull/644

phracek:
> Let's start with #644

The testing matrix is already merged. Let's start with support python-312. Go ahead.

phracek:
Let's start with first round of tests.

[test]

phracek:
Testing Farm should be fixed now, let's try it again

[test]

befeleme:
I started debugging `gunicorn-different-port-test-app`. The issue is apparently in the bundled pip. We've got `pip-22.2.2-py3-none-any.whl`. I tested the app with `pip-23.3-py3-none-any.whl` and it works seamlessly. But the question is what now. What's the policy of updating pip wheel in this repository? Won't the update destroy things for older Pythons? I see we currently have only pip-22.2.2 in the repository, but that's also too new for some of them. I will appreciate a pointer how to approach this.

zmiklank:
> I see a lot of errors which I don't understand - I hope they are related to the fact that the container doesn't exist on the remote repositories yet.
> 
> ```
> �[1m[INFO] Waiting for application container to start --user 12345 ...�[0m
> 
> Trying to pull quay.io/fedora/python3-312:0-testapp...
> Error: initializing source docker://quay.io/fedora/python3-312:0-testapp: reading manifest 0-testapp in quay.io/fedora/python3-312: unauthorized: access to the requested resource is not authorized
> Test for image 'quay.io/fedora/python3-312:0' FAILED (exit code: 1)
> 
> �[1m[INFO] Testing the image SCL enable�[0m
> 
> cat: /tmp/tmp.qb5xhb7ALI/./tmp.lHPCNvDUQZ.cid: No such file or directory
> ERROR[exec /bin/bash -c python --version] Expected 'Python 3.12.', got 'Error: no container with name or ID "bin/bash" found: no such container'
> Test for image 'quay.io/fedora/python3-312:0' FAILED (exit code: 1)
> 
> �[1m[INFO] Testing the image SCL enable�[0m
> 
> cat: /tmp/tmp.qb5xhb7ALI/./tmp.lHPCNvDUQZ.cid: No such file or directory
> ERROR[exec /bin/bash -c node --version] Expected '^v[0-9]*\.[0-9]*\.[0-9]*', got 'Error: no container with name or ID "bin/bash" found: no such container'
> Test for image 'quay.io/fedora/python3-312:0' FAILED (exit code: 1)
> 
> �[1m[INFO] Testing the image SCL enable�[0m
> 
> cat: /tmp/tmp.qb5xhb7ALI/./tmp.lHPCNvDUQZ.cid: No such file or directory
> ERROR[exec /bin/bash -c npm --version] Expected '^[0-9]*\.[0-9]*\.[0-9]*', got 'Error: no container with name or ID "bin/bash" found: no such container'
> Test for image 'quay.io/fedora/python3-312:0' FAILED (exit code: 1)
> cat: /tmp/tmp.qb5xhb7ALI/./tmp.lHPCNvDUQZ.cid: No such file or directory
> Error: no names or ids specified
> ```

The aim here is, afaik, not to pull remote image, but to use application image that has just been built.
I see the image build failed:
```
[1m[INFO] Building the gunicorn-config-different-port-test-app application image ...[0m

grep: warning: stray \ before -

[1m[INFO] TEST FAILED (positive), EXPECTED:0 GOT:1[0m
```

That could be the culprit here... I need to take a closer look though..

frenzymadness:
> I started debugging `gunicorn-different-port-test-app`. The issue is apparently in the bundled pip. We've got `pip-22.2.2-py3-none-any.whl`. I tested the app with `pip-23.3-py3-none-any.whl` and it works seamlessly. But the question is what now. What's the policy of updating pip wheel in this repository? Won't the update destroy things for older Pythons? I see we currently have only pip-22.2.2 in the repository, but that's also too new for some of them. I will appreciate a pointer how to approach this.

Python 2.7 uses `pip<21` because old pip doesn't understand `python_requires`, Python 3.6 uses pip installed from PyPI in the version that still supports it and the rest (3.7+) uses the bundled wheel. You can update the wheel because the latest pip still supports Python 3.7. You can take 22.3.1 from Fedora 38 or 23.2.1 from Fedora 39.

zmiklank:
We currently do not test the specific app at all if prep phase fails. I think we should implement similar approach if the s2i-build fails.

befeleme:
I pushed another commit with the pip 23.2.1 from Fedora 39. Please rerun tests, thank you!

frenzymadness:
I'm going to test this PR locally. In the meantime, we can let CI do it as well. [test]

frenzymadness:
I did a local build via `podman build` as well as `make tag` and tested the resulting image with one of the example apps we have here and it works fine for me. I have to take a look at what is wrong here because it doesn't seem to be a problem in the image files itself.

befeleme:
One step further.
I'm looking at container with built `django-different-port-test-app`. The installed django version is `2.2.28` and it's apparently too old - Python 3.12 doesn't contain distutils.  

Traceback:
```
(app-root) python -c "import django"
  File "<string>", line 1, in <module>
  File "/opt/app-root/lib64/python3.12/site-packages/django/__init__.py", line 1, in <module>
    from django.utils.version import get_version
  File "/opt/app-root/lib64/python3.12/site-packages/django/utils/version.py", line 6, in <module>
    from distutils.version import LooseVersion
ModuleNotFoundError: No module named 'distutils'
```

I'm going to track all the test failures and see how many issues there are total, then mitigate them.

befeleme:
Findings:

- `'setup-test-app'` and `setup-requirements-test-app` contain `setup.py` and need explicit `setuptools` to work. Solution: add setuptools to the `requirements.txt`. We've got setuptools 67.7.2 in Fedora 39, so I'd pin the version to match the tested in our environment.

- `django-test-app` - Django 4 doesn't officially support Python 3.12 [yet](https://forum.djangoproject.com/t/django-4-2-lts-compatibility-with-python-3-12/24508). We've got 4.2.3 (4.2.6 in testing) in Fedora 39, so it should somehow work even before the official stamp is done by the Django devs. With 4.2.6 container builds, but we encounter an [error](https://stackoverflow.com/questions/70319606/importerror-cannot-import-name-url-from-django-conf-urls-after-upgrading-to). It looks like the best way forward is to fix the test django app to stop importing the removed bits. I have not yet checked whether the new recommended way is backwards compatible.

- `numpy-test-app` - again, the installed numpy is probably too old. Version 1.22.4 doesn't have wheels for Python 3.12 which causes trouble. Next step: find which numpy version is compatible and add it to the test.

- `pipenv-test-app`, `micropipenv-test-app` - somehow, somewhere the app imports requests which import urrlib3 which tries to import `from .packages.six.moves.http_client import (` resulting in `ModuleNotFoundError: No module named 'urllib3.packages.six.moves'`. Next step: look whether the culprit is really requests and find a newer version for test.

- `npm-virtualenv-uwsgi-test-app` - uWSGI doesn't have wheels for Python 3.12, local build fails. Solution: wait for wheels?

- `mod-wsgi-test-app` - mod_wsgi also builds wheels during the image build, image builds successfully but when run, it outputs:
```
httpd (mod_wsgi-express)  : Syntax error on line 163 of /tmp/mod_wsgi-localhost:8080:1001/httpd.conf: Cannot load /opt/app-root/lib64/python3.12/site-packages/mod_wsgi/server/mod_wsgi-py312.so into server: /opt/app-root/lib64/python3.12/site-packages/mod_wsgi/server/mod_wsgi-py312.so: cannot open shared object file: No such file or directory
```
Solution: wait for wheels?

befeleme:
I'm unable to reproduce Python 3.11 test failure from the last run locally.

befeleme:
Rebased on top of master + added commits mitigating the two first issues. I'll continue to work on the others.

zmiklank:
[test]

befeleme:
[test]

frenzymadness:
Looking at the failures:
### django-different-port-test-app

```
File "/opt/app-root/src/project/urls.py", line 28, in <module>
    url(r'^admin/', admin.site.urls),
    ^^^
NameError: name 'url' is not defined
```

I guess we need to update the testing project to be compatible with the newest Django.

### pipenv-test-app, pyuwsgi-pipenv-test-app, micropipenv-test-app, pin-pipenv-version-test-app

Error:

```
grep: warning: stray \ before -
```

Not sure what it means and where it comes from, have to investigate more.

### mod-wsgi-test-app, micropipenv-requirements-test-app

```
httpd (mod_wsgi-express)    : Syntax error on line 163 of /tmp/mod_wsgi-localhost:8080:100001/httpd.conf: Cannot load /opt/app-root/lib64/python3.12/site-packages/mod_wsgi/server/mod_wsgi-py312.so into server: /opt/app-root/lib64/python3.12/site-packages/mod_wsgi/server/mod_wsgi-py312.so: cannot open shared object file: No such file or directory
Test for image 'quay.io/fedora/python3-312:0' FAILED (exit code: 1)
```

The extension is compiled successfully but it cannot be loaded for some reason.

befeleme:
> Looking at the failures:
> ### django-different-port-test-app
> 
> ```
> File "/opt/app-root/src/project/urls.py", line 28, in <module>
>     url(r'^admin/', admin.site.urls),
>     ^^^
> NameError: name 'url' is not defined
> ```
> 
> I guess we need to update the testing project to be compatible with the newest Django.

I've got a commit fixing that.

> ### mod-wsgi-test-app, micropipenv-requirements-test-app
> 
> ```
> httpd (mod_wsgi-express)    : Syntax error on line 163 of /tmp/mod_wsgi-localhost:8080:100001/httpd.conf: Cannot load /opt/app-root/lib64/python3.12/site-packages/mod_wsgi/server/mod_wsgi-py312.so into server: /opt/app-root/lib64/python3.12/site-packages/mod_wsgi/server/mod_wsgi-py312.so: cannot open shared object file: No such file or directory
> Test for image 'quay.io/fedora/python3-312:0' FAILED (exit code: 1)
> ```
> 
> The extension is compiled successfully but it cannot be loaded for some reason.

The extension is `/opt/app-root/lib64/python3.12/site-packages/mod_wsgi/server/mod_wsgi-py312.cpython-312-x86_64-linux-gnu.so` - the filename is different than expected but I need to look deeper where the expectation about `mod_wsgi-py312.so` comes from.

frenzymadness:
[test]

frenzymadness:
This is the log from `ct_build_image_and_parse_id`:
```
+ ct_build_image_and_parse_id /tmp/tmp.a3Kn3B0U4r/Dockerfile.6ZBj ' -t quay.io/fedora/python3-312:0-testapp .'
+ local tmpdir
+ local log_file
+ local ret_val
+ local dockerfile
+ local command
+ local pid_build
+ local pid_sleep
+ local sleep_time
++ mktemp
+ log_file=/tmp/tmp.tKl1W75wi7
+ sleep_time=10m
+ '[' -n /tmp/tmp.a3Kn3B0U4r/Dockerfile.6ZBj ']'
+ dockerfile='-f /tmp/tmp.a3Kn3B0U4r/Dockerfile.6ZBj'
++ echo 'docker build --no-cache -f /tmp/tmp.a3Kn3B0U4r/Dockerfile.6ZBj  -t quay.io/fedora/python3-312:0-testapp .'
++ tr -d \'
+ command='docker build --no-cache -f /tmp/tmp.a3Kn3B0U4r/Dockerfile.6ZBj  -t quay.io/fedora/python3-312:0-testapp .'
+ pid_build=49682
+ docker build --no-cache -f /tmp/tmp.a3Kn3B0U4r/Dockerfile.6ZBj -t quay.io/fedora/python3-312:0-testapp .
+ pid_sleep=49683
+ wait 49682
+ trap 'exit 0' ALRM
+ sleep 10m
++ exit 0

[1m[INFO] TEST FAILED (positive), EXPECTED:0 GOT:1[0m
```
I don't know whether the build is killed because it takes too long or it just fails but the strange thing is that the function should `cat` the logfile no matter what happens and that's not happening here. @zmiklank and @phracek any idea what might be wrong here? I see that the function is from 8 to 10 months old.

zmiklank:
> I don't know whether the build is killed because it takes too long or it just fails but the strange thing is that the function should cat the logfile no matter what happens and that's not happening here. @zmiklank and @phracek any idea what might be wrong here? I see that the function is from 8 to 10 months old.

Yes, that function is quite new. And as you say, it should **always** `cat` the logs. I'll look into it.

befeleme:
[test]

befeleme:
I added a fix for `mod-wsgi-test-app` - the details were reported [upstream](https://github.com/GrahamDumpleton/mod_wsgi/issues/865) and I've verified that suggested version 5.0.0rc1 works with Python 3.12.

befeleme:
[test]

befeleme:
4 tests actually failed, the successful result of the pipeline is moot.

zmiklank:
> 4 tests actually failed, the successful result of the pipeline is moot.

That is also work for me. I see the problem, will follow up with PR. Thanks for catching this.

I did not find out what is wrong with the `ct_build_image_and_parse_id` function yet. I plan to look at this further today.

zmiklank:
I might be on a right track now. It seems that the culprit is a forgotten `set -e` in a subshell in `ct_s2i_build_as_df` function:
https://github.com/sclorg/container-common-scripts/blob/250ae7c6b9312e2389b08d349a1a95c59ed1dde9/test-lib.sh#L911

This with combination of build failing (tried locally) lead to missing logs.
When I have been removing the exit on error bash option from test-lib I left this one as it this; I though - it is in subshell, what wrong can happen :D.

Unfortunately the fix can take some time. Simply removing the `set -e` could be a hotfix, but some more thorough fix will be needed too.
This PR indeed is great for catching the bugs we have here ;).

befeleme:
[test]

befeleme:
rebased on top of master, updated submodules, regenerated distgen content

[test]

zmiklank:
I have checked the logs to see if the fixes in test-lib.sh were indeed successful, but the common submodule does not seem to be updated. Its HEAD is on [250ae7c](https://github.com/sclorg/container-common-scripts/tree/250ae7c6b9312e2389b08d349a1a95c59ed1dde9).

befeleme:
You're right, rebased. Thanks to the better output I've also found the problem in `pin-pipenv-version-test-app` and `pyuwsgi-pipenv-test-app`, they require too old pipenv to work with Python 3.12: `PIN_PIPENV_VERSION=2021.5.29`. I don't know yet how to mitigate this, we need this version for compatibility with e.g. Python 2.

befeleme:
[test]

zmiklank:
[test-openshift]

befeleme:
[test]

befeleme:
[test]

befeleme:
[test]

befeleme:
`pyuwsgi-pipenv-test-app` - the failure is unfortunately known: https://github.com/unbit/uwsgi/issues/2579 and it seems we can't do much about it.

otherwise :tada: 
this is ready for review :D

befeleme:
I added three fixup commits for easier review. 
I have not added setuptools to the older {micro,}pipenv-test-app Pipfiles (https://github.com/sclorg/s2i-python-container/pull/643/commits/613cc876f254e81e8606590d96d74a15c53ba8c5), because that'd require regenerating Pipfile.lock and it didn't seem worth the effort.

befeleme:
[test]

befeleme:
+ one fixup with `mod_wsgi >= 5.0.0` rather than fixed on `5.0.0`.

manifest-minimal.yml
```diff
@@ -37,6 +37,9 @@ DISTGEN_RULES:
   - src: src/test/from-dockerfile/uwsgi.Dockerfile.tpl
     dest: test/from-dockerfile/uwsgi.Dockerfile.tpl
 
+  - src: src/test/pin-pipenv-version-test-app/environment
```
frenzymadness, line 40: One last nitpick - I'd keep the paths in sync - I mean - if the file is generated as `.s2i/environment`, I'd keep it in the .s2i folder in the source of the application as well.

manifest.yml
```diff
@@ -37,6 +37,11 @@ DISTGEN_RULES:
   - src: src/test/from-dockerfile/Dockerfile_no_s2i.tpl
     dest: test/from-dockerfile/Dockerfile_no_s2i.tpl
 
+  - src: src/test/pin-pipenv-version-test-app/environment
```
frenzymadness, line 40: Same here.

frenzymadness (commented):
One last change. Feel free to squash the fixup commits after that and run CI again. I think we are almost there.

befeleme:
Fixed, squashed. 

[test]

PR #644, opened 2023-10-12
Add Python 3.12 to the CI test matrix

befeleme:
Prep step for #643


zmiklank (approved):
Looks good, thanks :).

phracek (approved):
Thanks for bringing this pull request.

zmiklank:
Just note, that build and push action will be failing until you add the Dockerfile for python 3.12, Fedora. I am OK with that.
Other option is to add 3.12 to test matrix only, and add build-and-push configuration in the PR https://github.com/sclorg/s2i-python-container/pull/643.

PR #646, opened 2023-10-26
Fix typo in username for build-and-push python-311-minimal to quay.io/fedora

phracek:
Fix typo in username for build-and-push python-311-minimal to quay.io/fedora


phracek:
Running tests is not needed. GitHub Action is fixed by this pull request.

PR #647, opened 2023-11-09
use /bin/bash instead of /sbin/nologin as default shell for default user

zmiklank:
Related: https://github.com/sclorg/s2i-base-container/pull/284
Partly fixes: https://github.com/sclorg/s2i-base-container/issues/283


zmiklank:
[test-all]

phracek (approved):
LGTM. Thanks for this pull request, I did not hit any issues.
OpenShift 4 failures does not block this pull request.  It will be fixed by the other pull request and by fix in dancer-ex repository.

Issue #659, opened 2024-03-04
Publish arm64 images


tonycosentini:
### Container platform

Podman/Docker

### Version

Non-RHEL images hosted on Quay.

### OS version of the container image

CentOS Stream 9

### Description

Is it possible to start publishing arm64 images to Quay? I was planning on switching off of UBI8 and onto c9s, but there is no arm support.

UBI 8 example:
```
➜  docker manifest inspect registry.access.redhat.com/ubi9/python-311     
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 926,
         "digest": "sha256:9101b9924d06a4944649e9cb008496813479d66271ee1f2a1b6ddfc91228c97c",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 926,
         "digest": "sha256:d4b195e9afabfd9184f7d6ca58a5358ff9b11d6367d61c2d587bb6c695671097",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 926,
         "digest": "sha256:5eaecbc30e36eaf5c60c43515a1453af10df77f2c483535432eb66b656e1cf7a",
         "platform": {
            "architecture": "ppc64le",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 926,
         "digest": "sha256:36f0e63073521a14a584619d84d3c5332e05dd65f6856853fa6064992876b28b",
         "platform": {
            "architecture": "s390x",
            "os": "linux"
         }
      }
   ]
}
```

c9s example:
```
➜ docker manifest inspect quay.io/sclorg/python-311-c9s               
{
	"schemaVersion": 2,
	"mediaType": "application/vnd.docker.distribution.manifest.v2+json",
	"config": {
		"mediaType": "application/vnd.docker.container.image.v1+json",
		"size": 16094,
		"digest": "sha256:a4469a80a000d2d24c745ab31c635ec3b0c479bf56462b1079e05f2ddccf9ae0"
	},
	"layers": [
		{
[...]

PR #660, opened 2024-03-06
build Fedora and CentOS Stream images also for arm64

zmiklank:
Depends on: https://github.com/sclorg/s2i-base-container/pull/292
Resolves: https://github.com/sclorg/s2i-python-container/issues/659

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -97,3 +97,4 @@ jobs:
           dockerfile: ${{ matrix.dockerfile }}
           tag: ${{ matrix.tag }}
           image_name: ${{ matrix.image_name }}
+          archs: amd64, arm64


phracek (approved):
LGTM. But question. Why to not enable all archs like in postgresql-container
```
archs: amd64, s390x, ppc64le, arm64
```

I think the new request will come soon if we support part of architectures https://github.com/sclorg/postgresql-container/blob/master/.github/workflows/build-and-push.yml#L121

zmiklank:
> LGTM. But question. Why to not enable all archs like in postgresql-container

Sure, we can do that. I was going with minimal option, but I am OK also with enabling all the arch. If not for python, then at least for s2i base and core. Should I update the PR accordingly?

pkubatrh:
+1 to enabling all for s2i-core and s2i-base (since we already build all for postgresql anyway), I would keep only arm for python, as we didnt get any requests for other architectures yet.

zmiklank:
Done for s2i-base and s2i-core:
https://github.com/sclorg/s2i-base-container/pull/292/

frenzymadness (approved):
I like it and I'm fine with minimum approach here and all architectures for base and core.

PR #675, opened 2024-04-30
Support for Testing Python container by container-ci-suite

phracek:
This pull request uses testing s2i-python-container in OpenShift 4 environment
by container-ci-suite.


test/test_deploy_templates.py
```diff
@@ -0,0 +1,73 @@
+import os
+import sys
+import pytest
```
frenzymadness, line 3: If you split imports, there are usually three groups: standard library, third-party libs (installed dependencies) and local modules.

test/test_deploy_templates.py
```diff
@@ -0,0 +1,73 @@
...
+from pathlib import Path
+
+from container_ci_suite.openshift import OpenShiftAPI
+from container_ci_suite.utils import get_service_image, check_variables
+test_dir = Path(os.path.abspath(os.path.dirname(__file__)))
```
frenzymadness, line 8: You use all upper case names for constant and this is also constant.

test/test_imagestreams_quickstart.py
```diff
@@ -0,0 +1,76 @@
...
+    DEPLOYED_PSQL_IMAGE = "quay.io/sclorg/postgresql-12-c8s:c8s"
+    IMAGE_TAG = "postgresql:12"
+    PSQL_VERSION = "12"
+else:
+    BRANCH_TO_TEST = "master"
+    DEPLOYED_PSQL_IMAGE = "quay.io/centos7/postgresql-10-centos7:centos7"
+    IMAGE_TAG = "postgresql:10"
+    PSQL_VERSION = "10"
+
+if OS == "rhel7":
```
frenzymadness, line 31: A dictionary (mapping) might be better than many ifs.

test/test_python_ex_template.py
```diff
@@ -0,0 +1,44 @@
...
+import sys
+import pytest
+from pathlib import Path
+
+from container_ci_suite.openshift import OpenShiftAPI
+from container_ci_suite.utils import get_service_image, check_variables
+test_dir = Path(os.path.abspath(os.path.dirname(__file__)))
+
+if not check_variables():
+    sys.exit(1)
```
frenzymadness, line 11: What about some error message here is the exit code is non-zero?

frenzymadness (commented):
I understand this is just a draft so I did just a very quick review.

test/test_python_ex_template.py
phracek: The messages are here: https://github.com/sclorg/container-ci-suite/blob/main/container_ci_suite/utils.py#L242.

test/test_imagestreams_quickstart.py
phracek: Good point.

phracek:
[test-openshift-pytest]

phracek:
Let's use 'latest' tag instead of 'c8s' tag

[test-openshift-pytest]

phracek:
Let's use centos7 back again for testing.

[test-openshift-pytest]

test/test_deploy_templates.py
phracek: In case it is used, then UPPER_CASED

test/test_imagestreams_quickstart.py
phracek: DONE

test/test_python_ex_template.py
phracek: Done

test/test_deploy_templates.py
phracek: Split to three groups.

phracek:
[test-openshift-pytest]

phracek:
After adding condition for Python-2.7 and RHEL7.

Running re-test again.

[test-openshift-pytest]

phracek:
@frenzymadness It looks everything passed. Please take a look at it. Thanks.

test/test_deploy_templates.py
frenzymadness: Yes, but:
* pathlib is in the standard library and should be in the same group as os and sys
* pytest is third-party and should be alone in a group.

test/test_latest_imagestreams.py
```diff
@@ -0,0 +1,28 @@
...
+    sys.exit(1)
+
+VERSION = os.getenv("SINGLE_VERSION")
+
+
+# Replacement with 'test_latest_imagestreams'
+class TestLatestImagestreams:
+
+    def setup_method(self):
+        self.isc = ImageStreamChecker(working_dir=TEST_DIR / ".." / "..")
```
frenzymadness, line 23: What about:

```ipython
In [1]: from pathlib import Path

In [2]: p = Path("/a/b/c/d/e/f")

In [3]: p.parent.parent
Out[3]: PosixPath('/a/b/c/d')
```

```suggestion
        self.isc = ImageStreamChecker(working_dir=TEST_DIR.parent.parent)
```

test/test_python_ex_standalone.py
```diff
@@ -0,0 +1,36 @@
...
+    def setup_method(self):
+        self.oc_api = OpenShiftAPI(pod_name_prefix="python-testing", version=VERSION)
+
+    def teardown_method(self):
+        self.oc_api.delete_project()
+
+    def test_python_ex_template_inside_cluster(self):
+        service_name = "python-testing"
+        assert self.oc_api.deploy_s2i_app(
+            image_name=IMAGE_NAME, app=f"https://github.com/sclorg/s2i-python-container.git",
```
frenzymadness, line 29: F-string without any variable inside.

```suggestion
            image_name=IMAGE_NAME, app="https://github.com/sclorg/s2i-python-container.git",
```

frenzymadness (commented):
I've found only a few possible improvements and you can freely ignore them if you want. It looks good and it works. Thank you.

phracek:
[test-all]

phracek:
Third-party libraries were separated from build-in

[test-all]

PR #680, opened 2024-05-27
deprecate c8s images, as EOL for c8s is 31th of May 2024

zmiklank:


frenzymadness (changes requested):
Thank you for the PR. You can also remove the whole c8s definition from the multispec file.

zmiklank:
> Thank you for the PR. You can also remove the whole c8s definition from the multispec file.

I try to handle all container repositories the same way- it makes them easier for maintenance. And in the container meeting we agreed we would make this change with adding .exclude files, leaving Dockerfiles as they are (in distgen world removing from include matrix/adding to exclude matrix).

If you want the c8s to be removed anyway I understand and can do that (and varnish and postgres will need to be adjusted too).

frenzymadness:
> If you want the c8s to be removed anyway I understand and can do that (and varnish and postgres will need to be adjusted too).

I don't have a strong opinion here but I don't understand the benefit of having the configuration section for the distribution we won't use anymore. Or am I missing something?

zmiklank:
> > If you want the c8s to be removed anyway I understand and can do that (and varnish and postgres will need to be adjusted too).
> 
> I don't have a strong opinion here but I don't understand the benefit of having the configuration section for the distribution we won't use anymore. Or am I missing something?

I think the main reason behind it is that even though we will stop build the images anyone can still easily take the Dockerfiles and build their own images. But you are right this does not have much sense in distgen based approach, as the user would still need to adjust the matrix and regenerate the source codes.

zmiklank:
> Thank you for the PR. You can also remove the whole c8s definition from the multispec file.

Done.

frenzymadness:
Thank you!

PR #691, opened 2024-06-12
GitHub Action for running PyTest in OpenShift

phracek:
This pull request adds possibility to execute container-ci-suite as PyTest for s2i-python-container

It blocks #675


phracek:
See similar PR: https://github.com/sclorg/httpd-container/pull/216

PR #700, opened 2024-07-19
Use VERSION instead of SINGLE_VERSION

phracek:
VERSION is used in our whole sclorg organization.
Like here: https://github.com/sclorg/container-common-scripts/blob/master/test.sh#L81


phracek:
[test-openshift-pytest]

phracek:
Tests PASSED. We can merge it.

phracek:
The change does not break python tests or python code itself.

PR #701, opened 2024-07-25
Fix PyTest test suite

phracek:
This pull request adds condition also for Python-3.12.

The pull request also removes 'django-postgresql' storage as it does not make sense.
Only 'persistent' is used in real production. It also adds 'django' application alone.

Allow only persistent Django storage and Django alone for testing


phracek:
[test-all]

phracek:
It blocks our nightly builds

frenzymadness (approved):

phracek:
[test-all]

phracek:
[test-all]

PR #703, opened 2024-08-14
Fix links to available images in README.md

frenzymadness:


frenzymadness:
Links verified manually and they work.

PR #704, opened 2024-08-14
Python 3.12 minimal container image

frenzymadness:
We've received a request to provide minimal container image with Python 3.12. Let's see what CI thinks about its quality.


frenzymadness:
[test]

frenzymadness:
[test]

3.12-minimal/Dockerfile.fedora
```diff
@@ -0,0 +1,110 @@
+FROM quay.io/fedora/fedora-minimal:39
```
phracek, line 1: I would prefer to use F40 or even F41 if it is possible. F39 will be EOL soon in Dec 2024.

specs/multispec.yml
```diff
@@ -85,14 +85,27 @@ specs:
       distros:
         - fedora-39-x86_64
       fedora_version: "39"
+      minimal_image: "quay.io/fedora/fedora-minimal:39"
```
phracek, line 88: Fedora 40 would be better if it is feasible.

phracek (changes requested):
What do you think about my comments?

frenzymadness:
This PR is just about adding 3.12-minimal. I'll move both 3.12 to Fedora 40 soon (Fedora 41 has 3.13) and drop 3.11 on Fedora 38 as well.

befeleme (approved):
The changes look good, the CI is happy - I inspected the results and haven't found any issues.

PR #706, opened 2024-08-21
Allow testing s2i-python-container in C10S

phracek:
This pull request only allows testing s2i-python-container in C10S.

No Python functionality is broken.
Only GitHub Action is updated.


phracek:
[test]

PR #712, opened 2024-09-13
Support testing helm charts

phracek:
This pull request adds support for testing Helm charts
present in repository https://github.com/sclorg/helm-charts.

Similar tests in OpenShift are done in 'example' repositories like https://github.com/sclorg/django-ex


phracek:
Let's first round of tests

[test]

phracek:
@frenzymadness PTAL. Thanks. Does not break s2i-python container functionality. Only tests Helm charts that are replaces by templates and imagestreams.

PR #728, opened 2025-01-27
switch to ubuntu-latest from ubuntu-20.04

phracek:
Use ubuntu-latest as it is already used in postgresql-container

https://github.com/sclorg/postgresql-container/blob/master/.github/workflows/container-tests.yml#L8

dist-gen with RHEL10 support is only distributed on newer python version interpreters


phracek:
[test]

frenzymadness (approved):
The change looks good and the errors in the CI seem unrelated.

PR #729, opened 2025-01-27
Remove deprecated --no-python-version-warning option from pip

frenzymadness:
Discussion: https://github.com/pypa/pip/issues/13154

This should have no effect. The deprecation warning produced by the new pip causes the tests to fail.


frenzymadness:
[test]

frenzymadness:
@phracek this is another fix for the CI that should make the RHEL 10 PR look better.

PR #743, opened 2025-06-05
Update github action with testing version 3.12, and 3.12-minimal

phracek:
Update github action with testing version 3.12, and 3.12-minimal

by PyTest suite used in OpenShift 4.
The pull request does not modify python code or test suite itself at all.

The next pull request will add RHEL10 imagestreams that should be used by this pull request.


phracek:
Running tests is not needed.

PR #744, opened 2025-06-05
Add RHEL10 imagestreams

phracek:
Adds testing RHEL10 imagestreams in python container.

This pull request is blocked by https://github.com/sclorg/s2i-python-container/pull/743

### DIFF
--- a/test/test-lib-python.sh
+++ b/test/test-lib-python.sh
@@ -12,16 +12,13 @@ source "${THISDIR}/test-lib.sh"
 source "${THISDIR}/test-lib-openshift.sh"
 
 function ct_pull_or_import_postgresql() {
-  if [[ "${VERSION}" == *"minimal"* ]]; then
-    VERSION=$(echo "${VERSION}" | cut -d "-" -f 1)
-  fi
-  if [[ "${VERSION}" == "3.11" ]] || [[ "${VERSION}" == "3.12" ]]; then
+  if [[ "${VERSION}" == "3.11" ]] || [[ "${VERSION}" == "3.12" ]] || [[ "${VERSION}" == "3.12-minimal" ]]; then
     postgresql_image="quay.io/sclorg/postgresql-12-c8s"
     image_short="postgresql:12"
     image_tag="${image_short}"
   else
-    postgresql_image="quay.io/centos7/postgresql-10-centos7:centos7"
-    image_short="postgresql:10-centos7"
+    postgresql_image="quay.io/sclorg/postgresql-10-c8s:c8s"
+    image_short="postgresql:10-c8s"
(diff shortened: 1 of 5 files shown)


phracek:
Fix typo in 'test-lib-python.sh'

[test-all]

phracek:
The latest commit https://github.com/sclorg/s2i-python-container/pull/744/commits/4f1f147401f1e245787f5363dd9bcca750dd02ea fixes proper detection json file.
Also PyTest Django Templates uses proper ubi tag. Instead of `3.12-minimalubi10` it should be `3.12-minimal-ubi10`

[test-all]

test/test-lib-python.sh
```diff
@@ -62,7 +62,7 @@ function test_python_imagestream() {
 django-postgresql.json \
 django-postgresql-persistent.json"
   for template in $TEMPLATES; do
-    ct_os_test_image_stream_quickstart "${THISDIR}/imagestreams/python-${OS%[0-9]*}.json" \
+    ct_os_test_image_stream_quickstart "${THISDIR}/imagestreams/python-${OS//[0-9]/}.json" \
```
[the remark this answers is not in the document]
befeleme: Aach, I see. `//` replaces all matches of digits with nothing after `/`. `%` removes the suffix from the end. The same effect would have an expression `${OS%%[0-9]*}` with double % sign to indicate the removal of the longest match from the end of the string. I have no opinion on what's better, but I learned something new today :)

test/test-lib-python.sh
```diff
@@ -40,16 +40,14 @@ function ct_pull_or_import_postgresql() {
...
     tag="-ubi9"
+  elif [ "${OS}" == "rhel10" ]; then
+    tag="-ubi10"
   fi
-  if [[ "${VERSION}" == *"minimal"* ]]; then
-    VERSION=$(echo "${VERSION}" | cut -d "-" -f 1)
-  fi
-  if [[ "${VERSION}" == "3.11" ]] || [[ "${VERSION}" == "3.12" ]]; then
+  # On RHEL10 we have only 3.12-minimal image
+  if [[ "${VERSION}" == "3.11" ]] || [[ "${VERSION}" == "3.12" ]] && [[ "${VERSION}" == "3.12-minimal" ]]; then
```
befeleme: IIUC, this evaluates to: if ${VERSION} is 3.11 or if ${VERSION} is both "3.12" and "3.12-minimal", the block will be executed. Isn't the second condition always false? That would mean it will get only executed if ${VERSION} is 3.11. Is this intended? Am I misunderstanding?

befeleme: Oh, sorry, I see it's corrected in the next commit.

PR #755, opened 2025-09-15
Drop VERSION env variable and version label

frenzymadness:
And let's see what will happen in CI.


frenzymadness:
It seems that the support for this would have to be implemented in our common scripts, namely: https://github.com/sclorg/container-common-scripts/blob/dc69f42462c1f0ac4e5c79e0317f71f6094626db/tag.sh#L35

phracek:
> It seems that the support for this would have to be implemented in our common scripts, namely: https://github.com/sclorg/container-common-scripts/blob/dc69f42462c1f0ac4e5c79e0317f71f6094626db/tag.sh#L35

There is a pull request that removes it here: https://github.com/sclorg/container-common-scripts/pull/406

frenzymadness:
Let me experiment. I'll try to use the common scripts from the PR you've prepared. It should test it thoroughly.

PR #757, opened 2025-09-24
Add Python 3.12 minimal image based on RHEL 8

frenzymadness:
We have a CI configuration for this combo, but the docker file is missing for some reason.


frenzymadness:
[test]

PR #758, opened 2025-09-30
Add function `is_test_allowed` for specific OS and version.

phracek:
Some containers are not available for Helm Charts
and therefore the test should be skipped.

The Minimal containers that are allowed are mentioned in Red Hat Container Catalog see. https://catalog.redhat.com/en/search?searchType=Containers&q=python+minimal&p=1


phracek:
Let's re-run test after fixing `test_python_imagestream` for rhel8 where the image does not exist.

[test-openshift][test-openshift-pytest]

PR #828, opened 2026-05-27
Remove support for Python 3.11 (EOL in May 2026)

frenzymadness:
This should technically wait for the end of May, but I don't expect anything new for 3.11 coming from this repository, so we can likely merge it even sooner.

PR #837, opened 2026-08-14
Fix latest tag referencing non-existent 3.12-ubi8

aroyoredhat:
## Summary

The `latest` ImageStreamTag in `python-rhel.json` and `python-rhel-aarch64.json` currently references `3.12-ubi8`, which does not exist as a defined tag in those files. This was likely left behind when `3.12-ubi8` was removed.

This PR updates the `latest` tag to point to `3.12-ubi9` (the highest non-minimal Python 3.12 tag available) and updates the description from "UBI 8" to "UBI 9" to match. The corresponding entries in `imagestreams.yaml` are also fixed.

### Files changed:
- `imagestreams/python-rhel.json` — `latest.from.name`: `3.12-ubi8` → `3.12-ubi9`, description updated
- `imagestreams/python-rhel-aarch64.json` — same fix
- `imagestreams/imagestreams.yaml` — `latest` field updated for both rhel and rhel-aarch64 entries

### Related:
- https://github.com/sclorg/s2i-python-container/pull/828 (removed 3.11 support but left a dangling `latest` reference to the now-removed `3.12-ubi8`)

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Updates**
  * Updated the latest Python 3.12 image streams to use Red Hat Universal Base Image (UBI) 9.
  * Applied the update to both standard and AArch64 image variants.
  * Existing image version labels and metadata remain unchanged.


hhorak:
While moving to ubi9 (or even ubi10) for latest makes sense, I don't think that the 3.12-ubi8 being missing is actually expected -- @frenzymadness @phracek do you agree that 3.12-ubi8 should still exist? Maybe we should fix that problem first, only then looking at what latest references.

hhorak:
Confirmed `registry.access.redhat.com/ubi8/python-312` image exists, so we should have it in the image streams.

